PR #2103, opened 2014-12-19
Pull Lantern changes to support TCP (fallback) proxying in flashlight

oxtoacart:
Requires flashlight built after merging https://github.com/getlantern/flashlight-build/pull/4

Also need to update BaseS3Config to point at at new url for the config file, since the flashlight.yaml format has changed.  See https://github.com/getlantern/lantern-common/blob/master/src/main/java/org/lantern/BaseS3Config.java#L16.


myleshorton:
I believe this still needs the new flashlight binaries on devel @oxtoacart correct?

oxtoacart:
> I believe this still needs the new flashlight binaries on devel @oxtoacart correct?

Yes, and a new config url.  See https://github.com/getlantern/lantern/pull/2103. Would you like to try building flashlight to make sure I'm not the only one who can build it? I know you've got a lot of reporting stuff, so I'm happy to just do it too.

atavism:
> Would you like to try building flashlight to make sure I'm not the only one who can build it?

I'm happy to try building it too.

oxtoacart:
@atavism Great idea, thanks!  I've assigned it over to you.

You can build flashlight using tagandbuild.bash per the instructions here - https://github.com/getlantern/flashlight-build#building-flashlight

Maybe we call this release 1.6.0?

Note:

> Also need to update BaseS3Config to point at at new url for the config file, since the flashlight.yaml format has changed. See https://github.com/getlantern/lantern-common/blob/master/src/main/java/org/lantern/BaseS3Config.java#L16.

You can generate and upload a new flashlight.yaml per the instructions here - https://github.com/getlantern/flashlight-build/tree/devel/src/github.com/getlantern/flashlight#masquerade-host-management

oxtoacart:
@atavism, btw go 1.4.1 is out, which you can install with gonative:

`gonative build -version="1.4.1" -platforms="darwin_amd64 linux_386 linux_amd64 windows_386"`

atavism:
@oxtoacart Thanks! Off of the devel branch, right? Looks like everything builds fine for me:

```
➜  flashlight-build git:(devel) ✗ ./tagandbuild.bash 1.6.0
Updated tag '1.6.0' (was 7dea443)
Counting objects: 1, done.
Writing objects: 100% (1/1), 156 bytes | 0 bytes/s, done.
Total 1 (delta 0), reused 0 (delta 0)
To https://github.com/getlantern/flashlight-build
 + 7dea443...50d70d8 1.6.0 -> 1.6.0 (forced update)
Building flashlight version 1.6.0 (20150120.220956)
Number of parallel builds: 8

-->       linux/386: github.com/getlantern/flashlight
-->     linux/amd64: github.com/getlantern/flashlight
-->     windows/386: github.com/getlantern/flashlight
-->    darwin/amd64: github.com/getlantern/flashlight
```

oxtoacart:
@atavism Perfect!  You can copy the executables over into lantern using `./copyexecutabes.bash`.  You'll need to set some environment variables for codesigning as described in the README.

atavism:
Updated flashlight yaml: http://s3.amazonaws.com/lantern_config/cloud.1.6.0.yaml.gz
cloud.1.6.0.yaml.gz

I created a new branch with the binaries here: https://github.com/getlantern/lantern/tree/flashlight-1.6.0

atavism:
@oxtoacart Just realized these changes still need code reviewed. I'll do that today. To take a holistic approach and familiarize myself with the entire implementation, citing an old e-mail of yours, the proxying relies on three new libraries: balancer, fronted, and chained. And the actual changes to flashlight itself are in the tcpproxying branch, correct?

oxtoacart:
@atavism 

> the proxying relies on three new libraries: balancer, fronted, and chained.

Correct.

> And the actual changes to flashlight itself have been merged to devel already, correct?

Yes.  The binaries that you built already include the necessary Go changes.

src/main/java/org/lantern/Launcher.java
```diff
@@ -279,26 +279,11 @@ void launch() {
             }
         }
         
-        proxyTracker = instance(ProxyTracker.class);
+        //proxyTracker = instance(ProxyTracker.class);
         httpClientFactory = instance(HttpClientFactory.class);
 
         s3ConfigFetcher = instance(S3ConfigFetcher.class);
         
-        if (checkFallbacks) {
```
atavism: With these changes, we don't need to handle the check-fallbacks mode at all anymore, right? There's other places in this class we can remove references to it, and we also can remove support for --check-fallbacks as a command line option.

oxtoacart: The check-fallbacks mode is used to test the fallback servers, which we still use.

atavism: Okay, thanks. With your changes, it looks like --check-fallbacks simply prevents other services from starting now, similar to --disable-ui:
```java
                // don't need to start the rest of these services when running in check-fallbacks mode
                if (checkFallbacks) {
                    return;
                }
```
I didn't see that command line option being used anywhere else.

atavism: And unless I'm searching incorrectly, with this block of code removed, FallbackChecker is no longer being used anywhere now, which is why I assumed we were removing fallback server testing entirely with this update.

src/main/java/org/lantern/S3ConfigFetcher.java
```diff
@@ -129,8 +129,12 @@ private boolean downloadAndCompareConfig() {
 
         final S3Config config = this.model.getS3Config();
         if (newConfig.isPresent()) {
-            this.model.setS3Config(newConfig.get());
-            return !newConfig.get().equals(config);
+            S3Config nextConfig = newConfig.get();
+            boolean changed = !newConfig.equals(config);
```
atavism: I think this is supposed to be
```java
   boolean changed = !nextConfig.equals(config);
```

oxtoacart: Yes, good catch!

src/main/java/org/lantern/Launcher.java
oxtoacart: Ahh, good point!  Need to think about this.

oxtoacart:
@atavism Okay, I've made updates to address your comments.  While in there, I noticed that DefaultProxyTracker was dead code as well, so I took that out.  And I noticed that we weren't picking up the initial S3Config on login correctly, so I fixed that too.

Seems to be running well.

oxtoacart:
Oh, I also reverse merged from devel to make for a clean pull.

atavism:
@oxtoacart Great. I suspected we'd be able to get rid of a lot of files! Ran some tests in get-mode and everything is working fine for me as well. Merging.

Issue #2095, opened 2014-12-17
[feature] [lantern-ui] When "Show Lantern" is clicked, check for existing instance of the UI


atavism:
Display an alert so that the user's attention is immediately directed to the tab instead of opening a new instance.

PR #2115, opened 2014-12-24
Changes to persist single UI session when "Show Lantern" is clicked

atavism:
This PR corresponds to https://github.com/getlantern/lantern/issues/2095


oxtoacart:
@atavism @myleshorton Sorry to chime in late, but I thought we had discussed this a while back and concluded that:

1. For Chrome, we would use the fancy behavior (i.e. alert to focus window, auto-close when opening a new window and stopping Lantern).

2. For other browsers, we would just always open a new tab, which IIRC will always grab focus no matter what you're doing.

Not that I have strong feelings about this, just thought I'd save you some time if I can.

oxtoacart:
@myleshorton With this merged to devel, I'm experiencing all sorts of issues opening the lantern ui.  Is there a lantern-ui change that's supposed to go along with this?

myleshorton:
Wo, weird, I totally didn't mean to merge this at all, as there are definitely lantern-ui changes. Backing out.

myleshorton:
> For Chrome, we would use the fancy behavior (i.e. alert to focus window, auto-close when opening a new window and stopping Lantern).
> 
> For other browsers, we would just always open a new tab, which IIRC will always grab focus no matter what you're doing.

Yeah that's how we're planning on doing it.

myleshorton:
The odd thing is this merge doesn't appear in the commit history at https://github.com/getlantern/lantern/commits/devel

myleshorton:
I just realized what happened -- I think I hadn't re-fetched devel before committing the change to add India to the censored list. One sec.

atavism:
@myleshorton The latest changes--to send an alert when the browser is Chrome and the UI is already open in another tab; or always open the UI in a new tab otherwise--are in the 2095 branch. I also pushed an issue-2095 branch in case the numeric branch name causes problems. On the client-side, the latest changes are in 0.0.12. Let me know if you need me to submit another PR. 
> Yeah that's how we're planning on doing it.

myleshorton:
Cool. If you can open another PR @atavism that would be great.

PR #2131, opened 2015-01-03
Issue 2095 - UI relays browser name

atavism:
Changes to store browser name the UI is running in. This also includes updates to send an alert when the browser is Chrome and the UI is already open in an existing tab; otherwise, the UI is always opened in a new tab.


myleshorton:
Can you also update the lantern-ui submodule to point to 0.0.12 head with this pull?

atavism:
@myleshorton Done.

myleshorton:
OK. LGTM. I'll preemptively merge, although I'm sure the 0.0.12 review will create some changes that will move the submodule commit. Let's just handle that as necessary.

PR #2147, opened 2015-01-08
Closes #2124

### LINKED ISSUE #2124: Detect Lantern Pi on local network using multicast

myleshorton:


src/main/java/org/lantern/LanternUtils.java
```diff
@@ -1371,4 +1371,14 @@ public static boolean isShuttingDown() {
...
     }
+
+    /**
+     * Utility for whether or not we're running on a Pi. Just checks for ARM 
+     * for now.
+     * 
+     * @return <code>true</code> if we're on ARM, otherwise <code>false</code>.
+     */
+    public static boolean isLanternPi() {
+        return SystemUtils.OS_ARCH.toLowerCase().contains("arm");
```
oxtoacart: If you make this a command-line flag and use the OS_ARCH check to set the default of that flag, then we can test the lantern-pi code path on our usual development machines.

src/main/java/org/lantern/http/JettyLauncher.java
```diff
@@ -119,7 +120,9 @@ public void start(final int port) {
         connector.setAcceptQueueSize(5000);
 
 
-        if (this.model.getSettings().isBindToLocalhost()) {
+        // We always allow anyone to connect to Lantern Pi (arm).
+        if (this.model.getSettings().isBindToLocalhost() && 
+                !SystemUtils.OS_ARCH.contains("arm")) {
```
oxtoacart: Should this be using `LanternUtils.isLanternPi()`?

src/main/java/org/lantern/http/JettyLauncher.java
```diff
@@ -165,9 +168,11 @@ protected void doGet(HttpServletRequest req,
                 } else if (uri.startsWith("/proxy_google")) {
                     writeFileToResponse(resp, Proxifier.PROXY_GOOGLE);
                 } else {
-                    resp.addCookie(new Cookie("XSRF-TOKEN", model.getXsrfToken()));
-                    if (!resp.containsHeader("Content-Security-Policy")) {
-                        LanternUtils.addCSPHeader(resp);
+                    if (!LanternUtils.isDevMode() && !LanternUtils.isLanternPi()) {
```
oxtoacart: `isDevMode()` and `isLanternPi()` appear together a lot.  Can/should they be distilled into a single flag?

src/main/java/org/lantern/http/InteractionServlet.java
```diff
@@ -171,11 +173,13 @@ protected void processRequest(final HttpServletRequest req,
...
-            HttpUtils.sendClientError(resp, "invalid X-XSRF-TOKEN");
-            return;
+        if (!LanternUtils.isDevMode() && !LanternUtils.isLanternPi()) {
+            if (!SecurityUtils.constantTimeEquals(model.getXsrfToken(),
+                    req.getHeader("X-XSRF-TOKEN"))) {
+                log.debug("X-XSRF-TOKEN wrong: got {} expected {}", req.getHeader("X-XSRF-TOKEN"), model.getXsrfToken());
+                HttpUtils.sendClientError(resp, "invalid X-XSRF-TOKEN");
+                return;
+            }
         }
```
oxtoacart: Why turn off XSRF protection for lantern-pi mode?

src/main/java/org/lantern/http/JettyLauncher.java
```diff
@@ -165,9 +168,11 @@ protected void doGet(HttpServletRequest req,
...
+                        resp.addCookie(new Cookie("XSRF-TOKEN", model.getXsrfToken()));
+                        if (!resp.containsHeader("Content-Security-Policy")) {
+                            LanternUtils.addCSPHeader(resp);
+                        }
```
oxtoacart: Why not use the CSP policy for lantern-pi?  I imagine we can still do it if we just use the machine's IP address rather than "127.0.0.1" in our CSP header.

src/main/java/org/lantern/LanternUtils.java
myleshorton: We effectively shortcut that with using LanternUtils.isDev

src/main/java/org/lantern/http/InteractionServlet.java
myleshorton: Because it prevents accessing the Lantern UI from other machines on the network.

src/main/java/org/lantern/http/JettyLauncher.java
myleshorton: That would probably work, although we'd have to include both 127.0.0.1 and the local network address in order to keep all the current stuff working. With any caller already having to somehow guess the random path and port to even get here, however, to me the whole inclusion of CSP is pretty dubious. We've also seen it cause bugs even on localhost, and it's at least the first thing that fails on IE.

src/main/java/org/lantern/http/JettyLauncher.java
myleshorton: Fixed!

src/main/java/org/lantern/http/JettyLauncher.java
myleshorton: Fixed!

src/main/java/org/lantern/http/InteractionServlet.java
oxtoacart: Sorry, I don't get it.  Public sites use XSRF protection all the time.  What exactly is going wrong with XSRF on?

myleshorton: Actually I guess it's really just the CSP header that causes local network issues -- fixed!

oxtoacart: Okay, cool, thanks!

Cheers,
Ox

- Marge Piercy

atavism:
@myleshorton Here's where I'm sticking the local lantern address list. This section only appears when a hello multicast message from a Pi is received and that information is relayed to the front-end. The section is always hidden when the number of local lanterns is equal to zero. Clicking on the link opens the external UI in another tab.

image

src/main/java/org/lantern/multicast/LanternMulticast.java
```diff
@@ -0,0 +1,141 @@
...
+                        ms.receive(recv);
+                        final MulticastMessage msg = 
+                                JsonUtils.decode(new ByteArrayInputStream(buf), 
+                                        MulticastMessage.class);
+                        if (msg.isBye()) {
+                            endpoints.remove(msg.getEndpoint());
+                            Events.sync(SyncPath.LOCAL_LANTERNS, endpoints);
+                        } else {
+                            endpoints.add(msg.getEndpoint());
+                            Events.sync(SyncPath.LOCAL_LANTERNS, endpoints);
```
atavism: Could move the 
Events.sync(SyncPath.LOCAL_LANTERNS, endpoints); 
statement outside the conditional.

src/main/java/org/lantern/LanternUtils.java
```diff
@@ -1371,4 +1371,15 @@ public static boolean isShuttingDown() {
...
+    /**
+     * Utility for whether or not we're running on a Pi. Just checks for ARM 
+     * for now.
+     * 
+     * @return <code>true</code> if we're on ARM, otherwise <code>false</code>.
+     */
+    public static boolean isLanternPi() {
+        return LanternUtils.isDevMode() || 
+                SystemUtils.OS_ARCH.toLowerCase().contains("arm");
+    }
```
atavism: The name of the method is kinda weird since this also returns true if we're running in dev mode.

src/main/java/org/lantern/multicast/LanternMulticast.java
```diff
@@ -0,0 +1,141 @@
...
+            final InetAddress group = InetAddress.getByName(MC_ADDR);
+            final MulticastSocket ms = new MulticastSocket(MC_PORT);
+            ms.joinGroup(group);
+            
+            if (LanternUtils.isLanternPi()) {
+                sendHellos(group, ms);
+                addShutdownHook(group, ms);
+            }
+            
+            listen(ms);
```
atavism: Do we even need to listen on the Pi side? Whenever the "hello" and "bye" messages get sent, we end up receiving these locally as well.

myleshorton: Oh interesting. So we should also check for our own IP and ignore it probably huh? I'll make that change, in which case I think everyone can just listen right?

src/main/java/org/lantern/multicast/LanternMulticast.java
myleshorton: Yup nice one.

src/main/java/org/lantern/LanternUtils.java
myleshorton: Will change.

myleshorton:
I just made those changes, @atavism and cleaned things up a little bit while in there. Thanks for reviewing!

atavism:
@myleshorton Great, thanks. Tested this out, and it's working great. Will submit a corresponding client-side PR. I'm including an updated ARM version of flashlight with this PR. Merging.

myleshorton:
Sweet thanks!

PR #2149, opened 2015-01-08
toggle on and off auto proxy config file through OSX api

fffw:
Uses System Configuration Framework, reduces `configureNetworkServices` execution time from 25~30s to less than 1s for my 35 network services setup.

Now Google login page can be shown in 5s after clicked Sign in button.

Should be back compatible with OS X >= 10.6, but only tested on my 10.10.1 box.


myleshorton:
Looks great at first glance @fffw! I was going to suggest calling the cocoa APIs for prefs directly but very cool you just went ahead and did it! I'll test on my machine tomorrow.

Lantern Network Setup/Lantern Network Setup/main.m
```diff
@@ -1,102 +1,134 @@
...
+            [newPreferences setValue: wantedHost forKey:(NSString*)kSCPropNetProxiesHTTPProxy];
+            [newPreferences setValue:[NSNumber numberWithInt:1] forKey:(NSString*)kSCPropNetProxiesProxyAutoConfigEnable];
+            [newPreferences setValue:autoProxyConfigFileUrl forKey:(NSString*)kSCPropNetProxiesProxyAutoConfigURLString];
+            NSLog(@"Setting Proxy ON with: %@", newPreferences);
+        } else {//Turn proxy configuration OFF
+            [newPreferences setValue:[NSNumber numberWithInt:0] forKey:(NSString*)kSCPropNetProxiesProxyAutoConfigEnable];
+            NSLog(@"Setting Proxy OFF");
+        }
+        
+        success = SCNetworkProtocolSetConfiguration(proxyProtocolRef, (__bridge CFDictionaryRef)newPreferences);
```
oxtoacart: Will this overwrite any changes that other processes made between line 76 and here?

oxtoacart:
Nice!  I tested this on OS X 10.9.5 and it works for me.  The code looks good, with the only potential problem being a race condition with other programs that are also trying to modify network preferences (see comment on line 90).

fffw:
Only one process can gain the access through `SCPreferencesLock()`, so race condition will not be incurred. Pass `TRUE` as second parameter will wait for other process to unlock.

You revealed another issue that I should only commit and apply preferences once, rather that in the loop.

oxtoacart:
Ah, I totally missed that lock code.  Excellent!

oxtoacart:
@fffw This looks great!  And it works in my testing locally.

@myleshorton and I decided that it would be good to take this is a high-priority hotfix, which means that we actually need it in the master branch instead of devel.  Can you please resubmit to master and then I'll merge.

Thanks!

myleshorton:
Hey @oxtoacart and I just chatted about this more @fffw, and it would be better to merge it to the release-1.5.14 branch, as that will be in master soon.

myleshorton:
I ended up manually merging this to release-1.5.14 so will also just merge here to devel.

fffw:
Wow. Hope it works without trouble.

--  
Merlin Ran
Sent with Sparrow (http://www.sparrowmailapp.com/?sig)

myleshorton:
Me too! Working so far, and it's just in beta so hopefully we'll find out if there are any issues.

Issue #2168, opened 2015-01-13
[bug] New problems extracting flashlight


oxtoacart:
Observed on branch issue-2164 while running in Give mode:

```
11762  2015-01-13 08:18:41,210 INFO  [Lantern-Async-Event-Thread-0] lantern.LanternUtils.extractExecutableFromJar (LanternUtils.java:1285) - File /Users/ox.to.a.cart/.lantern/flashlight is out of date, deleting
11764  2015-01-13 08:18:41,212 INFO  [Lantern-Async-Event-Thread-1] lantern.LanternUtils.extractExecutableFromJar (LanternUtils.java:1285) - File /Users/ox.to.a.cart/.lantern/flashlight is out of date, deleting
11845  2015-01-13 08:18:41,293 INFO  [Lantern-Async-Event-Thread-0] lantern.LanternUtils.extractExecutableFromJar (LanternUtils.java:1292) - Moving /var/folders/j_/9dywssj524gf3100s4q9l48w0000gp/T/1421158720098-0/flashlight to /Users/ox.to.a.cart/.lantern/flashlight
11846  2015-01-13 08:18:41,294 INFO  [Lantern-Async-Event-Thread-0] pt.BasePluggableTransport.startClient (BasePluggableTransport.java:125) - Starting Flashlight client
11848  2015-01-13 08:18:41,296 WARN  [Lantern-Async-Event-Thread-1] lantern.LanternUtils.extractExecutableFromJar (LanternUtils.java:1288) - Could not delete old file at /Users/ox.to.a.cart/.lantern/flashlight
11849  2015-01-13 08:18:41,297 INFO  [Lantern-Async-Event-Thread-1] lantern.LanternUtils.extractExecutableFromJar (LanternUtils.java:1292) - Moving /var/folders/j_/9dywssj524gf3100s4q9l48w0000gp/T/1421158719775-0/flashlight to /Users/ox.to.a.cart/.lantern/flashlight
11850  2015-01-13 08:18:41,298 ERROR [Lantern-Async-Event-Thread-1] lantern.LanternUtils.extractExecutableFromJar (LanternUtils.java:1299) - Unable to move file to destination /Users/ox.to.a.cart/.lantern/flashlight: Destination '/Users/ox.to.a.cart/.lantern/flashlight' already exists
[...]

PR #2169, opened 2015-01-13
#2168 Synchronizing extraction of pluggable transports on class to avoid...

oxtoacart:
... race condition in extracting file

### DIFF
--- a/src/main/java/org/lantern/proxy/pt/BasePluggableTransport.java
+++ b/src/main/java/org/lantern/proxy/pt/BasePluggableTransport.java
@@ -65,8 +65,10 @@ public abstract class BasePluggableTransport implements PluggableTransport {
             path += ".exe";
         }
         try {
-            this.exe = LanternUtils.extractExecutableFromJar(path, 
-                    LanternClientConstants.DATA_DIR);
+            synchronized(this.getClass()) {
+                this.exe = LanternUtils.extractExecutableFromJar(path, 
+                        LanternClientConstants.DATA_DIR);
+            }
         } catch (final IOException e) {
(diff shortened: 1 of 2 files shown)


myleshorton:
I would consider using FileLock in LanternUtils instead, as it would solve the problem more generally and would isolate the locking to the file itself.

oxtoacart:
Neat idea.  Do you mean org.littleshoot.util.FileLockUtils  ?

myleshorton:
No java.nio.channels.FileLock directly -- something like


```java
FileOutputStream out = new FileOutputStream(file);
try {
    java.nio.channels.FileLock lock = out.getChannel().lock();
    try {
        // move file in here.
    } finally {
        lock.release();
    }
} finally {
    in.close();
}
```

oxtoacart:
Gotcha.  And this is to prevent problems when running two instances of Lantern, right?

myleshorton:
It's definitely a bit more verbose, but more robust I think too.

myleshorton:
> Gotcha. And this is to prevent problems when running two instances of Lantern, right?

Yeah, I'm basically thinking we just lock the flashlight file, although now that I think about it that'll just lock to the process not the thread -- maybe ignore me.

oxtoacart:
> although now that I think about it that'll just lock to the process not the thread

Correct. From the Javadoc for java.nio.channels.FileLock:

"File locks are held on behalf of the entire Java virtual machine. They are not suitable for controlling access to a file by multiple threads within the same virtual machine."

myleshorton:
OK, let's just go with your solution. I'm always wary of synchronizing on a class, but I can't think of anything better here.

oxtoacart:
Thanks!

PR #2182, opened 2015-01-21
Pointing to new peerscanner

oxtoacart:


src/main/java/org/lantern/Cli.java
```diff
@@ -67,7 +67,7 @@ public Cli(String[] args) {
         }
         catch (final ParseException e) {
             printHelp(options, e.getMessage()+" args: "+Arrays.asList(args));
-            return;
+            System.exit(0);
```
oxtoacart: This was a drive-by fix.  I noticed that if I provided an invalid command-line argument, Lantern would start anyway.  That seemed dangerous/annoying.

lantern-common
```diff
@@ -1 +1 @@
-Subproject commit 3c1f3cbee54d21e08e952eb53b90382c90c69616
+Subproject commit 81565de582437ce8d987ba1b2cc2d24ccf9d9d3c
```
oxtoacart: This captures https://github.com/getlantern/lantern-common/commit/81565de582437ce8d987ba1b2cc2d24ccf9d9d3c

myleshorton:
LGTM!

oxtoacart:
@myleshorton Thanks!

PR #2189, opened 2015-01-23
changed to use network endpoint #2188

myleshorton:
@atavism I realized in testing your changes I was pointing to the totally wrong host! Should be fixed with this PR.


atavism:
Ah, nice catch! Looks good to me.

Issue #2193, opened 2015-01-26
[lantern] Enable WebSockets transport for CometD to save some UI CPU usage by not polling


oxtoacart:

myleshorton:
Closed in e124147

PR #2194, opened 2015-01-26
#2193 Enabled websockets transport in CometD

oxtoacart:
Note - this may not take effect until https://github.com/getlantern/lantern-ui/pull/79 is merged, which includes the client-side change to enable websockets.


myleshorton:
LGTM!

Issue #2198, opened 2015-01-27
[lantern] When connectivity is lost and reestablished, Lantern doesn't notice


oxtoacart:
To reproduce:

1. rm -Rf ~/.lantern
2. Start Lantern
3. Disconnect network (e.g. disable wifi)
4. Wait until you see the spinner in the attached screenshot (this means that Lantern has detected connectivity was lost)
5. Reconnect network
6. Wait as long as you like, the spinner sticks around

![screen shot 2015-01-27 at 12 03 44 pm](https://cloud.githubusercontent.com/assets/5000654/5923907/c9f98158-a61c-11e4-83b5-930f489ece34.png)

myleshorton:
This is on devel right?

oxtoacart:
Yes, on devel.

atavism:
Seeing the same thing. Even when you refresh the UI after re-establishing connectivity, the spinner remains. I'm seeing this over and over in the log:
```
152368 2015-01-28 00:08:25,274 DEBUG [Sync-Exec-Thread] BayeuxServerImpl.447099582.debug (BayeuxServerImpl.java:136) - No authorizers, PUBLISH for channel /sync granted
152368 2015-01-28 00:08:25,274 DEBUG [Sync-Exec-Thread] BayeuxServerImpl.447099582.debug (BayeuxServerImpl.java:136) - <  {id=72, data=[org.lantern.state.CometDSyncStrategy$SyncData@24f2f2], channel=/sync}
152368 2015-01-28 00:08:25,274 DEBUG [Sync-Exec-Thread] BayeuxServerImpl.447099582.debug (BayeuxServerImpl.java:136) - << {id=72, successful=true, channel=/sync}
152368 2015-01-28 00:08:25,274 DEBUG [Sync-Exec-Thread] BayeuxServerImpl.447099582.debug (BayeuxServerImpl.java:136) - <  {id=72, successful=true, channel=/sync}
152368 2015-01-28 00:08:25,274 DEBUG [pool-3-thread-6] server.WebSocketTransport.schedule (AbstractWebSocketTransport.java:508) - Flushing 31hs2toqco2iqs1k8pqnyu894ki - last connect 8871 ms ago timeout=false metaConnectDelivery=false, metaConnectReply={id=7, successful=true, channel=/meta/connect}, messages=[{id=72, data=[org.lantern.state.CometDSyncStrategy$SyncData@24f2f2], channel=/sync}]
[...]

Issue #2196, opened 2015-01-27
[lantern] Friends screen hangs on login


oxtoacart:
I'm testing with the current devel branch of lantern, the 0.0.12 branch of lantern-ui (soon to be merged) and a flashlight built from the peerscanner branch.

When logging in for the first time (in Get mode), I've noticed that the friends screen hangs indefinitely.  Looking at the logs, it looks like some Lantern-internal stuff is trying to proxy through peers.  That shouldn't happen, because our QOS requirement should require running through fallbacks only.

![screen shot 2015-01-27 at 10 03 52 am](https://cloud.githubusercontent.com/assets/5000654/5921857/91a77176-a60c-11e4-99ec-03b60c3b4dc6.png)

oxtoacart:
Part of the problem is that we used to use [GetModeProxyFilter[(https://github.com/getlantern/lantern/blob/4db26a057a74849cfa197d36e8e835233cbdb4eb/src/main/java/org/lantern/proxy/GetModeProxyFilter.java) to temporarily turn on high QOS for everything, but since moving proxying into the Go layer, this class is no longer used.

oxtoacart:
The friends screen definitely does not hang in Give mode.

oxtoacart:
So, I've fixed the QOS issue, but the screen still hangs.  Interestingly, if I restart Lantern, the friends screen comes up very quickly after restart.  This could be because at that point we have a traditional chained fallback available and aren't domain fronting anymore.

oxtoacart:
I tried to see if I could replicate the problem after disabling chained servers, but I couldn't.  Then I reenabled chained servers, and I still can't replicate it.  So, it must not hang all the time, just sometimes.

oxtoacart:
When it does hang, I see from the log that we never even got into the friendsLoader Callable (i.e. the message "Loading friends" doesn't appear).  So it looks like it's not a problem with the friends API call itself, but something before that.

oxtoacart:
Okay, I think this may be a race condition in PublicIpAndTokenTracker.  I added some logging to it to show when a PublicIP is received, a refresh token is received, and when the PublicIpAndTokenEvent is raised.  Usually everything looks fine, but when the friends screen hangs, I see this in the log:

```
6154   2015-01-27 11:39:43,828 DEBUG [Lantern-Async-Event-Thread-5] event.PublicIpAndTokenTracker.onPublicIp (PublicIpAndTokenTracker.java:60) - Got public IP
[...]

PR #2199, opened 2015-01-27
#2196 Updated QOS settings to work with new flashlight, fixed race condi...

oxtoacart:
...tion in PublicIpAndTokenTracker

Requires this flashlight pull (binaries already included in here) - https://github.com/getlantern/flashlight-build/pull/37


src/main/java/org/lantern/event/PublicIpAndTokenTracker.java
```diff
@@ -10,58 +11,52 @@
...
+
     @Subscribe
-    public void onPublicIp(final PublicIpEvent pce) {
-        synchronized (this) {
-            this.hasPublicIp = true;
-            if (this.hasRefresh) {
-                log.debug("Posting event!!");
-                Events.asyncEventBus().post(new PublicIpAndTokenEvent(this.refresh));
-            }
+    synchronized public void onConnectivityChanged(ConnectivityChangedEvent cce) {
```
oxtoacart: This is the crux of it.  Instead of having a reset method call, we look for ConnectivityChanged in order to clear out the IP.

myleshorton: Isn't the effect the same since reset was only called in response to a connectivity changed event? Granted the reset call should be synchronized, but that seems like the only issue with the previous code.

oxtoacart: No, the issue is that reset was a direct call, whereas the PublicIpEvent was sent asynchronously via the EventBus.  Depending on how quickly the event is delivered, it might arrive before or after the reset call.  By removing reset and instead relying on the ConnectivityChanged event, we're letting the EventBus handle everything, and I'm assuming that the EventBus delivers messages in order.

myleshorton: So we actually have the Events.inOrderAsyncEventBus for making sure events are in order, which we should use if we need that. Beyond that, though, I wonder if there's an associated race condition in PublicIpInfoHandler. In particular, in seems like the only way in practice the public IP should get delivered before the reset is if it's cached. Any connectivitychangedevent should clear the cache, which it does, but it's not thread safe.

oxtoacart: So, PublicIpInfoHandler does the public ip check in response to a ProxyConnectionEvent, which happens as soon as the local proxy has started.

The reset call happens in startNetworkServices, which is called in response to a ConnectivityChangedEvent.  So I guess really the race here is between the connectivity check and the local proxy coming up.

In any case, making PublicIpAndTokenTracker only lose the stored IP in response to lost connectivity should give us what we need.  Previously, we would have stopped network services on the connectivity lost, and then cleared out the IP on connectivity being reestablished.  That has its own race between the connectivity event being delivered and the public ip event showing up.

I should note that if you look at the ticket, you'll see logging that clearly shows reset being called after the public IP has been captured, confirming that this is the root cause.  I should also note in my testing, with my change, I haven't been to reproduce the hang on the friends screen.

I can't test the scenario of connectivity being lost and reestablished because of #2198.

myleshorton: > So, PublicIpInfoHandler does the public ip check in response to a ProxyConnectionEvent, which happens as soon as the local proxy has started.

OK, so this is a related problem. With the new code on devel, there will only ever be a ProxyConnectionEvent once at startup. That means from that point on, we'll use that same public IP even if we change networks. It also means we'll never again generate a PublicIpEvent, which leads to other problems. So there's a bug with not learning about new IPs if we change networks for sure.

With the code on master, this doesn't happen, and reset() also always gets called before DefaultProxyTracker.init(), which means that reset always gets called before a PublicIpEvent is generated (since DefaultProxyTracker generates that event via generating a ProxyConnectionEvent). 

So it looks like both of these slipped through the cracks of the #2103 PR. In theory we could probably just make PublicIpInfoHandler actually do the new check on a ConnectivityChangedEvent for becoming connected instead of on ProxyConnectionEvent.

In terms of PublicIpAndTokenTracker the problem is now that it will never get a public IP after that first run. So when we get a ConnectivityChangedEvent for disconnecting, it will set hasPublicIp to false correctly, but PublicIpAndTokenTracker will never again get a PublicIpEvent because those are only generated in response to connections to the proxy. So the PublicIpAndTokenEvent will never get generated, and the friends screen or anything else waiting on that will hang.

I *think* just looking up the public IP in response to ConnectivityChanged might do it, but we have to somehow be sure the GetModeProxy has in fact started and is able to do the lookup.

oxtoacart: @myleshorton Nice debugging!

> OK, so this is a related problem. With the new code on devel, there will only ever be a ProxyConnectionEvent once at startup. That means from that point on, we'll use that same public IP even if we change networks. It also means we'll never again generate a PublicIpEvent, which leads to other problems. So there's a bug with not learning about new IPs if we change networks for sure.

Ahh, that makes sense.

> With the code on master, this doesn't happen, and reset() also always gets called before DefaultProxyTracker.init(), which means that reset always gets called before a PublicIpEvent is generated (since DefaultProxyTracker generates that event via generating a ProxyConnectionEvent).

My what a tangled web we wove!  I'm glad that we've been able to avoid a similar event spaghetti in flashlight and hope that we'll continue to do so.

> So it looks like both of these slipped through the cracks of the #2103 PR.

Yup, though given the complexity of the involved events I wouldn't really expect otherwise.  This is where QA'ing new releases comes in handy, and just in general getting all of us to run Lantern more frequently, as the manifestation of this particular problem is pretty obvious.

> In theory we could probably just make PublicIpInfoHandler actually do the new check on a ConnectivityChangedEvent for becoming connected instead of on ProxyConnectionEvent.

Well, it can't check until it's gotten a ProxyConnectionEvent, because it uses the proxy to get the IP.  However, I do think that driving it primarily by ConnectivityChangedEvent makes sense, just with the added requirement that ProxyConnectionEvent has happened at least once.

> In terms of PublicIpAndTokenTracker the problem is now that it will never get a public IP after that first run. So when we get a ConnectivityChangedEvent for disconnecting, it will set hasPublicIp to false correctly, but PublicIpAndTokenTracker will never again get a PublicIpEvent because those are only generated in response to connections to the proxy. So the PublicIpAndTokenEvent will never get generated, and the friends screen or anything else waiting on that will hang.

Makes sense.

> I think just looking up the public IP in response to ConnectivityChanged might do it, but we have to somehow be sure the GetModeProxy has in fact started and is able to do the lookup.

We can do what we do in PublicIpAndTokenTracker and wait on both conditions.  I'll take a crack at this.  I guess in order to test this we'll have to also resolve #2198, which I may or may not be able to do before I leave on vacation.

oxtoacart:
@myleshorton Okay, I've made updates along the lines of what we discussed and it seems to be working well, including fixing #2198 too.

There's a lantern-ui change now too - https://github.com/getlantern/lantern-ui/pull/81

oxtoacart:
@myleshorton Found one more issue related to starting when there's no internet connectivity.  Will fix shortly.

oxtoacart:
@myleshorton Drats, there's a chicken and egg problem here.  Flashlight needs its country code at startup, but that's only available by doing a GeoIP lookup, which in turn requires flashlight to be started.

oxtoacart:
@myleshorton Okay, fixed the chicken and egg problem.  Should be good now.

src/main/java/org/lantern/AppIndicatorTray.java
```diff
@@ -258,10 +257,9 @@ public boolean isActive() {
     }
 
     @Subscribe
-    public void onConnectivityStateChanged(final ProxyConnectionEvent csce) {
-        final ConnectivityStatus cs = csce.getConnectivityStatus();
-        LOG.debug("Got connectivity state changed {}", cs);
-        onConnectivityStatus(cs);
+    public void onConnectivityStateChanged(final ProxyConnectedEvent csce) {
+        LOG.debug("Got ProxyConnectedEvent");
+        onConnectivityStatus(ConnectivityStatus.CONNECTED);
```
oxtoacart: We no longer have different proxy connectivity statuses.  Once the proxy is connected, it is connected.

src/main/java/org/lantern/Proxifier.java
```diff
@@ -138,30 +138,11 @@ public synchronized void onSystemProxyChanged(
...
-            } catch (final ProxyConfigurationError e) {
-                LOG.warn("Could not unproxy?", e);
-            }
-            break;
-        default:
-            break;
-        
-        }
+        LOG.debug("Got connected event");
+        proxyWithChecks();
```
oxtoacart: Again, once the proxy is connected, it's connected.  We never learn about disconnects anymore.

src/main/java/org/lantern/PublicIpInfoHandler.java
```diff
@@ -46,72 +42,66 @@ public PublicIpInfoHandler(final Model model, final Censored censored,
...
-    /**
-     * We perform the public IP lookup on a proxy connection event because
-     * we need the proxy in order to perform the lookup. We need this both
-     * at startup as well as every time we re-connect to a proxy after
-     * potentially losing proxy connectivity.
-     * 
-     * @param pce The connection event.
-     */
-    @Subscribe
-    public void onProxyConnectionEvent(
```
oxtoacart: This dependency from PublicIp to ProxyConnection was really outdated, since we've been using a host-spoofed Get to geoip for a while now.  This realization simplified PublicIpInfoHandler quite a bit, since we can now just check every time that connectivity is established/restored.

src/main/java/org/lantern/PublicIpInfoHandler.java
```diff
@@ -46,72 +42,66 @@ public PublicIpInfoHandler(final Model model, final Censored censored,
...
-        if (!this.recheck.get()) {
-            log.debug("Not rechecking for public IP");
-            return;
+        InetAddress address = null;
+        for (int i = 0; i < 10; i++) {
+            // Try a few times to get the public ip address
+            // We do this because for some reason, if we've just restored
+            // connectivity and gotten a connected event, connecting to the
+            // geo service to do an IP lookup sometimes fails.
+            try {
```
oxtoacart: This was a little weird, but I saw that we would get Connectivity events but then fail to connect to geoip to do the public ip lookup.  Doing this retry with exponential backoff thing seems to do the trick.

src/main/java/org/lantern/SystemTrayImpl.java
```diff
@@ -269,14 +269,13 @@ public void onUpdate(final UpdateEvent update) {
...
-        final ConnectivityStatus cs = csce.getConnectivityStatus();
-        log.debug("Received connectivity state changed: {}", cs);
+    public void onConnectivityStateChanged(final ProxyConnectedEvent csce) {
+        log.debug("Proxy connected");
         if (!this.model.getSettings().isUiEnabled()) {
             log.info("Ignoring event with UI disabled");
             return;
         }
-        onConnectivityStatus(cs);
+        onConnectivityStatus(ConnectivityStatus.CONNECTED);
```
oxtoacart: Again, once the proxy is connected, that's it.  We don't learn about any other statuses for proxies anymore.

src/main/java/org/lantern/proxy/GetModeProxy.java
```diff
@@ -21,37 +21,81 @@
...
+        synchronized (this) {
+            started = this.started;
+            hasPublicIp = this.hasPublicIp = true;
+        }
+
+        startWhenReady(started, hasPublicIp);
     }
-    
+
+    private void startWhenReady(boolean started, boolean hasPublicIp) {
```
oxtoacart: We can't actually start the GetModeProxy until we have a PublicIp, since we need to do a geo lookup and pass the country to flashlight on startup.

src/main/java/org/lantern/proxy/GetModeProxy.java
```diff
@@ -21,37 +21,81 @@
...
-        fl.startStandaloneClient();
-        fl.addFallbackProxies(model.getS3Config().getFallbacks());
-        Events.asyncEventBus().post(new ProxyConnectionEvent(ConnectivityStatus.CONNECTED));
+    public void start() {
+        boolean started = false;
+        boolean hasPublicIp = false;
+        synchronized (this) {
+            started = this.started = true;
+            hasPublicIp = this.hasPublicIp;
+        }
```
myleshorton: I would just use AtomicBoolean and skip all this synchronization and local variables.

src/main/java/org/lantern/proxy/GetModeProxy.java
```diff
@@ -21,37 +21,81 @@
...
+    }
+
+    @Subscribe
+    public void onPublicIp(PublicIpEvent pie) {
+        boolean started = false;
+        boolean hasPublicIp = false;
+        synchronized (this) {
+            started = this.started;
+            hasPublicIp = this.hasPublicIp = true;
+        }
```
myleshorton: Same as above.

src/main/java/org/lantern/proxy/GetModeProxy.java
myleshorton: Or you could just make both those methods synchronized. Then you don't need volatile or the local variables but make sure the two stay in sync with no race condition.

oxtoacart: I didn't make the whole methods synchronized because I don't want the delivery of the event blocking on starting flashlight.  I just switched to using AtomicBooleans.

oxtoacart: Actually, as you say, using atomic booleans could have a race condition.

Other than the unnecessary volatile declarations, is there anything wrong with the code as written?

myleshorton: > I didn't make the whole methods synchronized because I don't want the delivery of the event blocking on starting flashlight.

That won't matter because the event is delivered asynchronously on its own thread.

> Other than the unnecessary volatile declarations, is there anything wrong with the code as written?

It looks technically correct but just unnecessarily verbose.

oxtoacart: >> I didn't make the whole methods synchronized because I don't want the delivery of the event blocking on starting flashlight.

> That won't matter because the event is delivered asynchronously on its own thread.

Okay, I went ahead and made the whole methods synchronized.

src/main/java/org/lantern/event/PublicIpAndTokenTracker.java
```diff
@@ -10,58 +11,52 @@
...
 @Singleton
 public class PublicIpAndTokenTracker {
 
     private final Logger log = LoggerFactory.getLogger(getClass());
-    
-    private boolean hasRefresh = false;
 
-    private boolean hasPublicIp = false;
+    private volatile boolean hasPublicIp = false;
+    private volatile String refresh;
```
myleshorton: Thanks on the synchronized change! You also technically don't need these to be volatile, but not a huge deal either way.

oxtoacart: Yeah, I figured I'd leave them volatile in case anyone ever adds code that reads them for any reason.

myleshorton: OK cool

myleshorton:
OK LGTM. Merging.

oxtoacart:
Thanks for reviewing!

PR #2424, opened 2015-04-21
Route HTTP traffic to trusted proxies only

xiam:
This PR adds support for routing plaintext HTTP traffic only to proxies deployed by Lantern, in this moment it may not represent any difference.

HTTPs can travel safely along trusted and untrusted proxies without compromising user's data, as the protocol itself provides point to point encryption (well, point to Trusted Third Parties and eventually to point, but that's outside our scope).

A new property `trusted` was added to our configuration settings, this feature depends on whether this flag is set true or not. If no servers marked as `trusted` are found Lantern will try to use any server (getting unsafe access is better than having no access at all).

Routing an HTTP URL would produce a log like:

```
curl -x 127.0.0.1:8787 http://www.google.com/humans.txt

Apr 21 19:56:19.621 - balancer: balancer.go:100 Dialing tcp://www.google.com:80 with (trusted) fronted proxy at nl.fallbacks.getiantem.org:443 using masquerade set cloudflare
```

While routing an HTTPs URL would look like:

```
curl -x 127.0.0.1:8787 https://www.google.com/humans.txt

Apr 21 19:57:05.401 - balancer: balancer.go:100 Dialing tcp://s3.amazonaws.com:443 with chained proxy at 178.62.243.211:443
```

You'll probably need to edit `lantern.yaml` manually to set the `trusted` field on fronted and chained proxies.

I also added many comments on the code that would make it easier to understand what is each part trying to do and fixed a few minor bugs.

See our discussion at: https://github.com/getlantern/lantern/issues/2290


src/github.com/getlantern/balancer/balancer.go
```diff
@@ -22,23 +22,42 @@ var (
 // Balancer balances connections established by one or more Dialers.
 type Balancer struct {
 	dialers []*dialer
+	trusted []*dialer
```
xiam: This property will be nil until calling `trustedDialers`.

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -49,7 +68,25 @@ func New(dialers ...*Dialer) *Balancer {
 // remaining Dialers until it either manages to connect, or runs out of dialers
 // in which case it returns an error.
 func (b *Balancer) DialQOS(network, addr string, targetQOS int) (net.Conn, error) {
```
xiam: This is the only function that returns a `net.Conn`, so it looks like a good place to check if the request is unencrypted and take action.

src/github.com/getlantern/chained/server.go
```diff
@@ -69,7 +69,7 @@ func (s *Server) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 		wg.Done()
 	}()
 	go func() {
-		go io.Copy(connIn, connOut)
```
xiam: The `wg.Wait()` was not actually doing nothing as the `io.Copy` was spawned into a goroutine. Now the `wg.Wait()` will actually wait that both calls to `io.Copy()` are completed before returning control to the parent function.

src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -7,40 +7,70 @@ import (
...
 	for _, s := range cfg.FrontedServers {
+		// Getting a dialer for domain fronting and a dialer to dial to arbitrary
+		// addreses.
 		fd, dialer := s.dialer(cfg.MasqueradeSets)
 		dialers = append(dialers, dialer)
 		if dialer.QOS > highestQOS {
+			// If this dialer as a higher QOS than our current highestQOS, set it as
+			// the highestQOSFrontedDialer.
 			highestQOSFrontedDialer = fd
+			highestQOS = dialer.QOS
```
xiam: We were missing a line that updated `highestQOS` to the greatest known value of `dialer.QOS` in order to actually get the highest QOS.

src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -7,40 +7,70 @@ import (
...
 			log.Debugf("Unable to configure chained server for %s: %s", s.Addr)
 		}
-
 	}
 
+	// Creating a balancer with all of our available dialers.
 	bal := balancer.New(dialers...)
 
+	// Locking balCh.
+	client.balChMu.Lock()
```
xiam: There's a small change that both `initBalancer` and `getBalancer` would push something to `balCh`. Since `balCh` is a channel of just one element it would block. It would be a very rare condition to reproduce.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -11,18 +11,20 @@ import (
 )
 
 const (
-	CONNECT        = "CONNECT" // HTTP CONNECT method
-	XFlashlightQOS = "X-Flashlight-QOS"
```
xiam: No need to export those constants.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -32,39 +34,59 @@ func (client *Client) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
+	if req.Method != httpConnectMethod {
 		panic("Intercept used for non-CONNECT request!")
 	}
 
-	// Hijack underlying connection
-	clientConn, _, err := resp.(http.Hijacker).Hijack()
-	if err != nil {
+	// Hijacking underlying connection.
+	var clientConn net.Conn
+	if clientConn, _, err = resp.(http.Hijacker).Hijack(); err != nil {
```
xiam: Using this error declaration pattern and comparison is great for quickly identifying which error are we actually referring to.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -78,51 +100,41 @@ func (client *Client) targetQOS(req *http.Request) int {
...
+	err = resp.Write(w)
 	if err == nil {
 		_, err = w.Write([]byte(msg))
 	}
 	return err
 }
-
-// hostIncludingPort extracts the host:port from a request.  It fills in a
-// a default port if none was found in the request.
-func hostIncludingPort(req *http.Request, defaultPort int) string {
```
xiam: We don't really need this as a separate function. The intention of this code is more evident if we move it to the context in which it's actually being used.

src/github.com/getlantern/flashlight/client/reverseproxy.go
```diff
@@ -38,22 +51,38 @@ func (client *Client) initReverseProxy(bal *balancer.Balancer, dumpHeaders bool)
...
 	log.Trace("Publishing reverse proxy")
+
+	// getReverseProxy() will be unblocked after this.
 	client.rpCh <- rp
+
+	// Unlocking rpCh.
+	client.rpChMu.Unlock()
+
+	// Setting the rpInitialized flag.
+	client.rpInitialized = true
```
xiam: We were not actually recording if this was initialized.

aranhoide:
I'll take a look at this.

aranhoide:
> (getting unsafe access is better than having no access at all)

I think this depends on the user's expectations and usage.  I think that approach does work for the typical usage of Lantern.  But if we start promising that we'll _never_ proxy plain HTTP through untrusted peers, defaulting on no access at all (or modifying the request to try HTTPS instead) may actually be wiser.

In principle this issue should never come up because we should always have trusted servers [1], but what to default on in such an extreme case is an interesting question.

[1] I'll upload new cloud configurations before merging this, because adding properties to servers there is backwards compatible.

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -22,23 +22,42 @@ var (
...
 		dialers: dhs,
 	}
 }
 
+// trustedDialers returns the subset of b.dialers that are considered as
+// trusted.
+func (b *Balancer) trustedDialers() []*dialer {
+	if b.trusted == nil {
+		b.trusted = make([]*dialer, 0, len(b.dialers))
+		// Lazy initialization of trusted dialers.
```
aranhoide: There is no synchronisation here, so if a second goroutine calls `trustedDialers` while the first call is at this point, they'll get an initially empty `b.trusted`, which will get filled afterwards while they're possibly using it.

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -49,7 +68,25 @@ func New(dialers ...*Dialer) *Balancer {
 // remaining Dialers until it either manages to connect, or runs out of dialers
 // in which case it returns an error.
 func (b *Balancer) DialQOS(network, addr string, targetQOS int) (net.Conn, error) {
-	dialers := b.dialers
+	var dialers []*dialer
+
+	// Checking destination port.
+	_, port, _ := net.SplitHostPort(addr)
+
+	// Are we attempting to connect to port 80 (plain HTTP)?
```
aranhoide: We actually support a few ports besides 80 and 443.  I don't know which of those are encrypted, or whether they pose any threat at all, but I thought I'd point it out:

https://github.com/getlantern/lantern/blob/9363758350d904e61753bc0e99a9f1edc49ec2b2/src/github.com/getlantern/flashlight/flashlight.go#L269

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -49,7 +68,25 @@ func New(dialers ...*Dialer) *Balancer {
...
+	if port == "" || port == "80" {
+		// Then try to use only a trusted dialer.
+		dialers = b.trustedDialers()
+		// Unless we don't have any...
```
aranhoide: See discussion in the PR.  Depending on how we communicate with our users about this, we might actually prefer to not proxy at all in this case.

src/github.com/getlantern/chained/dialer.go
```diff
@@ -88,7 +89,7 @@ func checkCONNECTResponse(r *bufio.Reader, req *http.Request) error {
 	if err != nil {
 		return fmt.Errorf("Error reading CONNECT response: %s", err)
 	}
-	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+	if resp.StatusCode < http.StatusOK || resp.StatusCode > 299 {
```
aranhoide: If you consider `200` a magic number here you may want to say go`http.StatusOK+99` instead of `299`.

src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -7,40 +7,70 @@ import (
...
 )
 
+// getBalancer waits for a message from client.balCh to arrive and then it
+// writes it back to client.balCh before returning it as a value. This way we
+// always have a balancer at client.balCh and, if we don't have one, it would
+// block until one arrives.
 func (client *Client) getBalancer() *balancer.Balancer {
+
+	// This balChMu will protect balCh and ensure it always have, at most, one
+	// element enqueued.
```
aranhoide: Why do we want to ensure that?

src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
```diff
@@ -17,6 +18,7 @@ client:
       pipelined: true
       weight: 1000000
       qos: 10{{end}}
+      trusted: false
```
aranhoide: Why are chained servers untrusted?  At the moment all of them are run by us.  When we introduce chained peers (and if we add them to the cloud.yaml as opposed to them being discovered only by kaleidoscope) we will need to make sure to mark those as untrusted.

src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -7,40 +7,70 @@ import (
...
 	client.balCh <- bal
+
 	return bal
 }
 
+// initBalancer takes hosts from cfg.FrontedServers and cfg.ChainedServers and
+// it uses them to create a balancer. It also looks for the highest QOS dialer
+// available among the fronted servers.
 func (client *Client) initBalancer(cfg *ClientConfig) (*balancer.Balancer, fronted.Dialer) {
+	var highestQOSFrontedDialer fronted.Dialer
```
aranhoide: Why move this up here and not `highestQOS`?

src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -7,40 +7,70 @@ import (
...
+
+	// The dialers slice must be large enough to handle all fronted and chained
+	// servers.
```
aranhoide: Does it really must, or is that just an optimisation?

src/github.com/getlantern/flashlight/client/balancer.go
aranhoide: Good catch!

src/github.com/getlantern/flashlight/client/balancer.go
aranhoide: What does need protection is the checking and setting of `client.balInitialized`, unless we are otherwise making sure that two concurrent calls to `getBalancer` will never happen (and if we're assuming that, it'd be nice to have it documented).  

Barring that, I don't think two concurrent writes to `client.balCh` could ever happen, since both `initBalancer` and `getBalancer` will receive from that channel before sending again.  The second one to try such a receive-and-send op would get blocked on the receive.

src/github.com/getlantern/chained/server.go
aranhoide: Good catch!  I wonder what ill effects this could be causing.

src/github.com/getlantern/flashlight/client/chained.go
```diff
@@ -82,10 +87,17 @@ func (s *ChainedServerInfo) Dialer() (*balancer.Dialer, error) {
 	}
 	d := chained.NewDialer(ccfg)
 
+	// Is this a trusted proxy that we could use for HTTP traffic?
+	var trusted string
+	if s.Trusted {
```
aranhoide: For clarity I'd make this "untrusted " if `s.Trusted` is false.

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -39,66 +39,90 @@ type Client struct {
...
-// onListening is a callback that gets invoked as soon as the server is
-// accepting TCP connections.
-func (client *Client) ListenAndServe(onListening func()) error {
-	l, err := net.Listen("tcp", client.Addr)
-	if err != nil {
-		return fmt.Errorf("Client unable to listen at %v: %v", client.Addr, err)
+// ListenAndServe makes the client listen for HTTP connections.  onListeningFn
+// is a callback that gets invoked as soon as the server is accepting TCP
+// connections.
+func (client *Client) ListenAndServe(onListeningFn func()) (err error) {
```
aranhoide: Why make it a named return value if it's never returned by name?  Doesn't make the code any shorter either, because it makes you declare `l`.

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -39,66 +39,90 @@ type Client struct {
...
 	}
-	client.l = l
 
-	onListening()
+	// We opened the port successfully at this point, copying listener to our
+	// client and executing the passed callback.
+	client.l = l
+	onListeningFn()
 
+	// Creating an HTTP server.
```
aranhoide: TBH all the comments in this body, especially this one, are just restating what the code says.  I find that more distracting than helpful.

src/github.com/getlantern/flashlight/client/fronted.go
```diff
@@ -54,9 +54,15 @@ type FrontedServerInfo struct {
 	// QOS: relative quality of service offered. Should be >= 0, with higher
 	// values indicating higher QOS.
 	QOS int
+
+	// Trusted: Determines if a host can be trusted with unencrypted HTTP
+	// traffic.
+	Trusted bool
 }
 
+// dialer
```
aranhoide: Did you mean to write a docstring here?

src/github.com/getlantern/flashlight/client/fronted.go
```diff
@@ -71,22 +77,34 @@ func (s *FrontedServerInfo) dialer(masqueradeSets map[string][]*fronted.Masquera
...
-		Label:  fmt.Sprintf("fronted proxy at %s:%d%s", s.Host, s.Port, masqueradeQualifier),
-		Weight: s.Weight,
-		QOS:    s.QOS,
-		Dial:   fd.Dial,
+
+	// Is this a trusted proxy that we could use for HTTP traffic?
+	var trusted string
+	if s.Trusted {
+		trusted = "(trusted) "
+	}
```
aranhoide: As said above, maybe it would be clearer to explicitly state "untrusted " vs "trusted " instead of assuming the former by default.

src/github.com/getlantern/flashlight/client/fronted.go
```diff
@@ -71,22 +77,34 @@ func (s *FrontedServerInfo) dialer(masqueradeSets map[string][]*fronted.Masquera
...
-			err := fd.Close()
-			if err != nil {
-				log.Debugf("Unable to close fronted dialer: %s", err)
+			if err := fd.Close(); err != nil {
+				log.Debugf("Unable to close fronted dialer: %q", err)
 			}
 		},
 	}
+
+	return fd, bal
```
aranhoide: This would perhaps be a better case for named return values.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -11,18 +11,20 @@ import (
...
+	httpXFlashlightQOS = "X-Flashlight-QOS"
 )
 
 // ServeHTTP implements the method from interface http.Handler using the latest
 // handler available from getHandler() and latest ReverseProxy available from
 // getReverseProxy().
 func (client *Client) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
-	if req.Method == CONNECT {
+	if req.Method == httpConnectMethod {
+		// CONNECT requests are often used for HTTPs requests.
```
aranhoide: I've always seen the acronym spelled HTTPS, with a capital S.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -32,39 +34,59 @@ func (client *Client) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
+
+	// intercept can only by used for CONNECT requests.
+	if req.Method != httpConnectMethod {
 		panic("Intercept used for non-CONNECT request!")
 	}
 
-	// Hijack underlying connection
-	clientConn, _, err := resp.(http.Hijacker).Hijack()
-	if err != nil {
+	// Hijacking underlying connection.
```
aranhoide: Nit: I think the imperative "hijack" is more idiomatic for comments.  I wouldn't have bothered to make this remark on a new comment, but I think changing things should only be done if it's clearly for the better.

src/github.com/getlantern/flashlight/client/handler.go
aranhoide: I think Go makes a special case for error variables so you don't have to declare the others?

src/github.com/getlantern/flashlight/client/handler.go
aranhoide: I actually prefer it as a separate function.  Then it can more easily be documented (because it's more obvious to what scope the comment applies), it can be tested on its own, its internal variables don't pollute the namespace of the containing function and their scope is more contained.

Again, if you had written this inlined in first place I wouldn't have bothered to suggest that you extract it as a function.   But I think we should exercise more restraint in changing things only to match our preferred style, unless it is in preparation for other changes or to follow some guideline we have agreed upon as a team.  Otherwise someone else may (e.g.) extract this as separate function next month, and then all the reward for the work of both programmers and reviewers would have been noise in git history.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -32,39 +34,59 @@ func (client *Client) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
+	// If no port is given, assuming it's 443 for HTTPs.
+	if port == "" {
+		port = "443"
+	}
+
+	// Creating a network address.
+	addr := host + ":" + port
 
-	// Establish outbound connection
+	// Establishing outbound connection with the given address.
```
aranhoide: Again, I find the old comment had more meaning to ink ratio.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -78,51 +100,41 @@ func (client *Client) targetQOS(req *http.Request) int {
...
+		StatusCode: http.StatusOK,
 		ProtoMajor: 1,
 		ProtoMinor: 1,
 	}
+
 	return resp.Write(writer)
 }
 
-func respondBadGateway(w io.Writer, msg string) error {
+func respondBadGateway(w io.Writer, msg string) (err error) {
```
aranhoide: Never returned by name.

src/github.com/getlantern/flashlight/client/reverseproxy.go
aranhoide: Good catch again!

src/github.com/getlantern/flashlight/client/reverseproxy.go
```diff
@@ -10,12 +10,25 @@ import (
...
+// getReverseProxy waits for a message from client.rpCh to arrive and then it
+// writes it back to client.rpCh before returning it as a value. This way we
+// always have a balancer at client.rpCh and, if we don't have one, it would
+// block until one arrives.
 func (client *Client) getReverseProxy() *httputil.ReverseProxy {
+
+	// This rpChMu will protect rpCh and ensure it always have, at most, one
+	// element enqueued.
+	client.rpChMu.Lock()
+	defer client.rpChMu.Unlock()
```
aranhoide: I think this is also unnecessary unless we have two concurrent initializations (and in that case, that should be prevented explicitly).

aranhoide:
Done with this pass of reviewing.  Summary:

- The change itself looks sound, except for a race condition.

- A few errors were spotted (thanks!).

- The added docstrings are really nice, thanks.

- Some of the other comments I find excessive, some of them just restating what the code says clearly enough.  As a general rule (which, like all rules, may be broken judiciously) comments other than docstrings should explain not so much what the code is doing but rather why it's doing that.

- I don't think the added mutexes around channels are necessary.  If anything, we should either protect the initialization of the channels themselves, or (slightly more dangerous) just clearly document our assumption that everything that is called under `Configure` happens serially with respect to other `Configure` calls, and thus can never be called more than once concurrently.

src/github.com/getlantern/balancer/balancer.go
xiam: Good catch!

src/github.com/getlantern/balancer/balancer.go
xiam: The `8080` looks like a really good candidate here, but I'm not sure about the other ports, `5222` and `5223` are commonly used for XMPP and `5222` is the plaintext one so maybe we should add it too. And what about `5228`? Why does Lantern support it?

One thing to keep in mind is that we cannot guarantee that Lantern will identify and protect unencrypted traffic automatically: users may as well use plain HTTP over port 443 on specially crafted servers and Lantern will assume that if it's using port 433 it should be SSL, also using port 8443 does not guarantee that the connection will be over SSL, maybe we can start by only proxying ports such as 80 and 8080 on our servers since that would fix the potential Javascript script injection we're trying to close, then we could discuss about supporting plaintext traffic over other ports.

src/github.com/getlantern/balancer/balancer.go
xiam: I agree with you, this would be a very rare thing to happen. I personally would prefer having no connection at all in such cases but users may have a different opinion. I agree we should discuss this further.

src/github.com/getlantern/balancer/balancer.go
myleshorton: I can't remember offhand why we opened 5228, but it was basically because people were using it.

I don't think we're making any assumptions about the actual data being run over any of these ports are we? In fact we don't add any extra encryption to any traffic other than always encrypting connections between the Lantern client and any proxy it hits (with that resulting in TLS over tunneled TLS in the case of connections that are already encrypted).

I guess I'm a little confused. The theoretical attack here is a mixed content attack. Most browsers actually protect against that for JavaScript (see https://qualys.jive-mobile.com/#jive-document?content=%2Fapi%2Fcore%2Fv2%2Fposts%2F2400). The attack is specifically when browsers load HTTPS pages that include HTTP content. Let's say that's a script (let's imagine the browser isn't protecting against script injections). In that case, though, the javascript would be loaded over HTTP and port 80 typically, so if we proxied port 80 we'd still be vulnerable to that.

I.e. only proxying port 80 on our servers wouldn't mitigate the attack since the attack is typically over port 80.

What am I missing?

src/github.com/getlantern/chained/dialer.go
xiam: I actually think that would be confusing too, since reading the `StatusOK` constant plus a number looks like trying to add a number to something that means a status response, and we're comparing the ordering of status codes with integers, which is already awkward. 

Either way I don't have a strong feeling on changing `200` by `http.StatusOK` in this particular case, since the `<` comparison is already a trick based on the actual integer value of `http.StatusOK`. Maybe these kind of constants are better suited for (in)equalities. So, if you feel this is more awkward than it was we may as well use the previous comparison.

src/github.com/getlantern/flashlight/client/balancer.go
xiam: Because we're declaring balCh as `client.balCh = make(chan *balancer.Balancer, 1)`, so if we attempt to write more that one messages to that channel it would block.

src/github.com/getlantern/balancer/balancer.go
myleshorton: Actually why not just avoid this extra state and create the trusted dialers as soon as you create the Balancer? Like in New just do something like:

```go
    trusted := make([]*dialer, 0, 0)
    untrusted := make([]*dialer, 0, 0)
    for _, d := range dialers {
        dl := &dialer{Dialer: d}
        dl.start()
        if d.Trusted {
            trusted = append(trusted, dl)
        } else {
            untrusted = append(untrusted, dl)
 	}
    }
```

src/github.com/getlantern/flashlight/client/balancer.go
xiam: I though it would make more sense leaving `highestQOS` in which it is now, since it's only used on that portion of the code, while `highestQOSFrontedDialer` is used there and at the end.

src/github.com/getlantern/flashlight/client/balancer.go
xiam: > What does need protection is the checking and setting of client.balInitialized, unless we are otherwise making sure that two concurrent calls to getBalancer will never happen (and if we're assuming that, it'd be nice to have it documented).

Good catch, I'll move `balInitialize` inside the safe zone provided by the mutex.

>  I don't think two concurrent writes to client.balCh could ever happen, since both initBalancer and getBalancer will receive from that channel before sending again. The second one to try such a receive-and-send op would get blocked on the receive.

Ah! Yeah, I see what you mean, I though of the (hypothetical) condition immediately after `old := <-client.balCh` (within this function) and `bal := <-client.balCh` (within `getBalancer()`), but after carefully reviewing it again I'm also convinced that escenario would not happen. Do you think this other case could be a risk? (see 1 and 2 like two different goroutine contexts)

```
1: bal := <-client.balCh // Within getBalancer().
2: old := <-client.balCh // Would block until a message is received.
1: client.balCh <- bal // This is getBalancer() putting back the message.
2: old := <-client.balCh // This will now continue, note that old is equal to bal
1: return bal // getBalancer() ending
2: old.Close() // Will close the bal that getBalancer() already returned.
```

src/github.com/getlantern/balancer/balancer.go
xiam: I was initially inclined to do that too, but I discovered e a couple of issues with that approach:

We need a set of all dialers (both trusted and untrusted) that we could use for encrypted traffic, and pick a subset of that set that only contains trusted dialers. Your snippet would create two mutually exclusive sets. That could be easily fixed, but we also need both sets to be sorted, we could do that with two different calls to `sort.Sort(byQOSAscending(trusted))` and `sort.Sort(byQOSAscending(all))` or we could create the `trusted` set after the set of all dialers is sorted, I decided over the later one just for clarity and moved the actual build of the list to another function to avoid having two `for` loops that looked too much alike in the same function (New), but now that I realize that a mutex needs to be created for that to work properly I think I'm going to avoid the mutex (and the complexity it carries) and just move this loop it inside New.

myleshorton: Makes sense. Sounds like a plan!

src/github.com/getlantern/balancer/balancer.go
xiam: > I.e. only proxying port 80 on our servers wouldn't mitigate the attack since the attack is typically over port 80.

Yes it would, because censors would not have any chance to tamper with plain HTTP resources at all: any HTTP resource would be intercepted and proxied to a trusted proxy over TLS:

Client -> Plaintext HTTP request (port 80) -> Proxy (over TLS) -> Trusted proxy -> facebook:80

If the censors are between Trusted proxy and facebook:80 then yes, they could inject scripts on facebook.com and feed them to our trusted proxy. But if censors are between client and trusted proxy, then they would not be able to see or tamper any HTTP traffic at all.

In particular with the trust flag, we'll take care of cases like this one:

Client -> Plaintext HTTP request (port 80) -> Proxy (over TLS) -> Bad proxy

In this case, Bad proxy could write anything it wants back to the client, which does not have any way to notice if someone tampered with the content.

But, we are still allowing potentially Bad Proxies to exists, they could be handy too:

Client -> HTTPs request (port 443) -> Proxy (over TLS) -> Bad proxy -> Internet -> Facebook:443

This is a completely different case, if Bad Proxy tampers with the content, the SSL connection would not be valid, since for tampering it would be required to decipher, inject evil stuff, and cipher the content again with a rogue certificate, the browser would alert the user of such situation.

src/github.com/getlantern/flashlight/client/chained.go
xiam: Not sure about it, this would be printed in the logs, I would not like reading that my content is routed over an "untrusted" peer. This legend will help us now that we are testing, but depending on the actual communication strategy we'll probably need to remove this advertising completely.

src/github.com/getlantern/flashlight/client/handler.go
xiam: I think you're right, let me review my wording.

src/github.com/getlantern/flashlight/client/handler.go
xiam: I think so, the useful thing I see here is the comparison that follows the declaration rather than the declaration itself:

```
if foo, err = baz(); err != nil {
  // Do something with err
}
```

Instead of:

```
foo, err = baz()

if err != nil {
  // Do something with err
}
```

The first case does not (easily) allow anything to happen between the error assignation and the comparison it can be easily read like a single *block*, while in the latter there's always a chance of putting more instructions in between:

```
foo, err = baz()

bar, err = baroo()

if err != nil {
  // You didn't catch the first error!
}
```

src/github.com/getlantern/balancer/balancer.go
myleshorton: Hey I think we're actually saying the same thing but just getting mixed up in the working --- I thought you were suggesting that limiting our servers to only proxy content to 80 and 8080 instead of to all the ports they currently proxy would limit this attack. In reality what will limit the attack is *not* sending port 80 content/unencrypted content to peers. I basically interpreted the following:

> only proxying ports such as 80 and 8080 on our servers since that would fix the potential Javascript script injection we're trying to close

as suggesting to only have our servers proxy those ports, but I believe you're suggesting to *send only those ports to our servers*, which is cool.

src/github.com/getlantern/flashlight/client/handler.go
xiam: I completely agree, I'll check them again.

xiam:
Thanks a lot @aranhoide, I agree with you on all your points, let me check my PR again.

src/github.com/getlantern/balancer/balancer.go
myleshorton: Yeah I would agree with not proxying at all there. If we don't have any trustedDialers, we've got bigger problems on our infrastructure side.

src/github.com/getlantern/balancer/balancer.go
xiam: > I believe you're suggesting to send those ports only to our servers, which is cool.

Yeah! exactly, I definitely could have used better wording.

src/github.com/getlantern/chained/dialer.go
aranhoide: I think anyone reading the Lantern source better be familiar with the fact that status codes correspond to integers.  I don't know them all by heart, though, so I like your constants.  If the awkwardness of mixing numbers and constants bothers you, you can hide it in a function, like

go```
func sameStatusCodeClass(statusCode1 int, statusCode2 int) {
    // HTTP response status "classes" come in ranges of 100
   classRange = 100
    return statusCode1 % classRange == statusCode2 % classRange
}
```

Then you can make that check like

go```
if !sameStatusCodeClass(resp.StatusCode, http.StatusOK) {
   ...
}
```

src/github.com/getlantern/flashlight/client/balancer.go
aranhoide: The mutex doesn't guarantee that the balancer returned to the caller of `getBalancer` in goroutine 1 has not been closed when goroutine 1 tries to actually use it, since the mutex gets released when `getBalancer` returns and thus it can't protect its use.  

Also, the balancer is the same one for all the callers of `getBalancer` between `Configure` calls.  If they're not prepared for the case that the balancer they get may be closed when they try and use it, we have a race condition.  It may be unlikely to cause problems if the callers of `getBalancer` only use the balancer shortly after obtaining it, but it's a race condition nonetheless and the mutex doesn't fix it.

src/github.com/getlantern/flashlight/client/handler.go
aranhoide: Yeah, I understand the pattern.  What I meant is I think I read somewhere that Go would allow you to say `clientConn, _, err := ...`, that is, use `:=` even though `err` has already been declared before.  So you could use this idiom without having to declare `clientConn` before.

But I tried this in a test program and I was wrong.  If you do this, `clientConn` would be scoped to the body of the `if`.  I must have confused this with some other special case involving errors.

src/github.com/getlantern/chained/dialer.go
xiam: Looks great!

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -22,21 +22,30 @@ var (
 // Balancer balances connections established by one or more Dialers.
 type Balancer struct {
 	dialers []*dialer
+	trusted []*dialer
 }
 
 // New creates a new Balancer using the supplied Dialers.
 func New(dialers ...*Dialer) *Balancer {
```
xiam: I changed this to create the sorted list of trusted dialers after the list of all dialers.

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -49,7 +58,16 @@ func New(dialers ...*Dialer) *Balancer {
 // remaining Dialers until it either manages to connect, or runs out of dialers
 // in which case it returns an error.
 func (b *Balancer) DialQOS(network, addr string, targetQOS int) (net.Conn, error) {
-	dialers := b.dialers
+	var dialers []*dialer
+
+	_, port, _ := net.SplitHostPort(addr)
+
+	if port == "" || port == "80" || port == "8080" {
```
xiam: Added port 8080.

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -163,9 +182,9 @@ func without(dialers []*dialer, i int) []*dialer {
 	}
 }
 
-// byQOSAscending implements sort.Interface for []*dialer based on the QOS
+// byQOSAscending implements sort.Interface for []*Dialer based on the QOS
 // (ascending)
-type byQOSAscending []*dialer
+type byQOSAscending []*Dialer
```
xiam: This allows us to sort the `dialers` argument of `New()` instead of the former `dhs`.

xiam:
@aranhoide Could you help me with another review?

Thanks!

src/github.com/getlantern/chained/server.go
myleshorton: Yes nice one!

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -22,21 +22,30 @@ var (
...
+	trusted []*dialer
 }
 
 // New creates a new Balancer using the supplied Dialers.
 func New(dialers ...*Dialer) *Balancer {
-	dhs := make([]*dialer, 0, len(dialers))
+	bal := new(Balancer)
+
+	// Sort dialers by QOS (ascending) for later selection
+	sort.Sort(byQOSAscending(dialers))
```
aranhoide: It probably doesn't matter here, but in general it's considered bad form to mutate[1] arguments passed in unless that's the very purpose of the function.  The safest sequence would be to copy `dialers` into `bal.dialers`, then sort `bal.dialers` (which would leave `dialers` alone), then iterate over `bal.dialers` to collect the trusted ones into `bal.trusted`.

[1] I wasn't sure whether `...this` declaration syntax would result in the same slice being passed when called with `this...` syntax, or whether a copy would be made.  It turns out the same slice is passed, so the callee can modify it in place, with changes visible to the caller.  Try this: 

```go
package main

import "fmt"

func main() {
	ss := []string{"zero", "one", "two"}
	fmt.Println(ss)
	change(ss...)
	fmt.Println(ss)
}

func change(strs ...string) {
	strs[2] = "whoops"
}
```

aranhoide:
@xiam: done with the second pass.  Nothing came up that will foreseeably ever cause problems in practice, but maybe it can't hurt to be a bit paranoid.

Feel free to merge this unless you prefer to do further tweaks.

src/github.com/getlantern/balancer/balancer.go
myleshorton: I would agree with that sentiment @aranhoide!

xiam: Now that you mention it, yes, this practice could lead to unexpected behaviour in any piece of code that depends on the variable that was passed. I'd agree with the change.

xiam: Hah! this also provided us with a better solution on the capacity issue.

xiam:
@aranhoide, @myleshorton what do you think of the new proposal for `balancer.New()`?

aranhoide:
@xiam: looks good.  I'll add the comments I talk about in https://github.com/getlantern/lantern/commit/16fa2f5986f6f01351fc2775c3c311e3dbdc2e67#commitcomment-10875558 and merge.

xiam:
@aranhoide Oh, I also added them at the bottom of the corresponding functions.

aranhoide:
QA: unit tests pass both on OS X and Linux.  I couldn't build a binary using `make` in either OS X or Linux (I'll ask the dev list; I don't suppose this can be related to this PR), but I built a binary with `go build` in the flashlight directory.  In my lantern.yaml, I set my fronted host as trusted, and left all chained proxies untrusted.  Browsing to https://thepiratebay.se went though the chained servers (which are prioritized), but browsing to http://nytimes.com went through my fronted proxy.

Nice work!

xiam:
Thanks for your review and comments @aranhoide!

PR #2427, opened 2015-04-22
Support multiple analytics tracking codes

atavism:
Changes to support different analytics tracking codes for each lantern android app, related to https://github.com/getlantern/firetweet/issues/44


xiam:
Thanks @atavism

PR #2430, opened 2015-04-23
program to find and delete test CloudFront distributions

aranhoide:
This also adds support to disable and delete distributions to cfr.go.

Deleting CloudFront distributions is a multi-step process that may
take up to half an hour:

http://docs.aws.amazon.com/AmazonCloudFront/latest/APIReference/DeleteDistribution.html

This will be uploaded to a cloud VM (e.g., the cloudmaster) and run
as a cron job to clean up CloudFlare distributions created by
cfr_test.go.

In the future this can be used to clean up other distributions that
we mark as unused.

This functionality can't be practically unit tested.  Its whole point is
that what it does takes too long to do in tests.  This is also true for
setting up test conditions: it takes up to 15m since you create a
new distribution until you can disable it, and same time since you
have disabled it until you can delete it.


aranhoide:
This has been working fine in my test cloudmaster for a couple weeks already, so I'll merge it to prevent branch rot.

myleshorton:
Nice thanks @aranhoide!

PR #2439, opened 2015-04-24
Adding docker dependency.

xiam:
Adding missing docker dependencies.


xiam:
Hello @aranhoide, 

You already closed the issue but I think this should be fixed, could you take a look?

Thanks,

aranhoide:
LGTM.  Works both in OS X (where I already had docker initialized) and in Ubuntu (where I hadn't run `make docker` yet).

PR #2441, opened 2015-04-24
force closing connection if request sent

### LINKED ISSUE #2330: Lots of too many open files errors
Reading lantern.yaml appears to be a common place, but we're getting this error in a variety of places and are clearly failing to close either a network connection or file descriptor somewhere.

fffw:
Resolves #2330. There will be no CLOSE_WAIT connections left.

I tried use `idletiming` instead of close immediately, but caused more CLOSE_WAIT state connection. Might worth digging into it but I guess the current way is enough.

### DIFF
--- a/src/github.com/getlantern/flashlight/client/handler.go
+++ b/src/github.com/getlantern/flashlight/client/handler.go
@@ -87,7 +87,14 @@ func (client *Client) targetQOS(req *http.Request) int {
 // responsible for responding to the initial CONNECT request with a 200 OK.
 func pipeData(clientConn net.Conn, connOut net.Conn, req *http.Request) {
 	// Start piping from client to proxy
-	go io.Copy(connOut, clientConn)
+	go func() {
+		io.Copy(connOut, clientConn)
+		// Force closing if EOF at the request half or error encountered.
+		// A bit arbitrary, but it's rather rare now to use half closing
+		// as a way to notify server. Most application closes both connections
+		// after completed send / receive so that won't cause problem.
+		clientConn.Close()
+	}()
 
 	// Respond OK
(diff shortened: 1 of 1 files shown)


fffw:
Well, `idletiming` is more bulletproof. Please hold on while I checking why it didn't work.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -87,7 +87,14 @@ func (client *Client) targetQOS(req *http.Request) int {
...
 func pipeData(clientConn net.Conn, connOut net.Conn, req *http.Request) {
 	// Start piping from client to proxy
-	go io.Copy(connOut, clientConn)
+	go func() {
+		io.Copy(connOut, clientConn)
+		// Force closing if EOF at the request half or error encountered.
+		// A bit arbitrary, but it's rather rare now to use half closing
+		// as a way to notify server. Most application closes both connections
+		// after completed send / receive so that won't cause problem.
+		clientConn.Close()
```
myleshorton: This seems like it only makes sense if the ```io.Copy(connOut, clientConn)``` call returns an error no? If it's just returning because of EOF on clientConn, that means clientConn is already closed, so no need to close it again. 

It seems like we might as well also call connOut.Close here as well, as there's no clientConn for connOut to write to at that point and it seems like there are cases where we'd otherwise rely on the server connection timing out (i.e. if the clientConn hasn't written any data and has just closed, the server connection could be waiting around for data up until some server-side timeout).

Does that make sense?

myleshorton:
OK, @fffw I'm going to merge this for now since it seems to fix the problem. Thanks!

PR #2447, opened 2015-04-27
use untrusted dialers if no trusted one available

### LINKED ISSUE #2440: ReverseProxy reports "No dialers left to try"
> Apr 24 14:19:35.150 - detour: detour.go:139 Dial detoured to pki.google.com:80 failed
> 2015/04/24 22:19:35 http: proxy error: No dialers left to try

But at the same time, `Client.intercept()` can dial. Must be subtle differences between the way invoking `detour.Dialer`.

fffw:
Resolves #2440

We also need to run genconfig.bash again to mark all current servers as trusted. Trying…

### DIFF
--- a/src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
+++ b/src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
@@ -17,8 +17,8 @@ client:
       authtoken: "{{.auth_token}}"
       pipelined: true
       weight: 1000000
-      qos: 10{{end}}
-      trusted: true
+      qos: 10
+      trusted: true{{end}}
   masqueradesets:
     cloudflare:{{range .masquerades}}
     - domain: {{.Domain}}
--- a/src/github.com/getlantern/flashlight/genconfig/fallbacks.go.tmpl
+++ b/src/github.com/getlantern/flashlight/genconfig/fallbacks.go.tmpl
@@ -12,5 +12,6 @@ var fallbacks = map[string]*client.ChainedServerInfo{ {{range .fallbacks}}
     Pipelined: true,
     Weight:    1000000,
     QOS:       10,
+    trusted:   true,
   }, {{end}}
 }


src/github.com/getlantern/balancer/balancer.go
```diff
@@ -71,13 +71,19 @@ func (b *Balancer) DialQOS(network, addr string, targetQOS int) (net.Conn, error
 
 	_, port, _ := net.SplitHostPort(addr)
 
-	if port == "" || port == "80" || port == "8080" {
+	// try proxy non-https traffic through trusted servers first
+	if port != "443" {
```
fffw: I reversed the test as http server may listen at arbitrary ports but https port is usually fixed.

fffw:
@xiam would you like to take a look? It's related to https://github.com/getlantern/lantern/pull/2424

xiam:
Sure @fffw

src/github.com/getlantern/balancer/balancer.go
xiam: There's an issue with this change and it's that we support many other different ports:

```
./flashlight.go:                AllowedPorts: []int{80, 443, 8080, 8443, 5222, 5223, 5228},
```

From that list we only want to take ports `80` an `8080` as probably untrusted. It is true that plaintext traffic may travel along any port but I think we are not ready to make that movement and claim that Lantern routes all plaintext traffic to trusted servers, as of today it would be more accurate to say we proxy traffic that passes over ports `80` and `8080` to trusted servers and that users must take care of other ports.

xiam: Related discussion: https://github.com/getlantern/lantern/pull/2424#discussion_r28865922

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -71,13 +71,19 @@ func (b *Balancer) DialQOS(network, addr string, targetQOS int) (net.Conn, error
...
 	} else {
 		dialers = b.dialers
 	}
 
 	for {
+		// in case dialers == b.trusted but no trusted dialer available,
+		// untrusted dialers are better than no dialer at all.
+		if len(dialers) == 0 {
+			dialers = b.dialers
+		}
```
xiam: We discussed this a bit here https://github.com/getlantern/lantern/pull/2424#discussion-diff-28866183, probably the best way to fix this is making sure to provide users with (at least one) trusted dialers instead of letting the client pass when no dialers are available.

src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
```diff
@@ -17,8 +17,8 @@ client:
       authtoken: "{{.auth_token}}"
       pipelined: true
       weight: 1000000
-      qos: 10{{end}}
-      trusted: true
+      qos: 10
+      trusted: true{{end}}
```
xiam: Good catch!

src/github.com/getlantern/balancer/balancer.go
fffw: Oh sorry I saw your initial comment of https://github.com/getlantern/lantern/pull/2424 but didn't notice later discussion. 
Ok I'll revert the changes to `balancer.go`

fffw:
@xiam Thanks for your clarification, I reset the PR to just add `trusted` field to server info.

src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
aranhoide: indeed!

aranhoide:
LGTM.

PR #2458, opened 2015-04-29
Added pubsub for general use and for the user's location closes #2457

### LINKED ISSUE #2457: Refactor geolookup code to use pub/sub
This turns out to be very important for enabling things like running in give mode.

myleshorton:
Curious about people's thoughts on this. Seems cleaner, but it seems like it could be better. For example, it might make more sense to refactor the yamlconf Manager Start method to separate out the CustomPoll to allow us to only start polling when we know the location.

If you want to take a look @aranhoide that would be helpful.


src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -142,9 +143,6 @@ func Run(updateHandler func(updated *Config)) error {
 
 func updateGlobals(cfg *Config) error {
 	globals.InstanceId = cfg.InstanceId
-	loc := &geolookup.City{}
-	loc.Country.IsoCode = cfg.Country
-	globals.SetLocation(loc)
```
myleshorton: This seems like it was a bug. Not sure why we ever set the global location here, but in practice it also frequently would *overwrite the location* discovered in our geolookup -- i.e. the likely cause of the many "xx" locations we see in analytics.

This change also makes it so that the only thing that ever modifies the location is the geolookup itself, which I think makes sense.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -380,26 +378,18 @@ func (cfg Config) fetchCloudConfigForAddr(proxyAddr string) ([]byte, error) {
...
+	if country != "" && country != "xx" {
+		ret, err := cfg.fetchCloudConfigForCountry(client, country)
+		// We could check specifically for a 404, but S3 actually returns a 403 when
+		// a resource is not available.  I thought we'd better lean on the side of
+		// robustness by avoiding hardcoding an S3-ism here, than to try and save an
+		// extra request every now and then.
+		if err == nil {
+			return ret, err
+		} else {
+			log.Debugf("Couldn't fetch cloud config for country '%s'; trying the default one", country)
```
myleshorton: This will get called (albeit somewhat tangentially) once we learn the location and the location event is propagated.

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -31,8 +31,25 @@ var (
...
+}
+
+func GetCountry() string {
+	loc := GetLocation()
+	if loc == nil {
+		return ""
+	}
+	return loc.Country.IsoCode
+}
+
```
myleshorton: Move these to the more specific class, ditch global setting of the location.

src/github.com/getlantern/flashlight/pubsub/pubsub.go
```diff
@@ -0,0 +1,19 @@
+// package pubsub provides a global publish and subscribe infrastructure
+// for lantern
+package pubsub
+
```
myleshorton: This is a wrapper around https://github.com/asaskevich/EventBus to make it easier to use at the expense of these being global.

src/github.com/getlantern/yamlconf/yamlconf.go
```diff
@@ -241,17 +242,25 @@ func (m *Manager) processUpdates() {
 
 func (m *Manager) processCustomPolling() {
 	for {
-		mutate, waitTime, err := m.CustomPoll(m.getCfg())
+		_, waitTime, _ := m.OnDemandPoll()
+		time.Sleep(waitTime)
+	}
+}
+
+func (m *Manager) OnDemandPoll() (func(cfg Config) error, time.Duration, error) {
```
myleshorton: Allows things to trigger a lookup outside of the polling schedule. Added a lock to make sure it never conflicts with the polling itself.

myleshorton:
OK this should be significantly cleaner, @aranhoide. The pub sub system now uses types, and we only start polling for config changes once we've determined the user's location.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -285,7 +285,7 @@ func runServerProxy(cfg *config.Config) {
 		},
 		AllowedPorts: []int{80, 443, 8080, 8443, 5222, 5223, 5228},
 
-		// We allow all censored countries plus us, es and mx because we do work
+		// We allow all censored countries plus us, es, mx, and gb because we do work
```
aranhoide: Good catch, sorry for having failed to update the comment.

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -104,7 +122,7 @@ func write() {
 			consecutiveFailures += 1
 			// If available, publish last known location
 			if oldLocation != nil {
-				service.Out <- location
+				service.Out <- oldLocation
```
aranhoide: Good catch!

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -115,9 +113,17 @@ func Init() (*Config, error) {
 			return
 		},
 	}
-	initial, err := m.Start()
+	initial, err := m.Init()
 	var cfg *Config
 	if err == nil {
```
aranhoide: Here it would be a good place to declare the `Once` I mention in the next comment.
```go
var once sync.Once
```

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -115,9 +113,17 @@ func Init() (*Config, error) {
...
+		er := pubsub.Sub(&geo.City{}, func(city *geo.City) {
+			log.Debugf("Got location: %v", city.Country.IsoCode)
+			m.StartPolling()
```
aranhoide: Why start a new polling goroutine whenever our location changes?  I see you added a mutex in yamlconf to protect against some of the damage, but this still builds up unnecessary polling requests.

This could be prevented by creating a `sync.Once` out of this callback (see above) and changing this line to
```go
once.Do(func () { m.StartPolling() })
```

src/github.com/getlantern/flashlight/config/config.go
aranhoide: I'm not sure why this was being done, but I think this was done on purpose.   Actually, after removing this I don't think there is any point to having a country in the config anymore.

Plausible reasons:

- allow setting geolocation for testing stuff

- save location across runs so the user dot doesn't start in null island until first geolocation request.

Since none of these looks too compelling, I'm fine with removing these lines *and* cfg.Country.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -380,26 +383,18 @@ func (cfg Config) fetchCloudConfigForAddr(proxyAddr string) ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Unable to initialize HTTP client: %s", err)
 	}
-	// Try and wait to get geolocated, up to a point.  Waiting indefinitely
-	// might prevent us from ever getting geolocated if our current
-	// configuration is preventing domain fronting from working.
```
aranhoide: I think this comment had a valid point.  If we wait indefinitely for geolookup, but the (domain fronted) geolookup server can't be reached for any reason related to the current config, then the client can never recover from that.

src/github.com/getlantern/flashlight/pubsub/pubsub.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/asaskevich/EventBus"
+)
+
+var (
+	bus = EventBus.New()
+)
+
+// Pub publishes the given interface to any listeners for that interface.
+func Pub(topic interface{}) {
```
aranhoide: Telling topics by type, while it saves an argument in Pub calls, seems more odd and limiting to me than the convenience warrants.  Exposing a geolookup.Topic string constant and using that for Pub and Sub calls seems cleaner to me.

src/github.com/getlantern/flashlight/pubsub/pubsub_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+func TestPublish(t *testing.T) {
+	var received string = ""
+	Sub(&str{}, func(s *str) {
+		received = s.Val
+	})
+	Pub(&str{Val: "test"})
+
+	runtime.Gosched()
+	if received != "test" {
```
aranhoide: `received` is mutated in a different goroutine without synchronization.  Although apparently it works now, I don't think Go guarantees any kind of consistency here.  I'd signal success through a channel with a short timeout instead.

src/github.com/getlantern/yamlconf/yamlconf.go
```diff
@@ -241,17 +245,25 @@ func (m *Manager) processUpdates() {
 
 func (m *Manager) processCustomPolling() {
 	for {
-		mutate, waitTime, err := m.CustomPoll(m.getCfg())
+		_, waitTime, _ := m.poll()
```
aranhoide: No reason to have `poll` return three values if its only caller only wants one.

src/github.com/getlantern/yamlconf/yamlconf.go
```diff
@@ -204,6 +201,13 @@ func (m *Manager) Start() (Config, error) {
 	return m.getCfg(), nil
 }
 
+// StartPolling starts polling if there is a custom polling function defined.
+func (m *Manager) StartPolling() {
```
aranhoide: I don't see why we should allow this to be called more than once.  It should either error or silently ignore further calls.

src/github.com/getlantern/yamlconf/yamlconf.go
```diff
@@ -241,17 +245,25 @@ func (m *Manager) processUpdates() {
...
 	for {
-		mutate, waitTime, err := m.CustomPoll(m.getCfg())
+		_, waitTime, _ := m.poll()
+		time.Sleep(waitTime)
+	}
+}
+
+func (m *Manager) poll() (func(cfg Config) error, time.Duration, error) {
+	m.pollMutex.Lock()
+	defer m.pollMutex.Unlock()
```
aranhoide: As said, I think concurrency should have been prevented by the caller, so these locks shouldn't be necessary.

aranhoide:
Done with my first pass.

I think building up polling threads every time our location changes should be fixed.

I'm a bit uneasy about blocking config polls until geolocation succeeds, because there may be a catch-22 situation if geolocation fails for reasons that a new config could fix.  OTOH, this may be too far fetched to worry about, and a timeout doesn't fit very well in the pub/sub scheme, and in the worst case we still have auto-updates, so I won't lose sleep over this either way.

myleshorton:
Thanks Snyder! Agreed on the location change issue. I was thinking it was only really relevant when your *country* changes, which is rare, but it's actually any time your IP changes I believe, which is much more common.

> I'm a bit uneasy about blocking config polls until geolocation succeeds, because there may be a catch-22 situation if geolocation fails for reasons that a new config could fix.

I thought about that too, but I think the saving grace is that the most likely reason for a geolookup to fail is domain fronting itself (especially since the geolookup service has been rock solid), but that would also cause the config lookup to fail. Will look at other comments too. Thanks again.

src/github.com/getlantern/flashlight/config/config.go
myleshorton: > save location across runs so the user dot doesn't start in null island until first geolocation request.

That one actually makes sense to me. Let me see:

1. If that's actually supported now
1. If I can just add it

myleshorton: Actually I just stripped it -- the lat/long you could make more of an argument for caching, but less so the country I think.

src/github.com/getlantern/flashlight/pubsub/pubsub_test.go
myleshorton: That's where the runtime.Gosched comes into play -- it yields to the other go routines, and there's only one other routine in this case.

src/github.com/getlantern/flashlight/config/config.go
myleshorton: Cool -- nice idea. I just added that in yamlconf itself actually, which seems a bit more bulletproof. Good catch.

src/github.com/getlantern/flashlight/pubsub/pubsub.go
myleshorton: OK, I just added some changes here, but actually to use strings instead of ints. I want to encourage compile time checking here, which is the whole purpose of doing it by type, but I think the int approach is a happy medium of sorts (i.e. it would feel very odd to throw in arbitrary ints as topics, encouraging callers to take a peek at how it works and to add topics to pubsub.go).

src/github.com/getlantern/yamlconf/yamlconf.go
myleshorton: Fixed!

src/github.com/getlantern/yamlconf/yamlconf.go
myleshorton: Done!

myleshorton:
OK just added those changes @aranhoide or some slight variation of them. Let me know what you think!

src/github.com/getlantern/flashlight/pubsub/pubsub_test.go
aranhoide: I'm not sure that runtime.Gosched counts as synchronization over `received` though.  I agree if Go just ran these statements straightforwardly, `received` should have been mutated by the time it is checked.  And since this test passes, something close enough must be happening in the current version of Go.  All I'm saying is they reserve the right to mess with execution (e.g. for optimization) in ways that break otherwise reasonable assumptions about observability of changes to variables across goroutines.

https://golang.org/ref/mem

aranhoide: Then again, I'm not particularly worried in this case.  This is not production code and I expect in the worst case the test could give us some false alarm some day.

src/github.com/getlantern/flashlight/config/config.go
aranhoide: Sounds good.

src/github.com/getlantern/flashlight/pubsub/pubsub.go
aranhoide: ints work for me too.  The only problem is that I'd have preferred to have the constants in the module that defines what they're about (e.g. geolookup in this case).  But it's no big deal either way.

If you want to really enforce that constants are used you can make topics instances of a structure that is local to the pubsub package.

aranhoide:
Hm yes, I was under the impression that we were trying to download the cloud config directly, at least if the domain fronted request fails.  And maybe we should do that, but that's a different issue.

This morning I realised that the same effect could be achieved without changing `yamlconf` or `config.Init` at all, by just adding a `geolookup.WaitForLocation()` function that could be used by anyone that wants to block until we have a location.  Then use that in `fetchCloudConfigForAddr`.  There is really no practical difference between waiting *before* we start polling for updates or just doing that wait on our first run of the poll, is it?

`geolookup.WaitForLocation` could use the `ret := <- myChan; myChan <- ret; return ret` pattern instead of pubsub.  That would save us a dependency, since the UI is not using pubsub anyway to get its location updates.  It would also return immediately if `geolookup.location != nil` when it's called.  I don't think it makes a difference in the case of `config.go`, but in general whoever subscribes for the Location pubsub channel will get their callback hit in the *next* geolookup even if we have successfully geolocated in this run, so they need to remember to check for location first.

aranhoide:
While I think this purpose could be met without this dependency, I suppose you expect this will be used for other things.  Otherwise LGTM and tests pass, so merging.

src/github.com/getlantern/flashlight/pubsub/pubsub_test.go
myleshorton: Yeah I realized it would be just as easy and more idiomatic to just use channels, so that's changed.

myleshorton:
Oh sorry didn't realize you had merged! I think the test was broken at least, so I'll merge that in as well.

Issue #2394, opened 2015-04-16
 "net/http: request canceled while waiting for connection"


aranhoide:
This is the most frequent error log at loggly at the time of this writing.

fffw:
From loggly, those errors are reported from two packages --- `flashlight.statserver` and `flashlight.geolookup`, but triggered from single line of code: [geolookup.LookupIPWithClient](https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/geolookup/geolookup.go#L122)

fffw:
`geolookup` uses a `http.Client` with 20 seconds timeout to access geoserver. It will `CancelRequest` after 20 seconds. [`http.Transport`](http://golang.org/src/net/http/transport.go#L539) will report this error as a consequence.

Other clients, including Loggly client and cloud config fetcher, use `HTTPClient` from `flashlight/util` without a timeout, so this error will not occur to them.

We may need to remove this timeout and compare the error count of geolookup versus fetching cloud config.

fffw:
The intervals of fetching cloud config (1 min) and doing geolookup (30 secs) are too small. They don't change often, continuously pulling is a waste of bandwidth. I suggest an interval of 30 min - 1 hour or even longer.

fffw:
Oh sorry @xiam I just noticed that this issue was assigned to you! We may discuss a bit about it.

xiam:
Don't worry @fffw, I've been not able to reproduce it. We may as well keep on with the discussion.

myleshorton:
Hey @fffw I actually think we should probably include *some* timeout and should add a timeout to the other services. Otherwise we'll just be left with idle connections that can ultimately contribute to file descriptor limit issues. I would just make the timeouts really high, like 300 seconds.

fffw:
I'm not sure enough but I'll check if dialer and enproxy have timeout themselves. Anyway, 20 seconds is long enough to establish connection so we should dig deeper on why the error is reported so frequently.

xiam:
It seems like the sleep time before retrying to geolocate a peer in case of error was too small, also we were missing some details in the log.

> I would just make the timeouts really high, like 300 seconds.

@myleshorton I think 300s is too much, the error we were seeing was just the last error and we can't be sure if every request before that last one timed out too. Instead of trying with 300s I proposed a more conservative increment of 60s, put more time between attempts and added more details (attempt number and error message) to the log.

Please see this PR https://github.com/getlantern/lantern/pull/2474

[...]

PR #2460, opened 2015-04-30
remove timeout of geolookup

fffw:
To stop errors in https://github.com/getlantern/lantern/issues/2394 from being reported.

### DIFF
--- a/src/github.com/getlantern/geolookup/geolookup.go
+++ b/src/github.com/getlantern/geolookup/geolookup.go
@@ -5,14 +5,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"time"
 
 	"github.com/getlantern/golog"
 )
 
 const (
 	geoServeEndpoint = `http://geo.getiantem.org/lookup/%s`
-	geoLookupTimeout = 20 * time.Second
 )
 
 var (
@@ -107,7 +105,6 @@ func LookupIPWithClient(ipAddr string, httpClient *http.Client) (*City, error) {
 		log.Trace("Using default http.Client")
 		httpClient = defaultHttpClient
 	}
-	httpClient.Timeout = geoLookupTimeout
 
 	var err error
 	var req *http.Request


xiam:
Hello @fffw,

This will indeed stop timeout errors from being reported but may contribute to have idle connections (and open file descriptors). Why not increasing the timeout instead?

fffw:
@xiam It might not be the case as OS itself has timeouts to establish a connection. But I agree to hold on this PR and dig deeper to find the root cause of frequent timeouts (cloud config fetcher didn't have so many errors).

xiam:
Hey @fffw, would you still like to keep this one opened?

fffw:
@xiam Thanks for mention it. Let's keep watching

myleshorton:
I meant to comment on this one guys. In general I think we should never have sockets on either the client or server side that have no connection timeouts or read timeouts, as pretty much everything should eventually timeout, and not doing it is just begging for file limit issues.

Can you say more about OS-level timeouts @fffw? I'm unaware of any OS doing that.

On the specific case of the geolookup, I think if anything we should significantly increase the timeout. 20 seconds is way more than enough *if everything is operating normally*, but in cases of high packet loss due to anything really -- flaky local connection, flaky national connection, connecting from a plane, etc, connections can take longer than that. FireFox I believe uses 300 seconds, for example, but Lantern historically has used timeouts of around 60 seconds for things like this. I think increasing that timeout is a useful and even a necessary change.

fffw:
In general, I think each OS should have a timeout during TCP SYN phase. I just googled a bit.

On *nix, timeout for non-established TCP connections can be obtain through`sysctl net.inet.tcp.keepinit`, which defaults to 75000 ms.

The value differs among Windows versions. https://support.microsoft.com/en-us/kb/170359 has some explanation.

A consistent timeout would be good, but I think it might be better to be applied on dialer and connection, rather than http.Client. At least, "i/o timeout" is more meaningful than "net/http: request canceled while waiting for connection"

PR #2469, opened 2015-05-04
log detailed error if dialDetour() failed

fffw:
Now the error message will look like:

> May 04 01:08:52.524 - detour: detour.go:139 Dial detoured to ssl.gstatic.com:443 failed: No dialers left


aranhoide:
LGTM

Issue #2398, opened 2015-04-16
"x509: certificate signed by unknown authority"


aranhoide:
Seen in loggly, seems to appear in FireTweet a lot.  Maybe something wrong with the CA setup there?  P0 because of FireTweet focus, and we're already on it anyway.

myleshorton:
To add a little more context, this is happening somewhere in fronted, although unfortunately we don't have the line number in logs. We should ideally at least report the actual root cert -- could be a MITM attempt.

atavism:
@myleshorton Created a PR for this after I saw your comment. It prepends log messages with the file name and line number: https://github.com/getlantern/flashlight-build/pull/182

myleshorton:
It would be great to just log the cert we don't trust.

xiam:
I have a few hints but still no luck debugging the exact cause of this, just a quick update, it looks like the @myleshorton suggestion it's a bit complicated to implement, this error is produced right within the Go's [crypto/x509](https://golang.org/src/crypto/x509/verify.go#L104) package, and this verification is triggered by a call to `httpClient.Do()`. I'm taking a different approach now trying to see if I can produce this error by deleting system's certs on a vm.

xiam:
Going to try one more thing with `vms.dialer.tlsConfig`, maybe we can get some debugging info from it.

fffw:
Started from Apr 28th, only tens of them reported each day, but it burst since May 4th.

I exported all messages of this error on 2.0.0-beta4 in last 7 days. The masquerade domains was not distributed evenly. One domain, *www.hawkhost.com*, contributed 436 out of 1394 items, occurred each day across timezones. Other domains only had less than 5 occurrences each.

xiam:
@fffw I'm following this issue and I was able to find some some recent logs of it, however the tlsInfo bit was not recorded, do you think we can remove [this check](https://github.com/getlantern/lantern/pull/2472/files#diff-d89b8c910f43a32ce0cc53f817ddef06R148) temporarily?

fffw:
Of course @xiam . It's just for logging purpose anyway.

xiam:
Looks like the error count for this one has decreased in beta6, only `636` occurrences this week so far.

I've noticed that most of the logged errors mention the same domain name, so that's a great clue. As far as I understand if this domain name is misbehaving it should not have be included by the genconfig script in the first place, right?

xiam:
Chrome does not complain about the authenticity of the CA.

[...]

PR #2472, opened 2015-05-04
Adding tlsInfo() function, it could help debugging https://github.com/ge...

xiam:
I could not reproduce the "x509: certificate signed by unknown authority" error but this function could help getting more debugging information.

See https://github.com/getlantern/lantern/issues/2398


fffw:
I greped all occurrences of `UnknownAuthorityError` in go source code.

```
./crypto/x509/root_windows.go:92:			return UnknownAuthorityError{c, nil, nil}
./crypto/x509/root_windows.go:132:			return UnknownAuthorityError{c, nil, nil}
./crypto/x509/root_windows.go:134:			return UnknownAuthorityError{c, nil, nil}
./crypto/x509/verify.go:92:// UnknownAuthorityError results when the certificate issuer is unknown
./crypto/x509/verify.go:93:type UnknownAuthorityError struct {
./crypto/x509/verify.go:103:func (e UnknownAuthorityError) Error() string {
./crypto/x509/verify.go:319:		err = UnknownAuthorityError{c, hintErr, hintCert}
```

It seems that the only place which triggers this error is at https://golang.org/src/crypto/x509/verify.go#L319. `hintErr` and `hintCert` should have been filled and `UnknownAuthorityError` should print more info if `hintErr` is not `nil`. But we didn't get any hint printed. TBH I don't know what it implies.

xiam:
Thanks @fffw! I fixed the wrong argument and I initially thought we would not be able to remove the "garbage" bytes but I kept on investigating and I found another solution. Could you give it a look again?

src/github.com/getlantern/fronted/masquerade.go
```diff
@@ -137,10 +137,13 @@ func (vms *verifiedMasqueradeSet) doVerify(masquerade *Masquerade) bool {
...
 		httpClient := vms.dialer.HttpClientUsing(masquerade)
-		req, _ := http.NewRequest("HEAD", "http://www.google.com/humans.txt", nil)
+		req, err := http.NewRequest("HEAD", "http://www.google.com/humans.txt", nil)
+		if err != nil {
+			errCh <- fmt.Errorf("http.NewRequest Error: %s", err)
+		}
 		resp, err := httpClient.Do(req)
 		if err != nil {
-			errCh <- fmt.Errorf("HTTP ERROR FOR MASQUERADE %v: %v", masquerade.Domain, err)
+			errCh <- fmt.Errorf("HTTP ERROR FOR MASQUERADE %v: %v, tlsInfo: %s", masquerade.Domain, err, vms.dialer.tlsInfo(masquerade))
```
fffw: Cool @xiam! New `tlsInfo` is pretty clean.

To avoid adding unrelated info to Loggly, instead of print `tlsInfo` for any errors, can we only print it if the error is`x509.UnknownAuthorityError`, or defined in https://golang.org/src/crypto/x509/verify.go ?

fffw: BTW, if you can change the message prefix to lower-case? It's the only occurrence of all-capitalized log.

xiam: > To avoid adding unrelated info to Loggly, instead of print tlsInfo for any errors, can we only print it if the error isx509.UnknownAuthorityError

Good idea!

xiam:
I took your suggestions @fffw, care to look again?

fffw:
@xiam I would like using type assertion trick.

Code is good to me in general, so I just added a new commit to your branch. Could you just review the change I made?

src/github.com/getlantern/fronted/masquerade.go
```diff
@@ -137,10 +138,19 @@ func (vms *verifiedMasqueradeSet) doVerify(masquerade *Masquerade) bool {
...
-		req, _ := http.NewRequest("HEAD", "http://www.google.com/humans.txt", nil)
+		req, err := http.NewRequest("HEAD", "http://www.google.com/humans.txt", nil)
+		if err != nil {
+			errCh <- fmt.Errorf("http.NewRequest Error: %s", err)
+		}
 		resp, err := httpClient.Do(req)
 		if err != nil {
-			errCh <- fmt.Errorf("HTTP ERROR FOR MASQUERADE %v: %v", masquerade.Domain, err)
+			errmsg := fmt.Sprintf("HTTP error for masquerade %v: %v", masquerade.Domain, err)
+			if _, ok := err.(x509.UnknownAuthorityError); ok {
```
xiam: Didn't know this can be done but makes total sense. Great trick!

xiam:
@fffw awesome! thanks.

src/github.com/getlantern/fronted/masquerade.go
myleshorton: Me neither -- cool!

myleshorton:
OK I'm merging this in guys. Looks good.

PR #2474, opened 2015-05-05
Adjusting error threshold in peer geolocation.

xiam:
`retryWait` was producing small values and error reporting was incomplete.

This could help fixing https://github.com/getlantern/lantern/issues/2394

PR #2475, opened 2015-05-06
adjust log format in statserver/peer.go

fffw:
The log we just merged from https://github.com/getlantern/lantern/pull/2474 was like below

> May 06 01:59:32.616 - flashlight.statserver: peer.go:106 Failed geolocation attempt 9/9: "Could not get response from server: \"Get http://geo.getiantem.org/lookup/178.62.198.216: net/http: request canceled while waiting for connection\"". Waiting 4m16s before retrying.

It would be better to revise it a bit as
1. it results different message groups on Loggly, as the waiting time varies;
2. quotation mark and backslash makes the text a bit clutter.

@xiam How do you think?

### DIFF
--- a/src/github.com/getlantern/flashlight/geolookup/geolookup.go
+++ b/src/github.com/getlantern/flashlight/geolookup/geolookup.go
@@ -46,7 +46,7 @@ func Configure(newClient *http.Client) {
 	if service == nil {
 		err := registerService()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


fffw:
both %q and %v will add a quotation mark around the error message, %s might be more appropriate.

xiam:
Hello @fffw, I think you're right, lets remove the quotes!

myleshorton:
Nice!

Issue #2402, opened 2015-04-18
Errors fetching https://s3.amazonaws.com/lantern_config/cloud.2.0.0-nl.yaml.gz


myleshorton:
yamlconf: Custom polling failed: Unable to read yaml from https://s3.amazonaws.com/lantern_config/cloud.2.0.0-nl.yaml.gz: Unable to fetch cloud config at https://s3.amazonaws.com/lantern_config/cloud.2.0.0-nl.yaml.gz: Get https://s3.amazonaws.com/lantern_config/cloud.2.0.0-nl.yaml.gz: EOF

xiam:
Also Get https://s3.amazonaws.com/lantern_config/cloud.2.0.0-nl.yaml.gz: Bad Gateway

myleshorton:
Closing this since we're not seeing it anymore.

PR #2476, opened 2015-05-06
add fix for https://github.com/getlantern/lantern/issues/2402

atavism:
This is related to the error fetching ``cloud.yaml`` we've been seeing for a long time in Loggly. The ``resp.Body.Close() `` is insufficient, and we need to explicitly indicate the request should be closed after we finish reading the response body. More details here: https://stackoverflow.com/questions/17714494/golang-http-request-results-in-eof-errors-when-making-multiple-requests-successi

### DIFF
--- a/src/github.com/getlantern/flashlight/config/config.go
+++ b/src/github.com/getlantern/flashlight/config/config.go
@@ -415,6 +415,12 @@ func (cfg Config) fetchCloudConfigForCountry(client *http.Client, country string
 		// Don't bother fetching if unchanged
 		req.Header.Set(ifNoneMatch, lastCloudConfigETag[url])
 	}
+
+	// make sure to close the connection after reading the Body
+	// this prevents the occasional EOFs errors we're seeing with
+	// successive requests
+	req.Close = true
+
 	resp, err := client.Do(req)
 	if err != nil {
 		return nil, fmt.Errorf("Unable to fetch cloud config at %s: %s", url, err)


fffw:
Great catch @atavism ! I believe we can apply it somewhere else. Merging.

myleshorton:
Interesting @atavism thanks. To me the thinking here seems slightly incorrect though TBH. We create a new util.HTTPClient on each poll, which in turn creates a new transport. While I agree we want to close those connections each time, the result of the current code would appear to me to be too many open file descriptors and not errors on the connection itself or issues with reusing the connection since we're recreating the transport each time.

I actually think we should change util.HTTPClient significantly. To me it should use something closer to DefaultTransport in https://golang.org/src/net/http/transport.go that actually has reasonable timeouts for the handshake, connecting, reading, etc. I also think it should just set DisableKeepAlives to true and should have a second method util.PersistentHTTPClient that uses keep alives since we do really want the connections to stay open for things like Loggly. 

It's also useful to note that these util.HTTPClient connections are or should be just going to our server on localhost, which itself has no timeouts. So this could easily be a major cause of the file descriptor issues, but I don't see it as a cause of the issues with the connections described here unless I'm missing something.

Either way it's clearly a crucial area, so great catch. I just logged the more general issue here:

https://github.com/getlantern/lantern/issues/2491

myleshorton:
It's also helpful to note that req.Close is just setting the Connection: close header on the request. This wasn't obvious to me without digging into the code, so figured I'd make a note of it.

PR #2490, opened 2015-05-07
Just fixes the test...

myleshorton:


aranhoide:
You may want to add this to test-and-cover, because the travis build was passing with the broken tests.

myleshorton:
Oh right -- I forgot about the need to do that. I might fix that issue (of having to manually add new packages) while I'm at it. Thanks!

myleshorton:
This also adds the pubsub package.

PR #2495, opened 2015-05-08
Fix conditional so it doesn't break all non-IE closes #2494

### LINKED ISSUE #2494: Changes for IE broke everything else!
Now the lantern ui only appears in IE =).

myleshorton:
This should fix it. Tested on OSX and Windows with IE and non-IE.


myleshorton:
Want to take a look at this @fffw or @atavism?

atavism:
Yep, this looks good. I just merged @fffw's other changes in too related to this. Closing.

myleshorton:
ok thanks @atavism!

PR #2498, opened 2015-05-08
Change to not use persistent connections by default closes #2491

### LINKED ISSUE #2491: Fix issues with util.HTTPClient
First, pretty much every time this is used it's intended to use the localhost proxy -- we should make sure it always is. Otherwise nothing will be tunneled using domain fronting as intended. Second, most callers aren't explicitly closing connections, so it can easily lead to connection leaks as is for that reason alone. Beyond that though, it doesn't use an connection, read, or TLS handshake timeouts, so can lead to connection leaks as a result of that as well.

myleshorton:
Also added timeouts to the transport.

### DIFF
--- a/src/github.com/getlantern/flashlight/util/http.go
+++ b/src/github.com/getlantern/flashlight/util/http.go
@@ -6,15 +6,53 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"time"
 
+	"github.com/getlantern/golog"
 	"github.com/getlantern/keyman"
 )
+
+var (
+	log = golog.LoggerFor("flashlight.util")
+)
+
+// PersistentHTTPClient creates an http.Client that persists across requests.
+// If rootCA is specified, the client will validate the server's certificate
+// on TLS connections against that RootCA. If proxyAddr is specified, the client
+// will proxy through the given http proxy.
+func PersistentHTTPClient(rootCA string, proxyAddr string) (*http.Client, error) {
+	return httpClient(rootCA, proxyAddr, true)
+}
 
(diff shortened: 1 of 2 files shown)


myleshorton:
You guys mind taking a peek at this @ffw, @aranhoide, or @atavism? Seems like a pretty clear possible source of connection leaks otherwise.

myleshorton:
@fffw can you look at this one?

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -6,15 +6,53 @@ import (
...
+		TLSHandshakeTimeout: 10 * time.Second,
+
+		// This method is typically used for creating a one-off HTTP client
+		// that we don't want to keep around for future calls, making
+		// persistent connections a potential source of file descriptor
+		// leaks. Note the name of this variable is misleading -- it would
+		// be clearer to call it DisablePersistentConnections -- i.e. it has
+		// nothing to do with TCP keep alives along the lines of the KeepAlive
+		// variable in net.Dialer.
+		DisableKeepAlives: persistent,
```
fffw: From https://golang.org/src/net/http/transport.go#L361 , if persistent is true, `http.Transport` just closes the connection when it's idle. Should we use `!persistent` instead?

myleshorton: Great catch @fffw! For some reason I don't seem to get these messages from you -- probably something off with my email filters, so I'm just seeing this now. Fixing.

myleshorton:
OK guys I'm going to merge this because otherwise we'll start to experience conflicts with new changes coming down the pipeline.

Issue #2517, opened 2015-05-14
Change config URL to one that we can serve dynamically


aranhoide:
We want to transition to a system where the configuration is served dynamically for each user.  In principle we do this so we can serve different *chained servers* to different clients, but serving the whole configuration dynamically gives us the flexibility to customize any other piece of configuration.

In order to do this transition seamlessly, we want to serve the config through an URL that we can point to a web server of our own.  Until such web server is ready, we want to point that URL to a S3 bucket.

While we could change old clients to use the new configuration by just overriding their existing cloud.yamls, we also want the config URL that comes hardcoded into new clients to point to the new URL.  This issue is just about overriding this.

I made [a separate ticket](https://github.com/getlantern/lantern/issues/2519) about the changes required in genconfig, because those changes must be made on top of an as-of-yet-unmerged branch.

PR #2493, opened 2015-05-08
streamline updating of country-specific cloud.yamls, fixes #2454

aranhoide:

PR #2518, opened 2015-05-14
change cloud config URL to one that can be served dynamically; fixes …

aranhoide:
…#2517

There is a CNAME entry in Cloudflare pointing to a 'config.getiantem.org' bucket that has the 'cn' and 'default' cloud configs.  This bucket name was required for TLS validation.

I'll modify the `genconfig` part of this next.  It needs to be implemented as a branch of the as-of-yet unmerged https://github.com/getlantern/lantern/pull/2493.  The new bucket name also requires a not-yet-released version of `s3cmd`, because of

https://github.com/s3tools/s3cmd/issues/437

### DIFF
--- a/src/github.com/getlantern/flashlight/config/config.go
+++ b/src/github.com/getlantern/flashlight/config/config.go
@@ -229,7 +229,7 @@ func (cfg *Config) ApplyDefaults() {
 	}
 
 	if cfg.CloudConfig == "" {
-		cfg.CloudConfig = fmt.Sprintf("https://s3.amazonaws.com/lantern_config/cloud.%v.yaml.gz", countryPlaceholder)
+		cfg.CloudConfig = fmt.Sprintf("https://config.getiantem.org/cloud.%v.yaml.gz", countryPlaceholder)
 	}
 
 	// Make sure we always have a stats config


myleshorton:
LGTM.

aranhoide:
Alas, I just realized that while this is necessary it is not enough.  We need to change the client so it will hit the config server through direct domain fronting, just like the geoip server is hit currently, and for the same reason: we need the IP of the cilent, which is not forwarded by flashlight servers.

Another option would be to whitelist some domains so flashlight servers will forward the user's IP to those.  That would work with old clients too.  But that would introduce an unnecessary hop.  So even if we do this for old clients I'd still do direct fronting in new clients.  I'll submit a new PR about this.

aranhoide:
It seems I was misunderstanding geolookup... it's actually passing the IP as a query parameter.  I don't think direct fronting is being used here either?  We could do the same for the config server, although that makes it easier to forge client IPs (?).

myleshorton:
I'm looking at this now @aranhoide

myleshorton:
This is actually super confusing @aranhoide. We are using a fronted.DirectHttpClient() for the geolookup, but we're also using it for analytics and the statserver. See:

https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/flashlight.go#L225

geolookup has a custom cname pointing to heroku on CloudFlare, but the others do not. As such, it's totally unclear how or even if they're actually working.

myleshorton:
I should say that DirectHttpClient does direct fronting. For the lookup of one's own IP, it doesn't pass the query parameter. It passes the query parameter when looking up geo info for other users'/servers' IPs.

aranhoide:
Oh I see.  I had misread that.  I thought that was a regular fronted client going through flashlight.  So maybe the others are working because their domains are frontable through Cloudflare?

myleshorton:
I don't think that can possibly be true for analytics. Statsserver is actually just using it to lookup the geo info for peers/servers to show on the map. @atavism any idea on that? Did you mean to try to use direct domain fronting for that?

myleshorton:
by "that" I mean analytics...

myleshorton:
I'm actually confused on the s3 buckets @aranhoide. Don't we want these to actually go to some server that can dynamically return the config?

myleshorton:
Or is that just a temporary fix to make the client side work @aranhoide?

aranhoide:
@myleshorton: yes, backing the new URL with a S3 bucket was meant as a transition measure.  I'll swap in the config server as soon as it's ready.

myleshorton:
Gotcha @aranhoide. OK cool. I'm looking at the change to do direct fronting for it now -- should have a PR on the same branch shortly.

aranhoide:
Oh I was looking at the same thing.  I can do the PR if you want, or I can keep reading on this as background for reviewing that..

myleshorton:
Either way man. We're probably doing the same thing. I'm probably 10 mins away from being done. Maybe we both just keep going and see if we're doing the same thing =)?

aranhoide:
You're probably closer.  I'm a bit befuddled by the fact that config/config.go is shared between client and server, while currently direct proxying is only being set up for clients.

myleshorton:
Hmmn...ok let me submit what I've got and just see what you think.

aranhoide:
Sounds good!

myleshorton:
This is trickier than I thought...let's chat about it on our call.

PR #2522, opened 2015-05-15
Refactored highest qos fronted dialer handling closes #2520

### LINKED ISSUE #2520: Use direct fronting to load dynamic config

myleshorton:
OK @aranhoide I think this is a little cleaner.

### DIFF
--- a/src/github.com/getlantern/flashlight/client/config.go
+++ b/src/github.com/getlantern/flashlight/client/config.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"math"
 	"sort"
 	"time"
 
@@ -23,6 +24,24 @@ type ClientConfig struct {
 // SortServers sorts the Servers array in place, ordered by host
 func (c *ClientConfig) SortServers() {
 	sort.Sort(ByHost(c.FrontedServers))
+}
+
+// HighestQOSFrontedDialer returns the fronted.Dialer with the highest QOS.
+func (c *ClientConfig) HighestQOSFrontedDialer() fronted.Dialer {
+	var highestQOSFrontedDialer fronted.Dialer
+	highestQOS := math.MinInt32
+	for _, s := range c.FrontedServers {
+		// Get a dialer for domain fronting (fd) and a dialer to dial to arbitrary
+		// addreses (dialer).
+		fd, dialer := s.dialer(c.MasqueradeSets)
+		if dialer.QOS > highestQOS {
+			// If this dialer as a higher QOS than our current highestQOS, set it as
+			// the highestQOSFrontedDialer.
+			highestQOSFrontedDialer = fd
+			highestQOS = dialer.QOS
+		}
+	}
+	return highestQOSFrontedDialer
 }
 
 // ByHost implements sort.Interface for []*ServerInfo based on the host
--- a/src/github.com/getlantern/flashlight/client/config.go
+++ b/src/github.com/getlantern/flashlight/client/config.go
@@ -28,20 +28,18 @@ func (c *ClientConfig) SortServers() {
 
 // HighestQOSFrontedDialer returns the fronted.Dialer with the highest QOS.
 func (c *ClientConfig) HighestQOSFrontedDialer() fronted.Dialer {
-	var highestQOSFrontedDialer fronted.Dialer
+	var hqfsi *FrontedServerInfo
 	highestQOS := math.MinInt32
 	for _, s := range c.FrontedServers {
-		// Get a dialer for domain fronting (fd) and a dialer to dial to arbitrary
-		// addreses (dialer).
-		fd, dialer := s.dialer(c.MasqueradeSets)
-		if dialer.QOS > highestQOS {
+		if s.QOS > highestQOS {
 			// If this dialer as a higher QOS than our current highestQOS, set it as
 			// the highestQOSFrontedDialer.
-			highestQOSFrontedDialer = fd
-			highestQOS = dialer.QOS
+			hqfsi = s
+			highestQOS = s.QOS
 		}
 	}
-	return highestQOSFrontedDialer
+	fd, _ := hqfsi.dialer(c.MasqueradeSets)
+	return fd
 }
 
 // ByHost implements sort.Interface for []*ServerInfo based on the host
--- a/src/github.com/getlantern/flashlight/config/config.go
+++ b/src/github.com/getlantern/flashlight/config/config.go
@@ -10,7 +10,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
-	"strings"
 	"time"
 
 	"github.com/getlantern/appdir"
@@ -21,7 +20,6 @@ import (
 	"github.com/getlantern/yamlconf"
 
 	"github.com/getlantern/flashlight/client"
-	"github.com/getlantern/flashlight/geolookup"
 	"github.com/getlantern/flashlight/globals"
 	"github.com/getlantern/flashlight/server"
 	"github.com/getlantern/flashlight/statreporter"
@@ -33,7 +31,6 @@ const (
 	cloudflare              = "cloudflare"
 	etag                    = "ETag"
 	ifNoneMatch             = "If-None-Match"
-	countryPlaceholder      = "${COUNTRY}"
 )
 
 var (
@@ -218,7 +215,7 @@ func (cfg *Config) ApplyDefaults() {
 	}
 
 	if cfg.CloudConfig == "" {
-		cfg.CloudConfig = fmt.Sprintf("config.getiantem.org/cloud.%v.yaml.gz", countryPlaceholder)
+		cfg.CloudConfig = fmt.Sprintf("https://config.getiantem.org/cloud.yaml.gz")
 	}
 
 	// Make sure we always have a stats config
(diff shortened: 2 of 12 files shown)


src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -365,18 +353,10 @@ func (cfg Config) fetchCloudConfig() (bytes []byte, err error) {
...
 	log.Tracef("fetchCloudConfigForAddr '%s'", proxyAddr)
 
-	if proxyAddr != "" {
-		// Wait for proxy to become available
-		err := waitforserver.WaitForServer("tcp", proxyAddr, 30*time.Second)
-		if err != nil {
-			return nil, fmt.Errorf("Proxy never came up at %v: %v", proxyAddr, err)
-		}
-	}
+	dialer := cfg.Client.HighestQOSFrontedDialer()
```
aranhoide: Two things:

 - The config URL needs be changed to plain HTTP instead of HTTPS (see [this comment](https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/fronted/dialer.go#L54)).  I'll change that.

`config/config.go` is shared between the client and the server.  In the server side, `dialer` becomes `nil` and we get a runtime panic trying to access it.  We could try to somehow create a fronted dialer in the server side.  I haven't checked how hard that would be, but anyway that seems silly since servers need no fronting.  We could somehow pass a `dialer` to this function but it gets hard to reason about whether the dialer will be in sync with the current config.

I think the easiest thing to do would be to have a cfg.GetDirectHttpClient() method like such (from the top of my head; I haven't checked it):

```go
func (cfg *Config) GetDirectHttpClient() (*http.Client, error) {
    if Config.Role == "server" {
        ret, err := util.HTTPClient("", "")
        if err != nil {
            return nil, fmt.Errorf("Couldn't create direct server-side HTTP client: %v", err)
        }
        return ret, nil
    } else {
        return cfg.Client.HighestQOSFrontedDialer().DirectHttpClient(), nil
    }
}        
```

aranhoide:
@myleshorton I made some changes as per [this comment](https://github.com/getlantern/lantern/pull/2522#discussion_r30406424).  I also fixed some yamlconf tests that were calling `StartPolling`.  I agree these changes are an improvement overall.  Feel free to merge after you've reviewed [my commit](https://github.com/getlantern/lantern/commit/1c138aa517f6474d037e31af85e287568b1138d5) to fix the server side.

aranhoide:
Hm, this seems to break lantern-android too.  I'll look into that.

myleshorton:
Nice! I like the approach. I'm tied up for another 45 min but will look
then and likely just merge it.

aranhoide:
In https://github.com/getlantern/lantern/commit/205a43c9ebfdfaa001eadf06fcab006973da4428 I meant "lantern-android tests" insead of "lantern-config tests"...

src/github.com/getlantern/flashlight/client/config.go
```diff
@@ -25,6 +26,24 @@ func (c *ClientConfig) SortServers() {
...
 }
 
+// HighestQOSFrontedDialer returns the fronted.Dialer with the highest QOS.
+func (c *ClientConfig) HighestQOSFrontedDialer() fronted.Dialer {
+	var highestQOSFrontedDialer fronted.Dialer
+	highestQOS := math.MinInt32
+	for _, s := range c.FrontedServers {
+		// Get a dialer for domain fronting (fd) and a dialer to dial to arbitrary
+		// addreses (dialer).
+		fd, dialer := s.dialer(c.MasqueradeSets)
```
aranhoide: A lot of `fronted.Dialer`s are created and discarded here.  Only the one that is returned has a chance of ever being `Close`d.  The code prior to this was storing the `hqfd` in the client so it could be closed.  Maybe these only need closing if they're ever used?

aranhoide:
There are still problems with this.  We are not closing what used to be `client.hqfd`.  I'm not sure why we need to close that, and whether that applies to all the `fronted.Dialer`s that are [created and immediately discarded](https://github.com/getlantern/lantern/pull/2522/files#diff-cb59cf23692ec32aadfa1e50c2ee60e4R36) in `HighestQOSFrontedDialer`.

aranhoide:
Unless I misunderstood something, we were leaking a helluva lot of `fronted.Dialer`s; all the ones that were considered and discarded in HighestQOSFrontedDialer.  We are still leaking all the dialers that are created whenever we reconfigure the client.

Even if we fix the leakage, each of these Dialers starts a goroutine to verify its masquerades, and keeps a pool of (currently configured as 30) connections.  Having more than one of these running at a given moment feels wasteful to me.  I think we better have a scheme where we only have a single one of these Dialers per Client and reuse it until we reconfigure the client, at which point we Close it and open a new one.

aranhoide:
I'll make the changes I propose in a different branch and make a PR into this one.

myleshorton:
> Even if we fix the leakage, each of these Dialers starts a goroutine to verify its masquerades, and keeps a pool of (currently configured as 30) connections. Having more than one of these running at a given moment feels wasteful to me. I think we better have a scheme where we only have a single one of these Dialers per Client and reuse it until we reconfigure the client, at which point we Close it and open a new one.

Sounds right -- yeah I hadn't noticed that call to NewDialer goes all the way down into the masquerade verification code....

myleshorton:
@aranhoide dude sorry coding in go still takes more for friggin' ever. This fixes at least part of it.

aranhoide:
@myleshorton: np.  I'll take a look at this and look at how best to avoid leaking dialers and generally having too many of them around.  So I'll probably reply with yet more commits. :)

aranhoide:
@myleshorton your changes all LGTM.  I updated lantern-android and made https -> http replacement slightly safer.

The only reason not to merge this right now is that we're no longer closing the Dialer on Client.Stop().  This is never called in the flashlight project, but is called in lantern-android.  To be precise, there is a call to it in MobileClient.Stop(); I'm not sure whether that one is ever called.   

So I'm reading the code that uses direct fronting to try and find the cleanest way to have them all use the same dialer (and clean it up on client reconfiguration or Stop).

myleshorton:
Yeah I totally agree @aranhoide that's an issue on the closing dialers front. It seems like in theory somewhere we should just be able to sort the existing dialers by QoS and grab the highest QoS one. I agree the cleanest way to do it isn't obvious, however, and the current code for me actually chews through all available file descriptors pretty quickly and starts spitting ```too many open file``` errors, so I completely agree this isn't ready to merge.

myleshorton:
It seems like going back to the concept of pulling the config fetching out of config.go and getting the highest QoS dialer from the balancer might be the way to go. The problem is there's a bit of a chicken and egg issue in terms of setting up the poll function in config.go before we've created the fetcher, but I think we can handle that.

aranhoide:
I already [fixed](https://github.com/getlantern/lantern/commit/e6fe6445fdefde6fe7642e673c851fdce9811df5) the issue of creating unused dialers when choosing the one with the highest QoS.  Turns out that the QoS of the dialer is just that of the corresponding FrontedServerInfo, so I just found the highest QoS FrontedServerInfo and created that dialer alone.

I'm now working out the problem of cleaning up the Dialers.

myleshorton:
Oh right I forgot you did that -- cool. I *think* I still actually get those too many open file errors though -- I'll test again.

aranhoide:
@myleshorton yes, as long as we are not closing the old dialer we are leaking open connections (and hence file descriptors) on every config update.

myleshorton:
Makes sense @aranhoide. I am surprised they build up that quickly I must say. Why do the dialers themselves keep those connections open at all actually? Just to keep them warm?

aranhoide:
IIUC yes, that's the purpose.

aranhoide:
@myleshorton ^.  That said, it's been a while since I last looked at the connection pool code.

aranhoide:
@myleshorton : and I'm not claiming that this leakage fully explains the too-many-open-files errors.  Let's see how quickly these build up after we properly clean up the Dialers.

aranhoide:
A challenge in cleaning the Dialers as the code stands is that there is a possible race condition where services keep references to the `http.Client`s (fronters) obtained from them.  Although Dialer.Close() takes care to wait for all *ongoing* connections, it can't really know that e.g. geolookup has a `client` that might still get used after all the connections have been closed.  Even if we update the client in all services before we call Dialer.Close, and even considering that most services hold the http.Client in an atomic.Value, there is still a potential race condition in the gap between the time the value is `Load`ed and the time it is actually used.

I think I know how to fix this without using any globals.  After you read it you may want to introduce one global or two to make it easier to follow.  :)  Or maybe we find some other way.  We'll see.

aranhoide:
@myleshorton : while your DirectDomainTransport change passes flashlight tests, it seems to cause problems when trying to run it.  I get a lot of errors like this and I can't get geolocated.

```
May 16 05:19:24.708 - flashlight.geolookup: geolookup.go:117 Unable to get current location: Could not get response from server: "net/http: Client Transport of type *fronted.DirectDomainTransport doesn't support CancelRequest; Timeout not supported"
```

aranhoide:
@myleshorton : nevermind, fixed that by just adding a trivial wrapper over Transport.CancelRequest.

aranhoide:
[This sub-PR](https://github.com/getlantern/lantern/pull/2529) fixes all the Dialer leakage that I talked about for the regular Lantern client.  Some issues may remain for lantern-android.  I'll look into that tomorrow.

myleshorton:
Nice. I still need to review the PR, but a couple of things came to mind:

1. For DDF we don't really need a huge connection pool typically. Why not just create a new dialer with no pool just for this purpose or that takes the pool size as an argument?
1. Why can't we just close dialers when we get new configs, such as in flashlight.go and config.go? I imagine there's something that causes that not to work, but if not it seems pretty straightforward.

aranhoide:
1. Hm yes, I agree for current uses of DDF we probably don't need a pool.  On creation, a Dialer [still](https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/fronted/dialer.go#L143) [launches](https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/fronted/masquerade.go#L72) a bunch of goroutines/connections to verify masquerades.  This can probably be fixed or ameliorated too; it's a matter of how much we want to specialize for this particular usage.  Consider that the same problems apply for non-direct-fronted Dialers, which brings me to..

2. AFAICT Dialer.Close() immediately closes all ongoing connections in the connection pool.  For some reason, I was under the impression that it would wait for existing transfers to end, but I traced that path again and I no longer think that's the case.  OTOH, I also thought that users would be out of luck if using the pool after it's been closed, but I just traced that path and I think the Dialer will go on to work pretty much like an unpooled one (with the exception that it will be stuck with whichever masquerades had already been verified at the time of closing.)

aranhoide:
In summary: we probably don't need a full Dialer for DDF, so we could make a special case for these and that would be simpler than the withClient scheme.  But I think we'll need the withClient scheme (or something else that similarly allows us to Close Dialers more gracefully than what Dialer.Close is doing now) anyway for other (non direct) domain fronted Dialers anyway, so in the long run the special case won't buy us much.

Note also that in principle we could use the same Dialer for all direct and non-direct requests.

aranhoide:
https://github.com/getlantern/lantern/pull/2538 includes the most useful parts of this without some of the problematic ones.

aranhoide:
Fixed in #2538.

PR #2523, opened 2015-05-15
ui.Show() wait for url in a go routine closes #2513

### LINKED ISSUE #2513: IE 11-  Opens the browser but displays that "The page cannot be displayed" while opening on system start up
Steps:

1. Make IE as default browser.
2. Open lantern, change the setting to run lantern on system start up.
3. Restart the system.
4. Wifi connection is up before the lantern browser opens but still says that the page cannot be displayed. However, if I use show lantern to open the browser again, it works fine.

Note: Am unable to use the tab that is opening at the start up, it says "The page cannot be displayed" for all the websites. Once I start using a new tab, it is working fine.

fffw:
Fixes https://github.com/getlantern/lantern/issues/2513

`ui.Start()` runs `http.Server` in separate goroutine, if that goroutine didn't get chance to run before `ui.Show()`, this issue will happen.

`WaitForServer()` for at most 3 seconds to resolve it.

### DIFF
--- a/src/github.com/getlantern/flashlight/ui/ui.go
+++ b/src/github.com/getlantern/flashlight/ui/ui.go
@@ -4,11 +4,14 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 	"path/filepath"
 	"runtime"
+	"time"
 
 	"github.com/getlantern/golog"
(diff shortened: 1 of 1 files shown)


atavism:
This LGTM.

Issue #2497, opened 2015-05-08
Add back proxy everything mode


myleshorton:
A number of users have requested this and are using the 1.5.17 version instead of 2.0 because it supports it. I think we should just add it back in!

sheena3:
@myleshorton Will we remove auto-detection if we proxy everything?

fffw:
@sheena3 I think we can just skip auto-detection(detour) when user checks a "proxy everything" option.

@myleshorton From your impression, were users requesting it because auto-detection didn't work well, or because of security concerns?

myleshorton:
Hey sorry for the delay guys, but it's not totally clear why users keeps asking for it -- I've asked them those same questions without a clear or consistent response. In that one vk.com case it was because detour wasn't working though, although we still haven't discovered why.

atavism:
Yeah, we discussed this before, but I think it would be great to make these different modes configurable from the settings screen. We'd have auto-detection (to turn detour on/off) and "proxy everything" options. The UI changes probably merit a separate ticket.

myleshorton:
Agreed @atavism. To me initially it might just be proxy everything or not, with not just being detour, but not sure.

fffw:
Added an yaml option in https://github.com/getlantern/lantern/pull/2524

For UI part, do you guys prefer place it in setting screen or tray menu? An tray menu item might be more convenient to user, though a bit more code is required.

@atavism Your thoughts about adding an auto-detection check option? I can't think of a situation when user want just fixed whitelist but no auto-detection.

myleshorton:
Nice! Not sure on settings vs. tray menu. A check box in the tray could be challenging cross-platform but maybe not? In general it would be cool to move as much as possible into the tray. If you can get that working, that would be cool for sure.

fffw:
Checked menu is not implemented in systray on Linux yet. Considering the priority of other issues, I'll add the option to setup page for now.

fffw:
Resolved by https://github.com/getlantern/lantern/pull/2524

Issue #2500, opened 2015-05-11
balancer.Dialer panics "send on closed channel"


fffw:
Encountered while running Lantern with my changes of detour. But this bug may not directly triggered by my changes. Worth take a look.

```
panic: send on closed channel

goroutine 8790 [running]:
github.com/getlantern/balancer.(*dialer).onError(0xc2081abca0, 0x4f3bcc0, 0xc20b7015b0)
	/Users/fffw/workspace/getlantern/lantern/src/github.com/getlantern/balancer/dialer.go:111 +0x6c
github.com/getlantern/balancer.(*Balancer).DialQOS(0xc208a2b980, 0x44f6130, 0x3, 0xc209e4c470, 0xd, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/fffw/workspace/getlantern/lantern/src/github.com/getlantern/balancer/balancer.go:95 +0x5e0
github.com/getlantern/balancer.(*Balancer).Dial(0xc208a2b980, 0x44f6130, 0x3, 0xc209e4c470, 0xd, 0x0, 0x0, 0x0, 0x0)
	/Users/fffw/workspace/getlantern/lantern/src/github.com/getlantern/balancer/balancer.go:105 +0x89
github.com/getlantern/balancer.*Balancer.Dial·fm(0x44f6130, 0x3, 0xc209e4c470, 0xd, 0x0, 0x0, 0x0, 0x0)
	/Users/fffw/workspace/getlantern/lantern/src/github.com/getlantern/flashlight/client/reverseproxy.go:47 +0x79
github.com/getlantern/detour.func·016()
	/Users/fffw/workspace/getlantern/lantern/src/github.com/getlantern/detour/detour_conn.go:23 +0x5e
created by github.com/getlantern/detour.DialDetour
	/Users/fffw/workspace/getlantern/lantern/src/github.com/getlantern/detour/detour_conn.go:30 +0x16e
```

fffw:
Okay, it's due to a race condition. When a `Balancer` is closed, its dialers will stop together with their `errCh`. In the same or later time,  `Balancer.DialQOS()` calls one of the closed dialer's `onError()`, hence sends on its `errCh`.

Issue #2480, opened 2015-05-06
Reverts back to default settings when lantern is opened newly or using "show lantern"


sheena3:
Steps:

1. Change the settings by unchecking the "Securely report anonymous usage statistics to contribute to Lantern more info" and then close.
2. Now open a new lantern browser. Default setting appears i.e., the checkbox for "Securely report anonymous usage statistics to contribute to Lantern more info" is ticked.

User settings is not saved.

fffw:
I can't reproduce it follow the steps on beta 5. When I unchecked the option, it remains unchecked on newly opened page through "Show Lantern", or the one opened after Lantern restart.

The only occurrence is when I open multiple page, uncheck the option on one page, it is still checked on other pages, e.g., it doesn't get auto-refreshed. I think it's acceptable.

sheena3:
That is not what I am seeing. I rechecked this.

If it would help, I am testing in windows 8.1 and using chrome browser.

sheena3:
@fffw This happens only in chrome browser. In IE and Firefox, it works as expected.

myleshorton:
I also can't reproduce this either on Chrome or FireFox @sheena3. I wonder if it might be an issue with bad permissions from a previous Lantern install. Can you try completely deleting the folder at:

c:\Users\your_user_name\AppData\Roaming\Lantern

and running again?

sheena3:
Yes, that worked. But I still feel that this is a bug. How would the user know that they have to delete the Lantern folder completely before installing the new version?

myleshorton:
The only reason you would have to delete the folder is if you had installed a previous, unreleased version of Lantern that used admin privileges on that folder, making it impossible for newer versions to overwrite the file to save your settings. 

Very weird you're still seeing it with beta 5 though. I'm still unable to reproduce that. Maybe send me your logs?

myleshorton:
Wait never mind @sheena3 -- I was just able to make this happen as well. I'll work on it.

myleshorton:
This issue is really inconsistent -- sometimes it happens and sometimes it doesn't my guess is there's some sort of race condition on startup. 

To check if it's working relatively easily, you can do:

```
tail -f ~/Library/Logs/Lantern/lantern.log | grep Apply
```

while running. Each time you toggle the settings, you should see two Apply logs. When it doesn't work, you don't see those. Punting to beta 7 since this is clearly an issue but not as important as some of the other issues on the table associated with core functionality.

[...]

PR #2524, opened 2015-05-15
add ProxyAll client config option to yaml and setting screen #2497

fffw:
To verify, manually add an 'proxyall' option under 'client' section in lantern.yaml.


fffw:
Oh wait!  When I change proxyall option from true to false when Lantern is running, it triggers "send on closed channel" panic described in https://github.com/getlantern/lantern/issues/2500 . Good news is --- it's reproducible.

Changing this option from false to true works without causing panic.

myleshorton:
> Oh wait! When I change proxyall option from true to false when Lantern is running, it triggers "send on closed channel" panic described in #2500 . Good news is --- it's reproducible.

Nice!

myleshorton:
OK I'm on this one @fffw. I think this will be the primary new feature in beta 6 actually.

myleshorton:
I believe I'm experiencing #2480 with this change as well -- I basically select to proxy all, and that setting never gets saved to disk, and Lantern doesn't proxy all. I suspect it's something intermittent and possibly a race condition on startup. I reproduced this on OSX.

xiam:
Hello guys.

Any chance we could also enable the proxy-all feature from a command line option? It would really help us spotting and fixing some important issues.

Thanks!

myleshorton:
Yeah I agree that's useful for debugging.

myleshorton:
OK this generally looks good to me @fffw, although there is a bug with no longer unsetting the pac file on exit. We could just make this more explicit with the following patch to flashlight.go:

```
diff --git src/github.com/getlantern/flashlight/flashlight.go src/github.com/getlantern/flashlight/flashlight.go
index 0d8b1a0..8a2dd8e 100644
--- src/github.com/getlantern/flashlight/flashlight.go
+++ src/github.com/getlantern/flashlight/flashlight.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/getlantern/fronted"
@@ -43,7 +44,8 @@ var (
 	configUpdates = make(chan *config.Config)
 	exitCh        = make(chan error, 1)
 
-	showui = true
+	showui    = true
+	exitFuncs = make([]func(), 0)
 )
 
 func init() {
@@ -233,11 +235,18 @@ func runClientProxy(cfg *config.Config) {
 	watchDirectAddrs()
 
 	go func() {
-		defer pacOff()
+		addExitFunc(func() {
+			pacOff()
+		})
 		exit(client.ListenAndServe(pacOn))
 	}()
 }
 
+// addExitFunc adds a function to be called before the application exits.
+func addExitFunc(exitFunc func()) {
+	exitFuncs = append(exitFuncs, exitFunc)
+}
+
 func applyClientConfig(client *client.Client, cfg *config.Config) {
 	autoupdate.Configure(cfg)
 	logging.Configure(cfg, version, buildDate)
@@ -336,6 +345,10 @@ func useAllCores() {
 // exit tells the application to exit, optionally supplying an error that caused
 // the exit.
 func exit(err error) {
+	for _, ex := range exitFuncs {
+		log.Debugf("Calling exit func")
+		ex()
+	}
 	exitCh <- err
 }
 
```

myleshorton:
Hey @fffw I realized this should actually use a channel instead of a slice - both simpler and thread safe!

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -98,7 +98,6 @@ func doMain() error {
 
 	// Schedule cleanup actions
 	defer logging.Close()
-	defer pacOff()
```
fffw: It's me introduced the bug of not turning off pac by moving this line to below to near the call to `pacOn`. We can move it back here, or wrap the defer in a func, but I think add `exitFuncs` solely for this purpose is not worthy. How do you think @myleshorton ?

fffw:
Oh let me double check if it actually works!

src/github.com/getlantern/flashlight/flashlight.go
myleshorton: That defer only worked when the method blocked, which the new one doesn't do. I think either way is fine though.

fffw:
@myleshorton Used channel for exit functions as per your suggestion. Please take a look once you have time, thanks.

myleshorton:
LGTM thanks @fffw!

PR #2529, opened 2015-05-16
withClient

aranhoide:
See [this comment](https://github.com/getlantern/lantern/commit/e34ddc94168b7410886f8e2a16d257eafe993f1b#diff-f51f873318b593f0a492e1abb11a57c2R96).

I actually like this approach although I admit it's on the terse side.

One thing it doesn't try to do (yet) is to close the Dialer when the Client is closed.  Per se, that would be easy to do, but the ramifications are unclear to me.  I think we would need to stop a few loops that are currently going on forever, e.g., [this one](https://github.com/getlantern/lantern/blob/e34ddc94168b7410886f8e2a16d257eafe993f1b/src/github.com/getlantern/flashlight/flashlight.go#L309).  

Note that this doesn't matter for the regular lantern client, only for lantern-android.  I'll need to take a deeper look at why lantern-android wants to be able to Stop the client without exiting the app.

### DIFF
--- a/src/github.com/getlantern/flashlight/config/config.go
+++ b/src/github.com/getlantern/flashlight/config/config.go
@@ -23,7 +23,6 @@ import (
 	"github.com/getlantern/flashlight/globals"
 	"github.com/getlantern/flashlight/server"
 	"github.com/getlantern/flashlight/statreporter"
-	"github.com/getlantern/flashlight/util"
 )
 
 const (
@@ -65,7 +64,7 @@ type CA struct {
 }
 
 // Init initializes the configuration system.
-func Init() (*Config, error) {
+func Init(withClient func(func(*http.Client))) (*Config, error) {
 	configPath, err := InConfigDir("lantern.yaml")
 	if err != nil {
 		return nil, err
@@ -95,7 +94,7 @@ func Init() (*Config, error) {
 			}
 
 			var bytes []byte
-			bytes, err = cfg.fetchCloudConfig()
+			bytes, err = cfg.fetchCloudConfig(withClient)
 			if err == nil && bytes != nil {
 				mutate = func(ycfg yamlconf.Config) error {
 					log.Debugf("Merging cloud configuration")
@@ -328,25 +327,11 @@ func (cfg Config) cloudPollSleepTime() time.Duration {
 	return time.Duration((CloudConfigPollInterval.Nanoseconds() / 2) + rand.Int63n(CloudConfigPollInterval.Nanoseconds()))
 }
 
-func (cfg Config) fetchCloudConfig() (bytes []byte, err error) {
+func (cfg Config) fetchCloudConfig(withClient func(func(*http.Client))) (bytes []byte, err error) {
 	log.Debugf("Fetching cloud config...")
-
-	if cfg.IsDownstream() {
-		// Clients must always proxy the request
-		dialer := cfg.Client.HighestQOSFrontedDialer()
-		// We use direct domain fronting for accessing new configs.
-		client := dialer.NewDirectDomainFronter()
-		bytes, err = cfg.fetchCloudConfigForClient(client)
-	} else {
-		var client *http.Client
-		// In the server side, we use a direct (non-proxied) client.
-		if client, err = util.HTTPClient(cfg.CloudConfigCA, ""); err == nil {
-			// We need to ask for HTTPS explicitly for these.
-			bytes, err = cfg.fetchCloudConfigForClient(client)
-		} else {
-			err = fmt.Errorf("Unable to initialize HTTP client: %s", err)
-		}
-	}
+	withClient(func(c *http.Client) {
+		bytes, err = cfg.fetchCloudConfigForClient(c)
+	})
 	if err != nil {
 		bytes = nil
 		err = fmt.Errorf("Unable to fetch cloud config: %s", err)
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -5,9 +5,11 @@ import (
 	"flag"
 	"fmt"
 	"math/rand"
+	"net/http"
 	"os"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/getlantern/fronted"
@@ -27,6 +29,7 @@ import (
 	"github.com/getlantern/flashlight/statreporter"
 	"github.com/getlantern/flashlight/statserver"
 	"github.com/getlantern/flashlight/ui"
+	"github.com/getlantern/flashlight/util"
 )
 
 var (
@@ -89,6 +92,45 @@ func _main() {
 	os.Exit(0)
 }
 
+type clientWithCloseThunk struct {
+	client *http.Client
+	close  func()
+}
+
+type clientMaker struct {
+	make  func() clientWithCloseThunk
+	close func()
+}
+
+type makerChan chan clientMaker
+
+func newMakerChan() makerChan {
+	return make(chan clientMaker, 1)
+}
+
+func (ch makerChan) updateMaker(c clientMaker) (ret clientMaker) {
+	select {
+	case ret = <-ch:
+	default:
+	}
+	ch <- c
+	return ret
+}
+
+func (ch makerChan) getMaker() clientMaker {
+	ret := <-ch
+	ch <- ret
+	return ret
+}
+
+func (ch makerChan) makeWithClient() func(func(*http.Client)) {
+	return func(f func(*http.Client)) {
+		cc := ch.getMaker().make()
+		defer cc.close()
+		f(cc.client)
+	}
+}
+
 func doMain() error {
 	err := logging.Init()
 	if err != nil {
@@ -111,7 +153,9 @@ func doMain() error {
 
 	parseFlags()
 
-	cfg, err := config.Init()
+	mch := newMakerChan()
+
+	cfg, err := config.Init(mch.makeWithClient())
 	if err != nil {
 		return fmt.Errorf("Unable to initialize configuration: %v", err)
 	}
@@ -140,9 +184,9 @@ func doMain() error {
 	log.Debug("Running proxy")
 	if cfg.IsDownstream() {
 		// This will open a proxy on the address and port given by -addr
-		runClientProxy(cfg)
+		runClientProxy(cfg, mch)
 	} else {
-		runServerProxy(cfg)
+		runServerProxy(cfg, mch)
 	}
 
 	return waitForExit()
@@ -179,7 +223,7 @@ func parseFlags() {
 }
 
 // runClientProxy runs the client-side (get mode) proxy.
-func runClientProxy(cfg *config.Config) {
+func runClientProxy(cfg *config.Config, mch makerChan) {
 	var err error
 
 	// Set Lantern as system proxy by creating and using a PAC file.
@@ -215,17 +259,11 @@ func runClientProxy(cfg *config.Config) {
 	settings.Configure(cfg, version, buildDate)
 	proxiedsites.Configure(cfg.ProxiedSites)
 
-	hqfd := cfg.Client.HighestQOSFrontedDialer()
-	if hqfd == nil {
-		log.Errorf("No fronted dialer available, not enabling geolocation, stats or analytics")
-	} else {
-		// An *http.Client that uses the highest QOS dialer.
-		hqfdClient := hqfd.NewDirectDomainFronter()
-
-		geolookup.Configure(hqfdClient)
-		statserver.Configure(hqfdClient)
-		analytics.Configure(cfg, false, hqfdClient)
-	}
+	updateClientDirectFronter(cfg, mch)
+	wdc := mch.makeWithClient()
+	geolookup.Configure(wdc)
+	statserver.Configure(wdc)
+	analytics.Configure(cfg, false, wdc)
 
 	// Continually poll for config updates and update client accordingly
 	go func() {
@@ -238,20 +276,11 @@ func runClientProxy(cfg *config.Config) {
 
 			client.Configure(cfg.Client)
 
-			hqfd = cfg.Client.HighestQOSFrontedDialer()
+			updateClientDirectFronter(cfg, mch)
 
-			if hqfd != nil {
-				// Create and pass the *http.Client that uses the highest QOS dialer to
-				// critical modules that require continual comunication with external
-				// services.
-				hqfdClient := hqfd.NewDirectDomainFronter()
-
-				geolookup.Configure(hqfdClient)
-				statserver.Configure(hqfdClient)
-				settings.Configure(cfg, version, buildDate)
-				logging.Configure(cfg, version, buildDate)
-				autoupdate.Configure(cfg)
-			}
+			settings.Configure(cfg, version, buildDate)
+			logging.Configure(cfg, version, buildDate)
+			autoupdate.Configure(cfg)
 		}
 	}()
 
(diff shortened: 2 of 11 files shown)


aranhoide:
I haven't updated lantern-android to use this yet.  If you like this approach I'll get to that next.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -328,25 +327,11 @@ func (cfg Config) cloudPollSleepTime() time.Duration {
...
-		var client *http.Client
-		// In the server side, we use a direct (non-proxied) client.
-		if client, err = util.HTTPClient(cfg.CloudConfigCA, ""); err == nil {
-			// We need to ask for HTTPS explicitly for these.
-			bytes, err = cfg.fetchCloudConfigForClient(client)
-		} else {
-			err = fmt.Errorf("Unable to initialize HTTP client: %s", err)
-		}
-	}
+	withClient(func(c *http.Client) {
```
aranhoide: While `withClient` is always the same function for the lifetime of the Lantern client, `c` is always the latest fronter.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -89,6 +93,81 @@ func _main() {
...
+//
+// The function returned by this call is meant to be used like this:
+//
+// withClient(func(c *http.Client) {
+//    ... use `c` here ...
+// })
+//
+// No explicit cleanup is required in the body where `c` is used, but `c` should
+// never be used after the body has returned.  So don't assign `c` to variables
+// or data structures outside the body, don't use it inside a goroutine, etc.
```
aranhoide: That is, inside a goroutine created within the body.  This is possibly dangerous:

```go
withClient(func(c *http.Client) {
  go fetchConfig(c)
})
```
but this is fine
```go
go func() {
  withClient(func(c *http.Client) {
    fetchConfig(c)
  })
}
```
}

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -215,22 +296,19 @@ func runClientProxy(cfg *config.Config) {
...
-		hqfdClient := hqfd.NewDirectDomainFronter()
-
-		geolookup.Configure(hqfdClient)
-		statserver.Configure(hqfdClient)
-		analytics.Configure(cfg, false, hqfdClient)
-	}
+	updateClientDirectFronter(cfg, mch)
+	wdc := mch.makeWithClient()
+	geolookup.Configure(wdc)
+	statserver.Configure(wdc)
```
aranhoide: I don't think statserver nor analytics really needs a direct fronted dialer, but one change at a time...

myleshorton: statserver does -- it's used for geolookup for *local* stats on peers. I think analytics actually probably breaks with using DDF though.

aranhoide: Local in what sense?  Is the client's own IP used at all if the query includes an IP?  That is, if I do a geolookup query from my home to request the City of a peer in NY do I get a different response than if you do a query from your home?

Where is the server side code for geo.getiantem.org?

aranhoide:
@myleshorton I updated lantern-android to use this scheme to clean up direct fronted dialers.  This branch also includes two unrelated fixes:

- Include the trusted CAs in the built-in config, which is useless without this.  lantern-android was relying on an initial unfronted fetch of the online config to bootstrap its fronted configuration.

- Fix a test that was capturing ranged-over variables in a closure.  

So even if we don't merge this, these fixes had better be cherry-picked into whatever we eventually rely on.

I think this is it for the client side.  IIUC we are still leaking non-direct fronted Dialers, but that's been the case for a while and we can make a different PR for that (using withClient if we merge this).

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -29,18 +29,12 @@ var (
...
 func init() {
-	var err error
-	// Initial attempt to get configuration, without a proxy. If this request
-	// fails we'll use the default configuration.
-	if clientConfig, err = getConfig(); err != nil {
-		// getConfig() guarantees to return a *Config struct, so we can log the
-		// error without stopping the program.
-		log.Printf("Error updating configuration over the network: %q.", err)
-	}
+	clientConfig = defaultConfig()
```
aranhoide: I don't think we really need this global, but one thing at a time...

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -79,11 +73,14 @@ func NewClient(addr, appName string) *MobileClient {
 		}
 	}
 
-	analytics.SessionEvent(hqfdc, sessionPayload)
+	mch.MakeWithClient()(func(c *http.Client) {
+		analytics.SessionEvent(c, sessionPayload)
```
aranhoide: If we don't really need to direct front to analytics, then we don't need the `mch` here and thus we don't need to store it in the `MobileClient` at all.

aranhoide:
The pattern that this uses has been used for a while in various languages.  Here it is explained in the context of Go, although the author seems to miss a couple caveats:

http://carefulcode.blogspot.com.es/2013/05/introduction-to-rifl.html

http://carefulcode.blogspot.com.es/2013/05/rifl-examples.html

Go has finalizers, so this could be implemented more directly.  The problem with them is they're tied to the garbage collector, which is conservative, so there are few guarantees about when they'll be run.  There seems to be some consensus that they should not be relied upon.  There was even talk about deprecating them, although apparently that won't happen:

https://groups.google.com/forum/#!topic/golang-dev/DMiUkpS1uyQ

myleshorton:
Hey @aranhoide I haven't yet had a chance to review all of this, but I just wanted to quickly say I really like the general direction. It's certainly an ambitious change, but I feel like we've already made the decision to dive deep and do this right, so to me ambitious is fine, and overall the strength of your approach is that it's significantly less error prone. Good stuff, and hopefully I'll have time to review it in full tomorrow.

src/github.com/getlantern/flashlight/withclient/withclient.go
```diff
@@ -0,0 +1,138 @@
...
+	// Generates structures like the one above.  A closure over a fronted.Dialer.
+	make func() clientWithCloseThunk
+	// Placeholder for any cleanup required when this clientMaker is no longer
+	// current because we have got a new one (by a config update).
+	close func()
+}
+
+// To synchronize access to the current clientMaker.  Used as a promise that
+// can be updated.
+type MakerChan chan clientMaker
```
myleshorton: Neat! Didn't realize you could create a chan type.

src/github.com/getlantern/flashlight/withclient/withclient.go
```diff
@@ -0,0 +1,138 @@
...
+	client *http.Client
+	// Placeholder for any cleanup required when this reference is no longer
+	// needed.  We keep track of currently used references with a WaitGroup,
+	// and we use this function to signal we're Done() with this one.
+	close func()
+}
+
+type clientMaker struct {
+	// Generates structures like the one above.  A closure over a fronted.Dialer.
+	make func() clientWithCloseThunk
```
myleshorton: Maybe a name other than make just to avoid confusion with the make builtin function?

aranhoide: Good point; I'll change it.

src/github.com/getlantern/flashlight/withclient/withclient.go
```diff
@@ -0,0 +1,138 @@
...
+	// Placeholder for any cleanup required when this clientMaker is no longer
+	// current because we have got a new one (by a config update).
+	close func()
```
myleshorton: Maybe the same with close to avoid confusion with the close builtin too?

src/github.com/getlantern/flashlight/withclient/withclient.go
```diff
@@ -0,0 +1,138 @@
...
+// use mostly placeholders in the server side.  Because of this and for
+// agility of exposition, the comments below refer to the client side.
+
+type clientWithCloseThunk struct {
+	// A particular reference to the directly fronted http.Client (aka fronter).
+	client *http.Client
+	// Placeholder for any cleanup required when this reference is no longer
+	// needed.  We keep track of currently used references with a WaitGroup,
+	// and we use this function to signal we're Done() with this one.
+	close func()
```
myleshorton: Here too I would think. Close is a nice name, but the fact that it's private means again it could be confusing with the close builtin.

aranhoide: Yes, I've renamed this to `release` which is more accurate anyway.

src/github.com/getlantern/flashlight/withclient/withclient.go
aranhoide: OK, will do.

src/github.com/getlantern/flashlight/withclient/withclient.go
```diff
@@ -0,0 +1,138 @@
...
+//
+// No explicit cleanup is required in the body where `c` is used, but `c` should
+// never be used after the body has returned.  So don't assign `c` to variables
+// or data structures outside the body, don't use it inside a goroutine, etc.
+func (ch MakerChan) MakeWithClient() func(func(*http.Client)) {
+	return func(f func(*http.Client)) {
+		cc := ch.getMaker().make()
+		defer cc.close()
+		f(cc.client)
+	}
```
myleshorton: Love this.

src/github.com/getlantern/flashlight/withclient/withclient.go
```diff
@@ -0,0 +1,138 @@
...
+			make: func() clientWithCloseThunk {
+				wg.Add(1)
+				return clientWithCloseThunk{
+					client: hqfdClient,
+					close:  wg.Done,
+				}
+			},
+			close: func() {
+				wg.Wait()
+				hqfd.Close()
```
myleshorton: Should we also call close on ```hqfdClient``` here?

aranhoide: Good catch; I didn't know that http.Clients needed closing.

myleshorton: Actually I think I misread the last function of https://golang.org/src/net/http/client.go -- doesn't look it actually can be closed.

src/github.com/getlantern/flashlight/withclient/withclient.go
```diff
@@ -0,0 +1,142 @@
...
+		clientRefMaker{
+			newClientRef: func() clientRef {
+				wg.Add(1)
+				return clientRef{
+					client:  hqfdClient,
+					release: wg.Done,
+				}
+			},
+			close: func() {
+				wg.Wait()
```
aranhoide: I just realised there's a race condition here.  Concurrency is tricky!  :)  I'll submit a fix for that.

myleshorton: I think you're saying that because wg.Add and will get called and then immediate wg.Done will get called in MakeWithClient possibly before close is ever called, is that correct?

I'm actually unclear on what the purpose of the WaitGroup is here. The important thing is that close *is called at all* not necessarily that it completes as far as I can tell. What am I missing there?

aranhoide: The WaitGroup is there to wait until all users of the Client are done with it before calling Dialer.Close().  My assumption is that Dialer.Close() abruptly closes all connections in the pool it manages.  I used to believe this was not the case, then I used to believe it was, now I'm not sure anymore; I'll trace the conpool code again. :/

Anyway,
 - if that is the case, then we don't want to close the Dialer until all users of the Client are done with it,
 - if that is not the case, then we don't need the WaitGroup nor most of this package.

So I'll try and make sure which one it is.

myleshorton: Honestly I like the added safety of the package regardless, but I'll also have a look at how things are closed.

myleshorton: It looks to me like the connections are just closed without waiting for any existing data to do its thing:

https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/connpool/connpool.go#L185

myleshorton: Makes sense on the purpose of the WaitGroup @aranhoide

aranhoide: > It looks to me like the connections are just closed without waiting for any existing data to do its thing:
>
> https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/connpool/connpool.go#L185

Yeah, that's what I thought too, but that line is closing a _queued_ connection, that is, one that hasn't yet been fed to any user of the pool.

aranhoide:
OK, so I think the Client was basically doing the right things, wrt cleanup of dialers in the valencia branch.  Returning the hqfd on client.Configure was ugly, but not a serious problem. 

So I think a smaller change is needed after all for #2520.  It would mainly be adding a Configure method to the `config` package that takes a http.Client and can be called after Init.  The first time it is called it starts the config custom polling.  On subsequent calls it just updates the package-level sync.Atomic var holding the http.Client, just like the other modules have been doing.

Another change that we need to keep is removing the support for country-specific cloud.yamls.

There were have been a few improvements in PR #2522 and in this one that, while not strictly necessary, are worth keeping:

- NewDirectDomainFronter is a better name than NewDirectHttpClient.

- Calling d.dialServer() instead of having its body copied in the dialer of DDF clients.

- Wrapping the transport of DDF clients to avoid the http vs https gotcha.

- Several fixes to lantern-android.

There are some things that came up in my review that would be probably worth fixing now or later:

- balancer keeps its dialers sorted by QOS, so we probably don't need to explicitly find and store the highest QOS dialer separately.  A Client.GetHighestQosDdfClient() could fetch it immediately.

- Instead of closing the Client.hqfd on Client.Stop, we should close its current balancer (which releases all Dialers including that one).

- Analytics should probably not use direct fronting.

- I'm still not sure statserver needs DDF, but it can't hurt either.  Anyway, we need to make the var that holds the http.Client a sync.Atomic.

aranhoide:
I'll make a new PR with the changes described above.

myleshorton:
I pretty much agree with all of that @aranhoide. The one high level thing that comes to mind is that one refactoring we did was simply to use DDF to fetch new configs and to skip all the geo lookup stuff and geo lookup event listening, and I think we still have to do that one way or another. To me a lot of these refactorings were essentially to do that in a cleaner way without introducing what feel like sketchy dependencies, like config potentially knowing about client, for example.

Anyway, it does all feel a little over the top, though, so I agree a smaller change is in order!

myleshorton:
Oh, and I keep trying to mention that statserver needs to do DDF from geo lookups of peer IPs -- that's all it uses it for. So it needs it for the same reason the geo stuff needs it.

aranhoide:
@myleshorton I made PR #2538 to address those.  It doesn't introduce any dependencies.  The config package just gains a `Configure(*http.Client)` function.

aranhoide:
@myleshorton : you need DDF to get a geolookup of your own IP, because the client doesn't know its own IP to start with, and it's inconvenient to do a separate IP check.  But I can't see why we need DDF to do geolookup when you know the IP you want to look up.  If I ask geo.getiantem.org for the location of a IP of a peer of mine, does geo.getiantem.org care about my own IP?

aranhoide:
That said, it's nice to save the extra hop anyway, so I have no objections to DDF for geolookup of peers.

myleshorton:
@aranhoide No that's right @aranhoide -- it's not strictly needed, but if we can do it using DDF we might as well.

PR #2532, opened 2015-05-16
fix lantern-android/libflashlight/bindings tests, add to testpackages…

aranhoide:
….txt


aranhoide:
Merging since these are pretty obvious changes and having this broken was messing with go-oracle for me.

PR #2538, opened 2015-05-18
Fetch config via direct domain fronting, fixes #2520

### LINKED ISSUE #2520: Use direct fronting to load dynamic config

aranhoide:
@myleshorton : I think the only really necessary thing that the withclient scheme did was to clean up the extra fronted.Dialers that we were creating for direct domain fronting, as a result of extracting 'HighestQOSFrontedDialer'.  If we leave that logic at the balancer, the balancer takes care of cleaning those up alright.  

This PR makes a smaller change, while cherry picking all the useful things that came up while working on #2520 and #2529, as per: 

https://github.com/getlantern/lantern/pull/2529#issuecomment-102879786

I also made the commits tidier so they make it easier to understand the changes.  I think it will be faster to review this if you read the individual commits.

### DIFF
--- a/src/github.com/getlantern/lantern-android/client/client.go
+++ b/src/github.com/getlantern/lantern-android/client/client.go
@@ -59,7 +59,7 @@ func NewClient(addr, appName string) *MobileClient {
 
 	hqfd := client.Configure(clientConfig.Client)
 
-	hqfdc := hqfd.DirectHttpClient()
+	hqfdc := hqfd.NewDirectDomainFronter()
 
 	// store GA session event
 	sessionPayload := &analytics.Payload{
--- a/src/github.com/getlantern/lantern-android/client/client.go
+++ b/src/github.com/getlantern/lantern-android/client/client.go
@@ -10,7 +10,6 @@ import (
 	"github.com/getlantern/analytics"
 	"github.com/getlantern/flashlight/client"
 	"github.com/getlantern/flashlight/globals"
-	"github.com/getlantern/flashlight/util"
 )
 
 const (
@@ -28,7 +27,8 @@ var (
 // MobileClient is an extension of flashlight client with a few custom declarations for mobile
 type MobileClient struct {
 	client.Client
-	closed chan bool
+	closed  chan bool
+	fronter *http.Client
 }
 
 // init attempts to setup client configuration.
@@ -75,8 +75,9 @@ func NewClient(addr, appName string) *MobileClient {
 	analytics.SessionEvent(hqfdc, sessionPayload)
 
 	return &MobileClient{
-		Client: client,
-		closed: make(chan bool),
+		Client:  client,
+		closed:  make(chan bool),
+		fronter: hqfdc,
 	}
 }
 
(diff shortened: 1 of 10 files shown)


myleshorton:
Awesome @aranhoide! I'll look at this first thing tomorrow but overall agree with the approach and thinking.

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -110,18 +104,11 @@ func (client *MobileClient) ServeHTTP() {
 // updateConfig attempts to pull a configuration file from the network using
 // the client proxy itself.
 func (client *MobileClient) updateConfig() error {
-	var err error
 	var buf []byte
-	var cli *http.Client
-
-	if cli, err = util.HTTPClient(cloudConfigCA, client.Addr); err != nil {
-		return err
-	}
```
myleshorton: Oh so it's not quite as bad as I thought -- I thought it was not proxying to the config at all, but it's just not doing DDF.

aranhoide: Not as bad, but close.  Note that

- The default configuration [didn't include](https://github.com/getlantern/lantern/pull/2538/files#diff-8be20108b452119099e317b2a3b4e03aR116) the trusted CAs.  This was probably a bug and not by design.

- The client was initially [trying](https://github.com/getlantern/lantern/pull/2538/files#diff-8be20108b452119099e317b2a3b4e03aL131) to fetch a new config without proxying.  Only if this succeeds will the trusted CAs be properly initialized.

BTW, I just realised that the trusted CAs were never updated after initial setup.  I just submitted a fix.

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -59,7 +52,7 @@ func NewClient(addr, appName string) *MobileClient {
 
 	hqfd := client.Configure(clientConfig.Client)
 
-	hqfdc := hqfd.DirectHttpClient()
+	hqfdc := hqfd.NewDirectDomainFronter()
```
myleshorton: I think we actually want this to be a util.HttpClient. This using DDF appears to just be a bug in both lantern-android and lantern.

src/github.com/getlantern/lantern-android/client/client.go
myleshorton: Nice! That timeout of 5 seconds was pretty darn aggressive too. Good catch.

src/github.com/getlantern/lantern-android/client/client.go
aranhoide: I agree.  I didn't change it (yet) because I hadn't made sure that was the case and that is orthogonal to using DDF for fetching the config.  Do you want me to change this for this PR, or shall we just open an issue to do a different one for that?

myleshorton:
Yeah you're right @aranhoide. I'll just merge this and create a different issue. Dude, great work on this all around - caught a couple of really key bugs. The Init and Configure combo in config.go is a little ugly, but certainly simpler than the alternatives this close to pushing this into production and could be worse.

myleshorton:
Just got confirmation from @atavism it looks like a bug on the analytics using DDF.

src/github.com/getlantern/lantern-android/client/client.go
myleshorton: I just created https://github.com/getlantern/lantern/issues/2540. I'm happy to just knock that out unless you're itching to get to it.

aranhoide: Go for it.  I'm actually itching to go back to the config server. :)

myleshorton: Cool -- on it.

aranhoide: After that is fixed, we can probably remove `fronter` as a field in `MobileClient`, but that's not a pressing issue of course.

PR #2541, opened 2015-05-18
Don't use direct domain fronting for analytics closes #2540

### LINKED ISSUE #2540: Don't use direct domain fronting for analytics

myleshorton:
OK this changes the analytics calls to not use DDF.

### DIFF
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -221,7 +221,7 @@ func runClientProxy(cfg *config.Config) {
 	if er != nil {
 		log.Errorf("Could not create HTTP client %v", er)
 	} else {
-		analytics.Configure(cfg, false, httpClient)
+		analytics.Configure(httpClient, cfg, cfg.Addr, version)
 	}
 
 	if hqfd == nil {
@@ -304,7 +304,6 @@ func runServerProxy(cfg *config.Config) {
 	}
 
 	srv.Configure(cfg.Server)
-	analytics.Configure(cfg, true, nil)
 
 	// Continually poll for config updates and update server accordingly
 	go func() {
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -217,7 +217,7 @@ func runClientProxy(cfg *config.Config) {
 	settings.Configure(cfg, version, buildDate)
 	proxiedsites.Configure(cfg.ProxiedSites)
 
-	httpClient, er := util.HTTPClient(cfg.CloudConfigCA, cfg.Addr)
+	httpClient, er := util.HTTPClient("", cfg.Addr)
 	if er != nil {
 		log.Errorf("Could not create HTTP client %v", er)
 	} else {
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -217,6 +217,13 @@ func runClientProxy(cfg *config.Config) {
 	settings.Configure(cfg, version, buildDate)
 	proxiedsites.Configure(cfg.ProxiedSites)
 
+	httpClient, er := util.HTTPClient(cfg.CloudConfigCA, cfg.Addr)
+	if er != nil {
+		log.Errorf("Could not create HTTP client %v", er)
+	} else {
+		analytics.Configure(cfg, false, httpClient)
+	}
+
 	if hqfd == nil {
 		log.Errorf("No fronted dialer available, not enabling geolocation, stats or analytics")
 	} else {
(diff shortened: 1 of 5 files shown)


myleshorton:
You mind taking a peek @aranhoide?

myleshorton:
Whoops -- never mind @aranhoide. Please hold!

myleshorton:
OK this should be all set now. @todd or @aranhoide you want to take a look?

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -217,6 +217,13 @@ func runClientProxy(cfg *config.Config) {
 	settings.Configure(cfg, version, buildDate)
 	proxiedsites.Configure(cfg.ProxiedSites)
 
+	httpClient, er := util.HTTPClient(cfg.CloudConfigCA, cfg.Addr)
```
aranhoide: Is the CloudConfigCA guaranteed to be good for analytics?  I guess we shouldn't rely on Google using the same CA as S3.

BTW, this reminds me we'll probably need to change the CloudConfigCA to Cloudflare's when we swap in the config server.  Or we could get a cert for the same CA?

myleshorton: Yeah this is interesting. I *think* CloudConfigCA  will actually be blank at this point typically, which I *think* means it'll use the Go defaults. We definitely have a related issue on the FT side as well.

I guess we should actually pull out the Google Analytics CA here though huh?

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -52,32 +54,11 @@ func NewClient(addr, appName string) *MobileClient {
...
-		}
-	}
-
-	analytics.SessionEvent(hqfdc, sessionPayload)
-
 	return &MobileClient{
 		Client:  client,
 		closed:  make(chan bool),
-		fronter: hqfdc,
+		fronter: hqfd.NewDirectDomainFronter(),
```
aranhoide: We could maintain this locally in `pollConfiguration`, since it's only ever used by its callees.  But it's not a big deal to have it here either.

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -101,6 +83,39 @@ func (client *MobileClient) ServeHTTP() {
...
+	if client.appName != "" {
+		if trackingId, ok := trackingCodes[client.appName]; ok {
+			sessionPayload.TrackingId = trackingId
+		}
+	}
+
+	// Report analytics, proxying through the local client. Note this
+	// is a little unorthodox by Lantern standards because it doesn't
+	// pin the certificate of the cloud.yaml root CA, instead relying
+	// on the go defaults.
```
aranhoide: If an empty CA works, we better use that in the desktop client too.

aranhoide:
Done with my review.  Summary: we're using the cloud config CA for analytics.  I don't think that'll work, and even if it did, we better not rely on it.

src/github.com/getlantern/flashlight/flashlight.go
aranhoide: It depends.  Do you think there's a realistic chance chance that Go defaults will miss the CA of a Google-run service? :)

xiam:
Hey guys, this is the one we need to review before rebuilding the .so for Firetweet, right?

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -217,6 +217,13 @@ func runClientProxy(cfg *config.Config) {
...
+	if er != nil {
+		log.Errorf("Could not create HTTP client %v", er)
+	} else {
+		analytics.Configure(cfg, false, httpClient)
```
atavism: I think we're calling this too soon. I'm getting ``May 19 00:55:10.898 - analytics: analytics.go:134 https://www.google-analytics.com/collect: http: error connecting to proxy http://localhost:10080: dial tcp 127.0.0.1:10080: connection refused`` -- We need to use waitforserver prior to the analytics call to make sure the client proxy is ready.

xiam:
Looks good and our tests run fine too, but an odd message shows up:

```
enproxy: conn_intf.go:239 Failing on Dest: www.google.com:80    ProxyHost:     Unable to issue write request: Bad response status for read: 302 Moved Temporarily
enproxy: conn_intf.go:239 Failing on Bad response status for read: 302 Moved Temporarily
```

I'll check conn_intf to make sure this is not something we should worry about.

src/github.com/getlantern/flashlight/flashlight.go
aranhoide: Makes sense.  Good catch!

xiam:
Yes, double checked and there's nothing to worry about that message, the problem is with the proxy that is being tested, so maybe the only one that remains is the comment from @atavism.

src/github.com/getlantern/lantern-android/client/client.go
myleshorton: Good point. Will do.

myleshorton:
OK guys merging this bad boy in...

PR #2544, opened 2015-05-19
Adding waitforserver when creating HTTP proxy client.

xiam:
Hello @atavism, @myleshorton, this is a suggested fix for https://github.com/getlantern/lantern/pull/2541, I think it would be better to handle the waitforserver call inside the proxy creation routine, could you give it a look?

### DIFF
--- a/src/github.com/getlantern/flashlight/util/http.go
+++ b/src/github.com/getlantern/flashlight/util/http.go
@@ -37,18 +37,8 @@ func HTTPClient(rootCA string, proxyAddr string) (*http.Client, error) {
 // proxyAddr is specified, the client will proxy through the given http proxy.
 func httpClient(rootCA string, proxyAddr string, persistent bool) (*http.Client, error) {
 
-	log.Debugf("Waiting for proxy server...")
-
-	// Waiting for proxy server to came online.
-	err := waitforserver.WaitForServer("tcp", proxyAddr, 60*time.Second)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -35,6 +36,18 @@ func HTTPClient(rootCA string, proxyAddr string) (*http.Client, error) {
...
+
+	// Waiting for proxy server to came online.
+	err := waitforserver.WaitForServer("tcp", proxyAddr, 60*time.Second)
+	if err != nil {
+		// Instead of finishing here we just log the error and continue, the client
+		// we are going to create will surely fail when used and return errors,
+		// those errors should be handled by the code that depends on such client.
+		log.Errorf("Proxy never came online at %v: %q", proxyAddr, err)
+	}
+
```
myleshorton: Yeah I like the general thinking here @xiam . Two issue I see though:

1. This method is at least designed to not require a proxy, so I would move this after line 78 ```if proxyAddr != "" {```
1. If the proxy is not a localhost address, this might not be what we want (it would probably make more sense to just let it fail). We could use the IP IsLoopback method to check that before making this call -- http://golang.org/pkg/net/#IP.IsLoopback

xiam: True. Great suggestion! Care to take a look again?

myleshorton:
LGTM and working well in testing. Thanks @xiam!

Issue #2482, opened 2015-05-06
Lantern does DNS resolutions in order to reach loggly and google analytics


xiam:
Captured some DNS resolutions while Lantern was idle:

![screen shot 2015-05-06 at 1 16 14 pm](https://cloud.githubusercontent.com/assets/385670/7500239/1c49977c-f3f2-11e4-8f4a-0978f26adfd4.png)

myleshorton:
We shouldn't actually be resolving the DNS for analytics or loggly at all since both of them are tunneled (along with everything else theoretically) so we the DNS shouldn't be resolved until the server side is reached.

fffw:
Loggly is not in the proxied site list, same for pure-journey-3547.herokuapp.com. We should add them.

google-analytics.com is in the list. Theoretically it should not be resolved directly. Are there any possibility that they came from other applications?

myleshorton:
Oh that makes sense @fffw. We used to not need them on the list since we were just configuring things internally to go through a proxy, but with detour we do. Analytics is also at least attempting do do direct domain fronting, which seems incorrect to me and could be related to a DNS lookup leak. I'll keep looking at this.

xiam:
Analytics is no longer showing up but loggly remains.

![screen shot 2015-06-02 at 4 54 42 pm](https://cloud.githubusercontent.com/assets/385670/7948252/246501c4-0948-11e5-8b2b-c2f59fd5b426.png)

I'm capturing traffic using a VM with just few services enabled so there is no chance that other app besides Lantern is trying to reach loggly.

fffw:
Yeah I can find accessing to logs-01.loggly.com directly from my logs, I think that's because my existing lantern.yaml doesn't contain loggly.com in proxied sites. Once we run `genconfig` next time, the changes to `proxiedsites/original.txt` will be propagated I think.

xiam:
Let me try again with a clean lantern.yaml

xiam:
Tested again but no luck guys:

![screen shot 2015-06-03 at 6 41 03 pm](https://cloud.githubusercontent.com/assets/385670/7973863/22ce1e26-0a20-11e5-8115-15832a243918.png)

And in fact google analytics came again after some time.

This is the current valencia branch I only used Lantern and cURL to carry on the test, so there is no risk other program is doing these requests. 

I already removed the `.lantern` directory and I'm going to continue digging into this one, any other thoughts?

xiam:
[...]

PR #2545, opened 2015-05-19
adding heroku and loggly to proxied sites list #2482

myleshorton:
This doesn't fully close out the issue but at least adds these to the list.


myleshorton:
I'm just merging this one in guys since there's basically no review necessary...

PR #2551, opened 2015-05-19
Added log for config update errors closes #2550

### LINKED ISSUE #2550: FT doesn't log errors if it can't fetch new configs.

myleshorton:


myleshorton:
Trivial change. Merging.

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -120,6 +120,7 @@ func (client *MobileClient) updateConfig() error {
 	var buf []byte
 	var err error
 	if buf, err = pullConfigFile(client.fronter); err != nil {
+		log.Fatalf("Could not update config: '%v'", err)
```
fffw: @myleshorton Just to make sure, `log.Fatalf` will quit the program, is it intended? `MobileClient.pollConfiguration()` will call this method over and over again.

myleshorton: Good catch. No not intended. I think we should actually be using ```getlantern/golog``` here and Errorf. We should do a new build @atavism.

Any reason we're not using getlantern/golog in lantern-android?

myleshorton: Just created https://github.com/getlantern/firetweet/issues/86 and https://github.com/getlantern/lantern/pull/2554

Can you review https://github.com/getlantern/lantern/pull/2554 @fffw?

fffw:
Sure @myleshorton. I'm not sure if @atavism has any special consideration to not use `getlantern/golog`, but it should be easy to revert it back if it's the case.

Issue #2552, opened 2015-05-19
Error building android-lib


myleshorton:
On OSX...

$ make android-lib
github.com/getlantern/tarfs
../../tarfs/tarfs.go:255: AssetFile.Size is ambiguous
../../tarfs/tarfs.go:255: cannot use f (type *AssetFile) as type os.FileInfo in return argument:
	*AssetFile does not implement os.FileInfo (missing Size method)
../../tarfs/tarfs.go:279: AssetDirectory.Size is ambiguous
../../tarfs/tarfs.go:279: cannot use f (type *AssetDirectory) as type os.FileInfo in return argument:
	*AssetDirectory does not im

myleshorton:
@fffw do you have any idea what this might be? Looks like something to do with genassets again?

myleshorton:
Or sorry I meant the above comment for @xiam!

xiam:
I could not reproduce the error so I checked out a clean repo and tried again, now I am having a completely different error, and just noticed the https://github.com/golang/mobile now requires Go 1.5:

```
# golang.org/x/mobile/app/internal/callfn
/src/golang.org/x/mobile/app/internal/callfn/callfn_arm.s:8 syntax error, last name: 0
make: *** [android-lib] Error 2
```

The temporal fix is this:

```
cd src/golang.org/x/mobile
git checkout cd5954b
```

We'll probably need to use gost here too.

After that I tried building again and it worked, so I am not sure what's happening, anyone else experiencing the same error?

fffw:
I encountered following error. Trying to touch docker to find out the reason.

```
finalize namespace chdir to /src/golang.org/x/mobile no such file or directoryFATA[0001] Error response from daemon: Cannot start container bbe437c89253c56ea9115ccea95343a017766527782d883b4f20e92b8eab36c9: finalize namespace chdir to /src/golang.org/x/mobile no such file or directory
```

xiam:
Probably the same thing, could you try this?

```
go get golang.org/x/mobile/cmd/gobind
```

Then try again.

fffw:
Oh I found it. I didn't `source setenv.sh` before `make android-lib`, so it just installed to another place. Should add it to `docker-golang-android` target.

fffw:
Although there are still lots of `javac` errors, the specific error in this issue is fixed by https://github.com/getlantern/lantern/pull/2558. @myleshorton Mind take a look?

[...]

PR #2558, opened 2015-05-21
resolve ambiguity in tarfs

fffw:
Fixes https://github.com/getlantern/lantern/issues/2552

### DIFF
--- a/src/github.com/getlantern/tarfs/tarfs.go
+++ b/src/github.com/getlantern/tarfs/tarfs.go
@@ -252,7 +252,7 @@ func (f *AssetFile) Readdir(count int) ([]os.FileInfo, error) {
 }
 
 func (f *AssetFile) Stat() (os.FileInfo, error) {
-	return f, nil
+	return &f.FakeFile, nil
 }
 
 // AssetDirectory implements http.File interface for a directory. It is always
@@ -273,8 +273,4 @@ func newAssetDirectory(name string) *AssetDirectory {
 
 func (f *AssetDirectory) Readdir(count int) ([]os.FileInfo, error) {
 	return emptyFileInfos, nil
-}
-
-func (f *AssetDirectory) Stat() (os.FileInfo, error) {
-	return f, nil
 }


myleshorton:
Nice! That fixes it for me, although I have to admit I don't understand the change. Can you explain @fffw?

src/github.com/getlantern/tarfs/tarfs.go
```diff
@@ -252,7 +252,7 @@ func (f *AssetFile) Readdir(count int) ([]os.FileInfo, error) {
 }
 
 func (f *AssetFile) Stat() (os.FileInfo, error) {
-	return f, nil
+	return &f.FakeFile, nil
```
fffw: `AssetFile` embeds 3 members. Seems another member also has `Size()` method, which causes ambiguity, although I don't know exactly which one. Returning `&f.FakeFile` explicitly to eliminate the ambiguity, as `FakeFile` is existed to provided an `os.FileInfo` interface.

myleshorton: Ah I think I see -- conflict between AssetFile and AssetDirectory correct?

src/github.com/getlantern/tarfs/tarfs.go
```diff
@@ -274,7 +274,3 @@ func newAssetDirectory(name string) *AssetDirectory {
 func (f *AssetDirectory) Readdir(count int) ([]os.FileInfo, error) {
 	return emptyFileInfos, nil
 }
-
```
fffw: Embed `AssetFile` already have `Stat()` method. See https://golang.org/doc/effective_go.html#embedding

myleshorton:
LGTM.

src/github.com/getlantern/tarfs/tarfs.go
fffw: I don't think so, though actually I don't know what causes the ambiguity. Removing `AssetDirectory` altogether doesn't stop this error.

PR #2565, opened 2015-05-22
Skip detour on Android

atavism:


myleshorton:
I think we need to change reverse proxy not to use detour too:

https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/client/reverseproxy.go#L47

myleshorton:
Lgtm

atavism:
@myleshorton Thanks!

PR #2574, opened 2015-05-26
more informative logging

fffw:
Didn't fix anything, but would be helpful to spot dynamic behaviour of Lantern client.


src/github.com/getlantern/balancer/balancer.go
```diff
@@ -86,9 +86,7 @@ func (b *Balancer) DialQOS(network, addr string, targetQOS int) (net.Conn, error
 		if d == nil {
 			return nil, fmt.Errorf("No dialers left")
 		}
-		if d.Label != "" {
-			log.Debugf("Dialing %s://%s with %s", network, addr, d.Label)
-		}
+		log.Debugf("Dialing %s://%s with %s", network, addr, d.Label)
```
fffw: Log is still useful even if there's no label

PR #2576, opened 2015-05-26
Changed to keep reading from UI in the event of a single failure #2480

myleshorton:
Some logging additions as well. Will flag the appropriate couple of lines.

### DIFF
--- a/src/github.com/getlantern/flashlight/ui/service.go
+++ b/src/github.com/getlantern/flashlight/ui/service.go
@@ -31,8 +31,10 @@ func (s *Service) write() {
 	for {
 		select {
 		case <-s.stopCh:
+			log.Trace("Received message on stop channel")
 			return
 		case msg := <-s.out:
+			log.Tracef("Creating new envelope for %v", s.Type)
 			b, err := newEnvelope(s.Type, msg)
 			if err != nil {
 				log.Error(err)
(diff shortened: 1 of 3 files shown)


src/github.com/getlantern/flashlight/ui/service.go
```diff
@@ -140,15 +147,16 @@ func read() {
 		// Delegating response to the service that registered with the given type.
 		if services[envType.Type] == nil {
 			log.Errorf("Message type %v belongs to an unknown service.", envType.Type)
-			return
+			continue
```
myleshorton: This and the following are the only actual changes.

src/github.com/getlantern/flashlight/ui/service.go
```diff
@@ -140,15 +147,16 @@ func read() {
...
 		}
 
 		env := &Envelope{}
 		err = json.Unmarshal(b, env)
 		if err != nil {
 			log.Errorf("Unable to unmarshal message of type %v: %v", envType.Type, err)
-			return
+			continue
```
myleshorton: This too.

myleshorton:
This at least makes things more robust if something odd happens. I suspect there's something else going on as well, but this should at least improve things.

myleshorton:
This is a quick one @atavism if you want to have a peek. Hopefully it doesn't conflict with your other PR.

atavism:
No, this shouldn't affect my changes. If we get an unknown message type (or can't decode a message), we keep the channel open anyway, assuming the service will register at a later point. Makes sense to me. Merging.

myleshorton:
Cool thanks @atavism!

PR #2582, opened 2015-05-27
Some lantern-android and analytics updates

atavism:


src/github.com/getlantern/analytics/analytics.go
```diff
@@ -4,22 +4,22 @@ import (
...
+	ProtocolVersion   = "1"
+	DefaultInstanceId = "1260961011.1389432370"
 )
 
 var (
-	log = golog.LoggerFor("analytics")
-
-	defaultHttpClient = &http.Client{}
+	log        = golog.LoggerFor("analytics")
+	httpClient = &http.Client{}
```
myleshorton: I would just declare the httpClient and not create one here. I realize that's not really a change you made, but without that the later check for httpClient == nil doesn't make any sense since it will never be nil, and initially we'll try to make direct analytics call until util.HttpClient returns.

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -55,12 +53,14 @@ func NewClient(addr, appName string) *MobileClient {
 
 	hqfd := client.Configure(clientConfig.Client)
```
myleshorton: This isn't related to your change @atavism, but it looks like init() in this file is actually never called. I'm unclear on how this is working at all, since that would mean that clientConfig is never created and should be nil.

myleshorton: It seems like clientConfig should just be initialized to defaultConfig() as soon as it's declared.

atavism: The ``init()`` function is always implicitly called in Go, right? It's for general package initialization.

atavism: I agree though we could just initialize it when it's declared. I went ahead and made that change.

myleshorton: Cool -- yeah I actually didn't realize go had that but it seems like it's mainly useful when you can't actually initialize something when it's declared -- somewhat akin to a static initializer block in java.

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -85,50 +80,33 @@ func (client *MobileClient) ServeHTTP() {
...
+	return err
 }
 
 // pollConfiguration periodically checks for updates in the cloud configuration
 // file.
 func (client *MobileClient) pollConfiguration() {
 
-	// initially poll the config immediately
-	pollTimer := time.NewTimer(time.Second)
+	pollTimer := time.NewTimer(cloudConfigPollInterval)
```
myleshorton: Wait so you don't want to grab the network config right away anymore @atavism? What's the thinking there?

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -55,12 +48,14 @@ func NewClient(addr, appName string) *MobileClient {
...
 	hqfd := client.Configure(clientConfig.Client)
 
-	return &MobileClient{
+	mClient := &MobileClient{
 		Client:  client,
 		closed:  make(chan bool),
 		fronter: hqfd.NewDirectDomainFronter(),
 		appName: appName,
 	}
+	go mClient.updateConfig()
```
atavism: @myleshorton We talked about moving the code to pull the configuration into its routine. I went ahead and did that and now I'm fetching it right away when we declare the client.

src/github.com/getlantern/analytics/analytics.go
```diff
@@ -68,8 +68,21 @@ type Payload struct {
...
+func Configure(trackingId string, version string, proxyAddr string) {
+	var err error
+	go func() {
+		httpClient, err = util.HTTPClient("", proxyAddr)
+		if err != nil {
+			log.Errorf("Could not create HTTP client via %s: %s", proxyAddr, err)
+			return
+		}
+		// Store new session info whenever client proxy is ready
+		sessionEvent(version, trackingId)
```
myleshorton: It looks like this will now record a session event every time regardless of whether or not the user has chosen to auto-report stats @atavism.

atavism: We shouldn't be doing any analytics setup if auto-report is disabled. I moved the ``Configure()`` inside the autoReport check in ``flashlight/analytics.go``. This wouldn't matter to FT where we have no such setting.

src/github.com/getlantern/lantern-android/client/client.go
myleshorton: Cool -- looks good @atavism

myleshorton:
Nice OK LGTM @atavism!

Issue #2578, opened 2015-05-26
Docker image builder is not working anymore


xiam:
This happens after upgrading docker, removing all my images and trying to build Lantern again:

```
INFO[0078] The command [/bin/sh -c cd $GOROOT/src && CGO_ENABLED=1 ./all.bash] returned a non-zero code: 1
make: *** [docker] Error 1
```

xiam:
This is the failing test to blame:

```
--- FAIL: TestFilePacketConn (0.01s)
        file_test.go:121: WriteTo failed: write ip: bad address
```

I suspect it has something to do with the docker and boot2docker upgrades.

xiam:
The test that we see in the log message attempts to write to `127.0.0.1` using the ICMP protocol, but it fails, this test assumes that if you have the UID 0 then you'll be able to craft ICMP packets, this does not happen in this particular docker machine (`Docker version 1.6.2, build 7c8fca2`) + boot2docker (`Boot2Docker-cli version: v1.6.2`), this is probably a security restriction.

This may be related to:
https://bugzilla.redhat.com/show_bug.cgi?id=1142311
http://www.cyberciti.biz/faq/unable-to-use-ping-in-centos-linux-rhel-7-docker-container/

And *might* be fixed by https://github.com/docker/docker/issues/1916

I'm working in a workaround now and I'm going to open an issue for the docker guys later.

yejingchen:
But I am unable to build lantern on fedora 23 because of exactly the same failure, and I didn't see any output indicating it is skipping the test.

xiam:
@yejingchen what's the error message for you?

yejingchen:
I don't remember there was an error message (otherwise I will notice them), the only thing I noticed is the output of `curl`, all the statistics is 0.

PR #2583, opened 2015-05-27
Patch for skipping a failing test than rendered our docker image usel…

xiam:
This should fix https://github.com/getlantern/lantern/issues/2578 and should allow us to continue using the docker image for building Lantern with go1.4.


atavism:
Working fine for me now.

yejingchen:
I am on commit 1c984c8 (which is after the commit skipping the failed test) but when the command fetching the patch failed, the building process continues as if it succeeded, which made the patch useless.

fffw:
Could you create an issue with the command outputs and description @yejingchen so we can track it independently?

Issue #2568, opened 2015-05-24
Lantern thinks cloud config is always a new version


myleshorton:
Not sure if it's because the new config isn't honoring the etag, but Lantern thinks the config is updated every time now.

aranhoide:
There are two problems here, one of which I don't fully understand.

First, the client is never sending any headers (or, incidentally, body, but I don't suppose we're using this for anything) in direct requests.  These are left behind [here](https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/fronted/dialer.go#L211).  Only the request URL is copied over.

But even once we fix that there's a more befuddling one.  The client never sees any ETag in the response headers of the config request.  If I try direct domain fronting with curl to a server I set up:

    curl -i -k -H "If-None-Match: my-etag" -H "Host: config-test.getiantem.org" https://198.41.16/cloud.yaml.gz

I get this in the response

    Etag: thisisanetag
    Someotherheader: thisisanotherheader

(these headers are set unconditionally).  But if I modify [this line](https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/config/config.go#L225) in the Lantern client to

```go
		cfg.CloudConfig = "https://config-test.getiantem.org/cloud.yaml.gz"
```

...and I print out the headers in the response when the client is trying to update its config, then I get no Etag (but I do get Someotherheader).

I couldn't find anywhere where our implementation of DDF would mess with the response, or anything in our request that could cause Cloudflare to strip the Etag itself.

A workaround for this would be to just use a custom header instead of ETag for this.  I think this is a sane thing to do anyway if we're serving this dynamically, because if we ever need to switch to other CDNs they are bound to do funny things with standard headers like ETag too.  

We'll need to release a fix for the (first) error described above anyway, so when the config server is ready should I just change the etag constant in config/config.go to something like X-Lantern-Etag and be done with it?

aranhoide:
While the 'If-None-Match' header makes it to the server alright, maybe I might as well change that one for a custom one too.

myleshorton:
Yeah the custom headers sound good to me - great digging over there!

myleshorton:
@aranhoide you could also always use the Last-Modified and If-Modified-Since headers if those make it through.

myleshorton:
Is this all set with your latest changes @aranhoide?

aranhoide:
[...]

PR #2592, opened 2015-05-29
fix client side of #2568

aranhoide:
- Copy over the headers of direct fronted requests.

- Change to proprietary headers for caching, so CDNs and middleware
  will hopefully keep their dirty hands off them.

These won't take effect until we deploy the config server, but in the meanwhile they won't make things any more broken either.

### DIFF
--- a/src/github.com/getlantern/fronted/dialer.go
+++ b/src/github.com/getlantern/fronted/dialer.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/url"
 	"strings"
 	"sync"
 	"time"
@@ -212,6 +213,8 @@ func (ddf *DirectDomainTransport) RoundTrip(req *http.Request) (resp *http.Respo
 	// create a copy.
 	norm := new(http.Request)
 	*norm = *req // includes shallow copies of maps, but okay
+	norm.URL = new(url.URL)
+	*norm.URL = *req.URL
 	norm.URL.Scheme = "http"
 	return ddf.Transport.RoundTrip(norm)
 }
--- a/src/github.com/getlantern/fronted/dialer.go
+++ b/src/github.com/getlantern/fronted/dialer.go
@@ -16,7 +16,6 @@ import (
 
 	"encoding/asn1"
 	"github.com/getlantern/connpool"
-	"github.com/getlantern/deepcopy"
 	"github.com/getlantern/enproxy"
 	"github.com/getlantern/golog"
 	"github.com/getlantern/proxy"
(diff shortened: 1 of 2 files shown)


src/github.com/getlantern/fronted/dialer.go
```diff
@@ -212,6 +213,9 @@ func (ddf *DirectDomainTransport) RoundTrip(req *http.Request) (resp *http.Respo
 	if err != nil {
 		return nil, fmt.Errorf("Unable to construct request for url '%s' with error '%v'", normalized, err)
 	}
+	if err = deepcopy.Copy(&norm.Header, req.Header); err != nil {
+		return nil, fmt.Errorf("Unable to copy request headers: %v", err)
+	}
 	return ddf.Transport.RoundTrip(norm)
```
myleshorton: I just realized there's a better way to do this whole request copying business -- I'll submit another PR.

myleshorton:
@aranhoide can you take a peek at my changes? I wonder if this code was the reason the etag stuff was breaking in the first place?

aranhoide:
@myleshorton

- By overwriting the url scheme of `norm` you are overwriting that of `req`, since URL is a pointer field.  I just sent a proposed change (give `norm` a new URL, copy the old one over before modifying it).

- This code could have caused this error by itself, because it was preventing the client from sending the server the `If-None-Match` header.  But even after fixing this, the error would persist, because something (either in our DDF code or in Cloudflare) was eating the `ETag` header before it could reach the client.

I have confirmed that with this branch as it stands, and pointing the client to the config server, this bug is fixed.

myleshorton:
Hey actually @aranhoide I don't think the extra memory fiddling in your patch is necessary since the *struct = *struct syntax does actually do a shallow *copy* (doesn't hit data structures). See:

https://play.golang.org/p/6rvtpgU9tX

aranhoide:
I think in that gist you are taking the address of the pointers themselves, not of the structures they point to.  If you remove the ampersands, you see they point to the same address.  Also, you can test directly that messing with the URL.Scheme of the copy modifies the original:

https://play.golang.org/p/34oqZWcD9H

myleshorton:
Ah yes sorry -- didn't realize the URL field in Request was actually a pointer:

```
    77	type Request struct {
    78		// Method specifies the HTTP method (GET, POST, PUT, etc.).
    79		// For client requests an empty string means GET.
    80		Method string
    81	
    82		// URL specifies either the URI being requested (for server
    83		// requests) or the URL to access (for client requests).
    84		//
    85		// For server requests the URL is parsed from the URI
    86		// supplied on the Request-Line as stored in RequestURI.  For
    87		// most requests, fields other than Path and RawQuery will be
    88		// empty. (See RFC 2616, Section 5.1.2)
    89		//
    90		// For client requests, the URL's Host specifies the server to
    91		// connect to, while the Request's Host field optionally
    92		// specifies the Host header value to send in the HTTP
    93		// request.
    94		URL *url.URL
```

Issue #2314, opened 2015-03-19
Include a GUID in Loggly messages


myleshorton:
That way we'll be able to tell if a bunch of messages are from the same GUID.

oxtoacart:
We don't have users.  Do you mean the instance id?

myleshorton:
Let's just call it a GUID!

oxtoacart:
Cool.  We already have an instanceid that we use for flashlight servers - https://github.com/getlantern/flashlight-build/blob/valencia/src/github.com/getlantern/flashlight/globals/globals.go#L13

We just need to auto-generate this and save it in the config file, and then we can add it to our Loggly data.

myleshorton:
Oh nice!

fffw:
Closed by https://github.com/getlantern/lantern/pull/2594

PR #2594, opened 2015-05-29
generate instanceid if it's empty

fffw:


src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -225,6 +228,10 @@ func (cfg *Config) ApplyDefaults() {
 		cfg.CloudConfig = "https://config.getiantem.org/cloud.yaml.gz"
 	}
 
+	if cfg.InstanceId == "" {
+		cfg.InstanceId = hex.EncodeToString(uuid.NodeID())
```
fffw: NodeID doesn't change on one device, which is suitable for an instance id. It would be better we prefix it with text, but either the role or country will change.

fffw:
This resolves https://github.com/getlantern/lantern/issues/2314

myleshorton:
LGTM @fffw thanks!

PR #2597, opened 2015-05-29
add severity to logs and update testify closes #2589

### LINKED ISSUE #2589: Add severity level to logs
This is useful for knowing, for example, if you've encountered any errors during a test run of Lantern.

myleshorton:
OK, here's another stab at this @xiam. I actually did not use gost get for the new stretchr package, but I figure we should just move to gb sooner rather than later and bring in all versioned dependencies at that time.


src/github.com/getlantern/golog/golog_test.go
```diff
@@ -6,26 +6,34 @@ import (
 	"io/ioutil"
 	"os"
 	"regexp"
+	"strings"
 	"testing"
 	"time"
 
-	"github.com/getlantern/testify/assert"
```
xiam: Looks like we initially forked this one in order to add a small change https://github.com/stretchr/testify/compare/master...getlantern:master.

The change was just adding more debugging to the output of a test, so it's OK to use the upstream package. We should probable deprecate `github.com/getlantern/testify` too if we don't plan to make major changes.

xiam:
This one looks better, thanks @myleshorton!

src/github.com/getlantern/golog/golog_test.go
myleshorton: Yeah exactly @xiam. Ours was actually worse at displaying more standard test output, though, which is what we need for this particular test, so I reverted. I agree we should probably just remove getlantern/testify, as I'm also guessing similar functionality is now upstream.

myleshorton:
Thanks @xiam!

PR #2600, opened 2015-05-31
add script to update the redis config from a cloud.yaml

aranhoide:
This adds a semi-manual way to update the configuration in the redis database that [the config server](https://github.com/getlantern/config-server) uses.

Soon(TM), servers will register themselves when they come up.  At the moment, launching servers in Tokyo is still a semi-manual process, so that would need be fixed first.

As soon as we officially stop supporting client versions that rely on static `cloud.yaml`s we may do a more aggressive rewrite of this whole system, to simplify it.


aranhoide:
Merging since I'm already using this to update the configuration served by the config server.  Please someone review it nevertheless, but keeping this out of Valencia now would be just an exercise in hypocrisy. :)

PR #2610, opened 2015-06-05
close both clientConn and connOut to resolve #2330

### LINKED ISSUE #2330: Lots of too many open files errors

fffw:
It is too simple to believe, but we just closed wrong connections. `io.Copy` will not return if there's nothing to read from `connOut` unless we close it. Though we should dig more into why `idletiming` doesn't cause `connOut.Read()` to return.

To verify, run below command at any time to see if the ingress and egress connection count equals.

```
PID=`ps -ef | grep lantern | grep -v grep | cut -d " " -f 4` && lsof -nPp $PID | grep TCP | grep "8787->" | grep EST | wc && lsof -nPp $PID | grep TCP | grep ":443 (ESTABLISHED)" | wc
```

### DIFF
--- a/src/github.com/getlantern/detour/detour.go
+++ b/src/github.com/getlantern/detour/detour.go
@@ -124,6 +124,7 @@ func Dialer(d dialFunc) dialFunc {
 					return dc, nil
 				}
 				log.Debugf("Dial %s to %s, dns hijacked, try detour", dc.stateDesc(), addr)
+				dc.conn.Close()
 			} else if detector.TamperingSuspected(err) {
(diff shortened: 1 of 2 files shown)


myleshorton:
LGTM! Nice catch @fffw

aranhoide:
Is it important to deploy this fix to our servers, or is it meant for the client side only?

fffw:
@aranhoide It's client side only

Issue #2390, opened 2015-04-14
Refactor config setup


myleshorton:
As @fffw suggested, it would simplify things to have the cloud.yaml just always be the immutable canonical version of the config and to have any user-specified settings in a completely different file.

fffw:
There are lots of other issues related to configuration.

*  Changes on setting page doesn't persist https://github.com/getlantern/lantern/issues/2480
*  Lantern thinks cloud config is always a new version https://github.com/getlantern/lantern/issues/2568
*  Command line option overrides config file https://github.com/getlantern/lantern/issues/2572

fffw:
Given 4 sources of configuration (command line, setting page, local config file, cloud config) and so many components depend on them, to minimize the glue code and make interface cleaner, an implementation would be as follows:

*  Makes config as a service, each component under `flashlight` subscribes the options it cares about, but also can query an option at anytime. For flexibility, the key should be dot separated string. The callback parameters can be `key, value, source, increment`.

*  Each source publishes the changes to config service. Config service checks the precedence to determine if publish the change or not. Source also subscribes changes to update itself accordingly. 

*  Separates local config from cloud one. Command line option > setting page = local config file > cloud config (if any conflicts from local config)

myleshorton:
Cool @fffw. Does this also make the cloud config immutable by default (i.e. the cloud version is always the canonical version, with local user settings separated from that)?

fffw:
Yes.

fffw:
Extracted user settings to `settings.yaml`

PR #2613, opened 2015-06-06
make config updates replace, instead of appending to, TrustedCAs. fixes #2612

### LINKED ISSUE #2612: trustedcas should be overwritten, not appended to, in config
To prevent buildup of redundant CA entries which causes higher CPU usage, and so we can remove CAs we no longer trust or need.

aranhoide:
This is a quick fix just in case https://github.com/getlantern/lantern/issues/2390 doesn't make it to next release.


myleshorton:
LGTM @aranhoide thanks.

PR #2617, opened 2015-06-08
Fallback server checker, parallelized

uaalto:
Based on code by @aranhoide , parallelized, outputs to stdout only on error.

Note: debug logging remains from flashlight/client dependency, it must be removed


uaalto:
One thing that might require attention: is fmt thread-safe? When testing this program I didn't encounter any instance of where the output was interwoven, but I'm finding it now in other code (rarely, but still). If this is the case, I should change fmt for log.

myleshorton:
So this is to replace the old fallback server checking script, is that correct @uaalto?

uaalto:
This is indeed a fallback server checking script, optimized to perform the checks concurrently. I'm not sure if the intention is to replace the old script entirely. It is intended for automated checks in a server, that's why it tries to minimize output unless a verbosity flag is set. Maybe @aranhoide  can elaborate on the specifics of its deployment, however.

aranhoide:
Yes, this is intended to replace the old checker, which is based on lantern-java and was configured through a Google Apps app.  Also, that one took too long to run a pass of tests (basically because we were testing the proxies serially) and gave false positives too often, for whatever reason.

This fits more seamlessly with our current toolchain, since it takes the same format as genconfig does for the fallbacks list (fallbacks.json) and it uses the same code the Lantern client uses to proxy through fallbacks.

aranhoide:
@uaalto : No, fmt is not thread-safe.  log writes to stderr, while I'd rather have this script output to stdout (this being a nice way to tell library-output logs from program output too).  So maybe errors should be collected somehow to be printed in a single (the main?) goroutine.  I'll first review the rest of your code and then maybe suggest other ideas about this.

aranhoide:
And btw, this is really based on code by @oxtoacart ; I merely extracted the fallback checking code from genconfig.go and changed it to check the response, in addition to checking that the request completed without error.

src/github.com/getlantern/flashlight/genconfig/checkfallbacks.go
```diff
@@ -0,0 +1,166 @@
...
+	flag.Parse()
+
+	if *help {
+		flag.Usage()
+		os.Exit(1)
+	}
+
+	numcores := runtime.NumCPU()
+
+	if *verbose {
```
aranhoide: Nice, although if log debug goes to stderr maybe we don't need to care about limiting log output.

src/github.com/getlantern/flashlight/genconfig/checkfallbacks.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	runtime.GOMAXPROCS(numcores)
+
+	fallbacks := make(chan fallbackServer)
+	loadFallbacks(fallbacks)
+	testFallbacks(fallbacks)
+}
+
+// Load the fallback servers list file. Failure to do so will result in
+// exiting the program.
```
aranhoide: Not a big deal, but maybe it would be best for this to take a `filename string` and return `([]string, err)`, so it would be simpler to unit test.  Then let the caller handle flags and either exit on error or feed the unmarshalledFallbacks to the channel.  I'd push those basically-untestable tasks to the `main` function.

This is probably only worth changing when we actually get around to writing unit tests for this.

src/github.com/getlantern/flashlight/genconfig/checkfallbacks.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+		close(fallbacks)
+	}()
+}
+
+// Test fallback servers. Output to stdout *only* when a test fails.
+func testFallbacks(fallbacks <-chan fallbackServer) {
+	workersWg := sync.WaitGroup{}
+
+	if *verbose {
```
aranhoide: A `log.Debug` would perhaps be more appropriate here.  (You probably inherited this printout, so sorry about that. :)

src/github.com/getlantern/flashlight/genconfig/checkfallbacks.go
```diff
@@ -0,0 +1,166 @@
...
+		fmt.Printf("Spawning %d workers\n", *numConns)
+	}
+
+	workersWg.Add(*numConns)
+	for i := 0; i < *numConns; i++ {
+		// Worker: consume fallback servers from channel and signal
+		// Done() when closed (i.e. range exits)
+		go func(i int) {
+			for fb := range fallbacks {
+				testFallbackServer(fb, i)
```
aranhoide: Nice.  In order to serialize error printouts you could have testFallbackServer return an error (which, incidentally, would turn all the `fmt.Printlf(...); return` into a single `return` statement) and here send any errors to an output channel.

src/github.com/getlantern/flashlight/genconfig/checkfallbacks.go
```diff
@@ -0,0 +1,166 @@
...
+			}
+			workersWg.Done()
+		}(i+1)
+	}
+
+	workersWg.Wait()
```
aranhoide: And here you could close the output channel.  You could run testFallbacks in a goroutine and have the main goroutine consume the output channel, so when it gets closed the program can exit.

src/github.com/getlantern/flashlight/genconfig/checkfallbacks.go
```diff
@@ -0,0 +1,166 @@
...
+		fmt.Printf("%v: error reading response body: %v", ip, err)
+		return
+	}
+	body := string(bytes)
+	if body != expectedBody {
+		fmt.Printf("%v: wrong body: %s", ip, body)
+		return
+	}
+
+	if *verbose {
```
aranhoide: Again, I think you can just log this out and forget about the `*verbose` flag altogether.  I don't remember if the verbose flag was my idea; if so, sorry. :)  I didn't know that log output went to stderr.

uaalto:
I think sending errors through a channel is a nice way of serializing them. It's the first idea I had when it first occured to me that fmt might be thread-unsafe. It isn't that obvious, because output of this script is clean and appears serialized. Log module is thread-safe, though. Does Lantern have a thread-safe logging facility?
In another order of things, if I push to this branch any changes, are the commits "appended" to this pull request or I need to open a new one and reference it somehow?

aranhoide:
@uaalto : the log module is indeed thread-safe, but log output goes to stderr.  We could have the caller of this script read stderr, but I think it's more proper to just write the fallback errors (program output) to stdout, and then we don't have to worry about interference from logs in this program or any libraries it uses.

Pushing changes to this branch will update the PR.  I'm done with the first pass of code review.  When you're ready for another one, just push your changes and ping me.

uaalto:
Thank you @aranhoide ! will do!

src/github.com/getlantern/flashlight/genconfig/checkfallbacks.go
uaalto: I'm going to keep it then for the next version, since I'll output to stdout.

src/github.com/getlantern/flashlight/genconfig/checkfallbacks.go
uaalto: The reason for this was to push the server addresses lazily to the channel as demanded by the consumer threads. But since the file is anyways eagerly loaded before that, maybe the lazy feeding can be done at main. Testing of loadFallbacks would then be focused on proper deserialization of the given file.

src/github.com/getlantern/flashlight/genconfig/checkfallbacks.go
uaalto: I checked what log.Debug does, and it sends it to stdout.

PR #2621, opened 2015-06-09
Forcing extended log

xiam:
This may help us decide if [this error](https://github.com/getlantern/lantern/issues/2398) is really that bad.


fffw:
LGTM. Thanks for tracing hard.

PR #2625, opened 2015-06-10
add US round robin and IP prefixes

aranhoide:
This is live in the production peerscanner and working well.


aranhoide:
Travis failures are due to thread issues in either framed or its unit tests.

aranhoide:
This is not needed anymore, let's resubmit if we ever launch fronted servers in the US.

PR #2635, opened 2015-06-12
[don't merge] refactor config thing

fffw:
It is far from complete, but created this PR for progress tracking and documentation.

The key point is to decouple and define a clear interface, then we can change the config implementation without affecting other parts. Ideally, each component of flashlight will

1. know the option name it care about and retrieve them directly, and

2. get initial configuration and watch for changes to reconfigure itself.

Main parts should be:

1.  A config repository that each option can be set and got individually, rather than a monolithic config object passing around everywhere.

2.  A notification mechanism. Each watcher should be notified in batch if and only if some of the options it cares changed.

3.  A way to write back changes to config file.

I implemented part 1 using [viper](https://github.com/spf13/viper), part 2 and 3 still needs effort.

viper has a few features we don't need and brings lots of 3rd party libraries in, which is not good but acceptable.


fffw:
Manual verification checklist
- [ ] test effectiveness of each cli options
- [ ] check if any reference to renamed cli options
- [ ] cli precedes everything else but doesn’t presist
- [ ] setting page can save to yaml
- [ ] yam should not override cloud ones
- [ ] separated cloud.yaml to override cloud ones
- [ ] thread safety
- [ ] each components work well and can be reconfigured
  - [ ] statsreporter
  - [ ] autoupdate
  - [ ] loggly
  - [ ] geolookup
  - [ ] proxiedsites
  - [ ] analytics
  - [ ] statserver
  - [ ] config itself using hqfd
  - [ ] client
  - [ ] server!!!

myleshorton:
Honestly @fffw this seems like a bit of overkill to me. As far as I can tell, we need fewer features not more. As you originally pointed out, we currently have only a few user-configurable settings. Why not just create a separate yaml file for those and just use whatever the server tells us to use otherwise (i.e. not muck around with client-side changes to the server config)? Doesn't that give us what we want?

fffw:
Yes I also think so now @myleshorton . I was originally thinking that by stopping passing around config object we gain more flexibility on further changes, but at the same time we lose the benefits of static type checking, and the changes affects nearly each component of `flashlight`. I'll limit the changes inside `config` component and only separate cloud config from local one.

As we still need to maintain local config dynamically (manually editing or through lantern-ui), one implementation would be to keep using `yamlconf` for existing `lantern.yaml`, and totally separate cloud config from it.

I'll submit a new PR and close this one once new code is ready.

fffw:
Created new PR at https://github.com/getlantern/lantern/pull/2680, close this one.

PR #2643, opened 2015-06-15
Fix for issue 2642: environment variables are not properly set up

uaalto:


fffw:
Good catch @uaalto ! Merging.

uaalto:
Thank you @fffw !

Issue #2462, opened 2015-04-30
Detect Lantern Pi on local network using multicast


atavism:

myleshorton:
Here's the old java implementation as a point of reference:

https://github.com/getlantern/lantern/blob/32150c873e5b2492e0478d65c375118efb627532/src/main/java/org/lantern/multicast/LanternMulticast.java

uaalto:
I'm building a multicast module, testable through an independent command. This will be in a 'multicast' branch, then I will pursue integration in an 'issue-2462' branch. Does this sound good?

fffw:
Yes exactly. A branch name with the convention, and create a PR once the
code is ready for a review or discussion.

uaalto:
Progress for this is in this branch: https://github.com/getlantern/lantern/tree/multicast/src/github.com/getlantern/multicast
When I finish the standalone implementation I will open a pull-request. The remaining part in order to close this issue would be to integrate it. For that, I need to better understand what does Flashlight do with the peer list. The only reference to this I see in the code is https://github.com/getlantern/lantern/blob/32150c873e5b2492e0478d65c375118efb627532/src/main/java/org/lantern/multicast/LanternMulticast.java#L133

Issue #1617, opened 2014-06-19
Lantern Raspberry Pi port


atavism:
Need a working version ASAP. Will be a key prize for our upcoming Indiegogo campaign.

uaalto:
Is this something we are still interested in? Maybe for the Go version?

myleshorton:
Yes definitely! We actually have to do it because it was a prize for an Indiegogo campaign we did. We put it on hold with so much else going on but also because we wanted to make sure to include auto-updates from the Go version.

Actually, there's a pending task for this that would be interesting for you to take a peek at -- adding multicast discovery to make it possible to see the Lantern Pi running on the local network and to connect to it. We had that functionality in the java version but haven't ported it yet. Check out:

https://github.com/getlantern/lantern/issues/2462

uaalto:
Great! I'll handle this. I've also identified a couple more issues/features that I can work on. Thank you!

myleshorton:
Awesome @uaalto!

uaalto:
Maybe we should close the issue with this pull request too: https://github.com/getlantern/lantern/pull/2646
Lantern runs on the Pi with these commits. The remaining step would be to integrate discovery on Lantern, which would be https://github.com/getlantern/lantern/issues/2462.

PR #2632, opened 2015-06-11
Multicast discovery

uaalto:
Provide discovery of Lantern processes in the local network through SSM multicast.
This module is a step towards https://github.com/getlantern/lantern/issues/2462
* It's prepared for integration through AddPeer/RemovePeer callbacks.
* Provides tests, and a testing command that can be run standalone to check for discovery

Tested on Raspberry Pi and/with OSX

PR #2646, opened 2015-06-16
Lanter Pi + ARM target (issues 2669, 2463, 2463 and 1617)

uaalto:
Support for cross-compiling Lantern ARM target using Docker. Builds a __headless__ version.
Closes https://github.com/getlantern/lantern/issues/2463

Issue https://github.com/getlantern/lantern/issues/1617 also would be closed. Remaining step for Lantern Pi is: https://github.com/getlantern/lantern/issues/2462


uaalto:
Issue https://github.com/getlantern/lantern/issues/1617 also would be closed. Remaining step for Lantern Pi is: https://github.com/getlantern/lantern/issues/2462

myleshorton:
Hey @uaalto can you try just removing framed from your branch as well in an effort to get travis tests passing and considering we're not actually using it?

uaalto:
Sure @myleshorton . Do you mean https://github.com/getlantern/lantern/tree/valencia/src/github.com/getlantern/framed?

myleshorton:
Yeah exactly @uaalto, althoughI would just do it on your branch so we can get the tests passing, and hopefully that will help get them passing on some of these other PRs as well.

Hey I just sent you an email too -- have to cancel our call unfortunately.

myleshorton:
> hopefully that will help get them passing on some of these other PRs as well.

Once we merge to Valencia and then reverse merge or just refresh the travis checks on the other branches of course...

uaalto:
@myleshorton I just removed it from testpackages to see what happens. The tests weren't failing locally on my computer.
I'm hesitant to remove framed entirely because is being used by some other packages (probably not used either, since Lantern builds without framed). These packages are:
* waddell
* go-igdman

In any case, one of the tests builds failed at balancer, after a long chain of _ERROR balancer: balancer.go:83 No trusted dialers!_

```
ERROR balancer: balancer.go:83 No trusted dialers!
--- FAIL: TestAll-2 (1.25s)
	Location:	balancer_test.go:226
	Error:		No error is expected but got No dialers left
	Messages:	Dialing should have succeeded
```

uaalto:
Handling issue #2669 as well.

src/github.com/getlantern/flashlight/ui/ui.go
```diff
@@ -62,11 +63,21 @@ func Handle(p string, handler http.Handler) string {
...
-		return fmt.Errorf("Unable to listen at %v: %v", addr, l)
+	// If we want to allow remote connections, we have to bind all interfaces
+	if allowRemote {
+		tcpAddr, err := net.ResolveTCPAddr("tcp4", addr)
+		if err != nil {
+			panic(err) // Dev error
+		}
+		if l, err = net.Listen("tcp", ":" + strconv.Itoa(tcpAddr.Port)); err != nil {
+			return fmt.Errorf("Unable to listen on all interfaces on port: %d", tcpAddr.Port)
+		}
```
myleshorton: Nice. I might make this a little more DRY with something like:

```
if allowRemote {
...
    addrToUse = ":" + strconv.Itoa(tcpAddr.Port));
else {
    addrToUse = addr
}

if l, err = net.Listen("tcp", addrToUse); err != nil {
    return fmt.Errorf("Unable to listen at %v: %v", addr, l)
}
```

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -354,6 +366,20 @@ func exit(err error) {
...
+	signal.Notify(c,
+		syscall.SIGHUP,
+		syscall.SIGINT,
+		syscall.SIGTERM,
+		syscall.SIGQUIT)
+	go func() {
+		<-c
+		exit(nil)
+	}()
+}
```
myleshorton: Nice!

myleshorton:
Hey @uaalto this generally looks great to me. The one thing it looks like this needs is also just the multicast project added into this repo for now while we're still using gost right? Otherwise it won't compile without everyone doing their own gost get (which is a legitimate approach, just not what we've been doing). That'll allow me to review the multicast project here too.

src/github.com/getlantern/flashlight/ui/ui.go
uaalto: Yes, although IRCC I did that just to provide more specific error codes. I changed that in an upcoming commit, because even more code was being repeated unnecessarily.

uaalto:
Sure, actually now I'm making progress on them in parallel. I'll merge them in a commit.

uaalto:
With this merge of the multicast branch, the related pull request https://github.com/getlantern/lantern/pull/2632 can be closed. 

The status is as follows:
* Lantern runs on the PI in headless mode. When in headless mode, now Lantern takes care of binding all interfaces so it can be accessed from network, not just the loopback device.
* Lantern on desktop detects the Pi presence in the network (and the other way around), and the UI is informed of that (see screenshot).
* Once the remote UI for the Lantern on the Pi is opened on the desktop environment, there seems to be an issue with the websockets service for updating the Local Lanterns list. Even though it is not necessary, it is actually a bug and should be showing a link to the desktop UI.

screen shot 2015-06-24 at 1 09 47 am

uaalto:
Ok! Ready for review.
Summary: It works both ways. Lantern "desktop" finds the Pi and can open it. The Pi finds the "desktop" too, but since the desktop doesn't run in headless mode it doesn't bind the network interfaces and won't be accessible. Failure detection works well too and the UI is responsive to the Lanterns appearing/disappearing from the network.

Here's a screenshot. The left is the desktop, the right is the Pi.
screen shot 2015-06-24 at 12 09 29 pm

src/github.com/getlantern/flashlight/ui/ui.go
```diff
@@ -62,11 +63,17 @@ func Handle(p string, handler http.Handler) string {
...
-		return fmt.Errorf("Unable to listen at %v: %v", addr, l)
+	// If we want to allow remote connections, we have to bind all interfaces
+	if allowRemote {
+		if l, err = net.Listen("tcp", ":" + strconv.Itoa(tcpAddr.Port)); err != nil {
+			return fmt.Errorf("Unable to listen on all interfaces on port: %d", tcpAddr.Port)
+		}
+	} else {
+		if l, err = net.Listen("tcp", tcpAddr.String()); err != nil {
+			return fmt.Errorf("Unable to listen at %v: %v", tcpAddr.String(), l)
+		}
```
myleshorton: This is where I was suggesting making it slightly more DRY with something like:

```go
var addr string
if allowRemote {
    addr = ":" + strconv.Itoa(tcpAddr.Port)
} else {
    addr = tcpAddr.String()
}
if l, err = net.Listen("tcp", addr); err != nil {
    return fmt.Errorf("Unable to listen at %v: %v", addr, l)
}
```

myleshorton: Also, it seems like the ListenTCP method is maybe a super minor improvement. I'm not a fan of this passing strings as arguments in general TBH, but that would make it something like:

```go
var addr string
if allowRemote {
    tcpAddr.IP = nil
}
if l, err = net.ListenTCP("ipv4", tcpAddr); err != nil {
    return fmt.Errorf("Unable to listen at %v: %v", tcpAddr, l)
}
```

myleshorton: Or to avoid mutating the arguments and just create a new TCPAddr

src/github.com/getlantern/flashlight/localdiscovery/localdiscovery.go
```diff
@@ -0,0 +1,79 @@
...
+	if service == nil {
+		var err error
+		service, err = ui.Register(messageType, nil, func(write func(interface{}) error) error {
+			return write(buildPeersList())
+		})
+		if err != nil {
+			log.Errorf("Unable to register Local Discovery service: %q", err)
+			return
+		}
+	}
```
myleshorton: This nil check is to guard against Start being called multiple times, correct @uaalto? If so maybe sync.Once is a better choice, as in:

```go
once.Do(start(portToAdvertise))
```

with ```start``` as a new private method?

myleshorton: Then you could skip the check for nil

uaalto:
Besides this, I also run `go vet`, which shows no output. I'm not able to make vet produce any output in any case, so I'm not sure if it's actually working.

myleshorton:
Hey sorry for the lengthy review here @uaalto, but below are some other quick checks that jump out. I believe we can get away with not exporting most of these:

```
multicast (lantern-arm *)]$ golint
message.go:16:2: exported const TypeHello should have comment (or a comment on this block) or be unexported
multicast.go:29:6: exported type Multicast should have comment or be unexported
multicast.go:43:6: exported type PeerInfo should have comment or be unexported
multicast.go:49:1: comment on exported function JoinMulticast should be of the form "JoinMulticast ..."
multicast.go:72:1: comment on exported method Multicast.StartMulticast should be of the form "StartMulticast ..."
multicast.go:78:1: comment on exported method Multicast.ListenPeers should be of the form "ListenPeers ..."
multicast.go:84:1: comment on exported method Multicast.LeaveMulticast should be of the form "LeaveMulticast ..."
```

myleshorton:
Here are some more, using the errcheck tool (go get github.com/kisielk/errcheck)

```
multicast (lantern-arm *)]$ errcheck
github.com/getlantern/lantern/src/github.com/getlantern/multicast/multicast.go:81:19	go mc.listenPeers()
github.com/getlantern/lantern/src/github.com/getlantern/multicast/multicast.go:96:10	mc.write(msg)
github.com/getlantern/lantern/src/github.com/getlantern/multicast/multicast.go:130:11	mc.write(msg)
github.com/getlantern/lantern/src/github.com/getlantern/multicast/multicast.go:156:27	mc.Conn.SetReadDeadline(time.Now().Add(time.Duration(mc.Period) * time.Second))
```

src/github.com/getlantern/flashlight/localdiscovery/localdiscovery.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		log.Errorf("Unable to register Local Discovery service: %q", err)
+		return
+	}
+
+	mc = multicast.JoinMulticast()
+
+	mc.AddPeerCallback = func(peer string, peersInfo []multicast.PeerInfo) {
+		peersMutex.Lock()
+		lastPeers = peersInfo
```
myleshorton: Why hold the lastPeers variable locally at all? Why not just do:

```go
service.Out <- buildPeersList(peersInfo)
```

```peersInfo```  here is already a copy, so there's no data we're really protecting. You can also eliminate all the locks then too.

File: src/github.com/getlantern/flashlight/localdiscovery/localdiscovery.go [hunk 6]
```diff
@@ -0,0 +1,90 @@
...
+		mc.Payload = portToAdvertise
+		mc.StartMulticast()
+	}
+
+	mc.ListenPeers()
+
+	go func() {
+		c := time.Tick(updatePeriod * time.Second)
+		for range c {
+			service.Out <- buildPeersList()
```
myleshorton: If we're already calling ```service.Out <- buildPeersList()``` every time peers change, do we need to also do it periodically?

src/github.com/getlantern/flashlight/localdiscovery/localdiscovery.go
```diff
@@ -0,0 +1,90 @@
...
+		return write(buildPeersList())
+	})
+	if err != nil {
+		log.Errorf("Unable to register Local Discovery service: %q", err)
+		return
+	}
+
+	mc = multicast.JoinMulticast()
+
+	mc.AddPeerCallback = func(peer string, peersInfo []multicast.PeerInfo) {
```
myleshorton: Personally I would also have the ```JoinMulticast``` function just take two function arguments for listening for joins and leaves. That way you could avoid exporting the ```AddPeerCallback``` and ```RemovePeerCallback``` methods, further limiting the exposed state and mutability of ```Multicast```.

myleshorton: You could override it as well with one that takes no args and just passes empty functions...

uaalto: The reason for this design is so the API is simpler if you don't need to set any of the callbacks, given that Go has no variable or named arguments. If you require arguments you need to provide them always, and also you have to check for documentation to recall the position, while if they are assigned to exported members of the struct you see clearly what they mean and if you don't add callbacks then there is no clutter.
On the other hand, your design forces the user to really think about those callbacks since is forced to provide them.
I'll change the design then.

src/github.com/getlantern/flashlight/localdiscovery/localdiscovery.go
uaalto: There was a (very short) delay in some situations. I thought about this and concluded that the bandwidth usage was worth the user experience. What do you think should prevail?

src/github.com/getlantern/flashlight/localdiscovery/localdiscovery.go
uaalto: Yeah, it's not used. Will remove it.

uaalto:
Also , Golint and Errcheck: I'll use them for them before any pull request.

src/github.com/getlantern/flashlight/localdiscovery/localdiscovery.go
myleshorton: Yeah that's why I was also suggesting adding a separate method that
internally would just call the same thing with empty functions. Again not a
big deal but to me I just think these various possible states kicking
around causes issues in the end, and in this case requires the internal
code to always check for nil on those functions.

On Fri, Jun 26, 2015 at 1:51 PM, Ulysses Aalto <notifications@github.com>
wrote:

> In src/github.com/getlantern/flashlight/localdiscovery/localdiscovery.go
> <https://github.com/getlantern/lantern/pull/2646#discussion_r33394686>:
>
> > +		panic("The " + messageType + " service is already registered")
> > +		return
> > +	}
> > +
> > +	var err error
> > +	service, err = ui.Register(messageType, nil, func(write func(interface{}) error) error {
> > +		return write(buildPeersList())
> > +	})
> > +	if err != nil {
> > +		log.Errorf("Unable to register Local Discovery service: %q", err)
> > +		return
> > +	}
> > +
> > +	mc = multicast.JoinMulticast()
> > +
> > +	mc.AddPeerCallback = func(peer string, peersInfo []multicast.PeerInfo) {
>
> The reason for this design is so the API is simpler if you don't need to
> set any of the callbacks, given that Go has no variable or named arguments.
> If you require arguments you need to provide them always, and also you have
> to check for documentation to recall the position, while if they are
> assigned to exported members of the struct you see clearly what they mean
> and if you don't add callbacks then there is no clutter.
> On the other hand, your design forces the user to really think about those
> callbacks since is forced to provide them.
> I'll change the design then.
>
> —
> <https://github.com/getlantern/lantern/pull/2646/files#r33394686>.
>



-- 
--
President
https://www.getlantern.org

uaalto: I get the point of the nil checks, which I think is the strongest argument against my design. I propose a single method, but that will take a nil callback passed as argument and just consider it a default callback.
The rationale is that if I provide a specialized method with no callbacks as parameters, then I think I should provide two more for just add and remove callbacks exclusively, for consistency. A single method which takes care of handling nil arguments.
Also, this is typical interface design in C, I don't know about Go. What do you think?

src/github.com/getlantern/flashlight/localdiscovery/localdiscovery.go
uaalto: I removed it. The hello function takes care of sending this information if the UI is closed when the event is received. I will test again with the Pi back at home.

src/github.com/getlantern/flashlight/localdiscovery/localdiscovery.go
uaalto: Actually, coming back to this I realized that it *is* used. It is necessary particularly for the _hello_ function, which will push the peer list to the UI as soon as it connects. While we can rely on the local peer list received as argument when there is an Add/Remove, we need this data to be held for this situation I describe.

uaalto:
Comments about this last commit:
* I couldn't test the changes with the Pi physical device, I have to do that back at home.
* There are some of the things that we discusses that once I got to the code I recalled why they were there, so I couldn't make all changes. I have written replied your comments where that's the case.

myleshorton:
Hey these all looks great @uaalto! I am getting a test failure now though:

```
go test
--- FAIL: TestMulticastMessages (0.02s)
	multicast_test.go:121: Multicast Hello message is incorrectly formatted
```

uaalto:
Oh, weird. It passes in my machine and shouldn't be machine-dependent or run-dependent.

myleshorton:
Hey @uaalto sorry for the delay! Here's what I get:

```
2015/07/08 17:47:42 {"Type":2,"Payload":""}
2015/07/08 17:47:42 {2 }
```

uaalto:
The error doesn't make much sense, specially since I can't reproduce it. The message type you get is that of a Bye. So either you are receiving it from previous test functions' or the wrong type is being sent. I don't see the first one happening with the current test code. Could you please try it again with these tiny changes?

myleshorton:
So @uaalto my last commit reverse merging from valencia completely clobbers the UI side of this -- sorry. I think it makes more sense to just do them again on top of the new code in a new PR. Hope that's OK!

uaalto:
Sure, no problem! I was expecting it :)

uaalto:
@myleshorton what was wrong exactly with your reverse merge? I see the UI working well and integrated with changes.

myleshorton:
Oh it clobbered all of your stuff in settings I thought but maybe it's ok @uaalto?

uaalto:
The changes to the UI are very localized to partials/modals.html, so the merge was clean.

uaalto:
I've tested everything again with the Pi, and in a different network. It works and the UI is fine.

myleshorton:
OK sweet! I'm going to merge this in. I think the one thing we have to be careful to test is the multicast server socket opening triggering warnings on different OSes, but hopefully we'll be OK.

uaalto:
Ok great!!!
I'll keep an eye on that potential issue.

PR #2652, opened 2015-06-17
Add latest version of FT to cloud config

atavism:
We store the latest version of FireTweet in the cloud config to compare against the version currently running on a user's device to determine whether or not the available update notification is presented.

### DIFF
--- a/src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
+++ b/src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
@@ -1,7 +1,7 @@
 # cloud.yaml contains the default configuration that's made available on the
 # internet.
 client:
-  firetweetversion: "0.0.6"
+  firetweetversion: "{{.ftVersion}}"
   frontedservers:
   - host: nl.fallbacks.getiantem.org
     port: 443
--- a/src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
+++ b/src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
@@ -1,6 +1,7 @@
 # cloud.yaml contains the default configuration that's made available on the
 # internet.
 client:
+  firetweetversion: "0.0.6"
   frontedservers:
   - host: nl.fallbacks.getiantem.org
     port: 443
--- a/src/github.com/getlantern/flashlight/genconfig/cloud.yaml
+++ b/src/github.com/getlantern/flashlight/genconfig/cloud.yaml
@@ -817,25 +817,25 @@ client:
     - domain: 10minutemail.com
       ipaddress: 162.159.250.16
     - domain: 2ch.hk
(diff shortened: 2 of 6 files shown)


atavism:
Related to https://github.com/getlantern/firetweet/pull/134

src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
```diff
@@ -1,6 +1,7 @@
 # cloud.yaml contains the default configuration that's made available on the
 # internet.
 client:
+  firetweetversion: "0.0.6"
```
fffw: It might be too tightly coupled, as we need to rerun `genconfig` for each new FT version. Alternative would be to put a `version.txt` besides `firetweet.apk`, and check the content directly. It's rather easy to update the txt file at the same time when updating apk.

atavism: OK we could do that. I feel like we'll probably start releasing a new version of FT only every few weeks though so wouldn't be such a big deal running ``genconfig`` every time

atavism: Putting the version file here: https://raw.githubusercontent.com/firetweet/downloads/master/version.txt
We can just update that file whenever we deploy the release APK as you said

myleshorton: Actually while just doing a fresh Lantern build I feel like regenerating the config every time automatically is probably a good idea, as the certs, IPs, etc, could theoretically change over time, and it's worthwhile to periodically update it.

atavism: Yeah, that was my thinking. I like regenerating the config more often. We could just run ``genconfig.bash`` whenever we run ``deploy.bash`` too (and still read the FT version number from file or pass it as an argument).

myleshorton: Yeah we could integrate that into the Makefile. You mind ticketing that on the Lantern side?

atavism: I created it here: https://github.com/getlantern/lantern/issues/2663

myleshorton:
OK with #2663 all set now is this one all set @atavism and @fffw?

atavism:
@myleshorton I made one more change to this PR:

- In genconfig, prior to generating the config, we fetch ``https://raw.githubusercontent.com/firetweet/downloads/master/version.txt`` and set the FT version number in ``cloud.yaml.tmpl`` -- consequently, we don't have to pass in any additional arguments and are always using the latest version number from ``version.txt`` -- which will just get incremented whenever we run ``deploy.bash``

myleshorton:
Nice -- LGTM @atavism thanks.

PR #2655, opened 2015-06-18
Detour detect ECONNREFUSED to resolve #2638

### LINKED ISSUE #2638: In Mac, I couldn't access gmail.com even after starting lantern
Steps:

1. Modified the DNS server to block all sites.
2. Now start lantern without proxy all
3. Unable to access gmail.com but able to access facebook, twitter, github

I am sure gmail.com is in the default proxied list but still unable to access after launching lantern. When I tested the same in Windows yesterday, it was working fine.

I am able to access the same when I used proxy all traffic.

fffw:
Now if we use fake DNS to point any site to 127.0.0.1, detour will work.

`lantern.log:Jun 18 01:53:41.667 - DEBUG detour: detour.go:136 Dial initially to gmail.com:443 failed, try detour: dial tcp 127.0.0.1:443: connection refused`

### DIFF
--- a/src/github.com/getlantern/detour/detect.go
+++ b/src/github.com/getlantern/detour/detect.go
@@ -3,7 +3,6 @@ package detour
 import (
 	"bytes"
 	"net"
-	"runtime"
 	"syscall"
 )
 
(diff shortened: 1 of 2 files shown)


myleshorton:
Do you think this is actually the right approach @fffw? I'm a little reluctant to do it if it's just to get the test working.

fffw:
ECONNREFUSED is returned on Android if connection is reset, I guess it may also happen in other circumstances we don't know. Accessing to an non-exist port is rare and sending the request via proxy does little harm.

ECONNRESET is really hard to simulate, it would be good to have a way to substitute.

myleshorton:
OK LGTM. Merging!

PR #2664, opened 2015-06-20
removed dependency on framed

myleshorton:
This also ends up removing the dependency on waddell, nattyway, and go-natty, all of which are currently unused but depend on framed, and framed is causing erroneous test failures.


myleshorton:
Guys I'm going to merge this in since it just removes these unused packages and tweaks one other test that used framed (go-igdman_test.go) in a trivial manner. This will hopefully get the tests for some of these other PRs passing.

Issue #2670, opened 2015-06-23
Lantern beta6 uses beta5 binary on Linux


xiam:
This is due to a bug in the wrapper script that does not check whether the main binary is newer than the old one.

PR #2672, opened 2015-06-23
When the system's Lantern binary is changed (on Linux) use it instead of the copy

xiam:
Added a verification that checks whether the system's Lantern binary got changed and replaces the user's copy of the binary in such case. 

See #2670.


myleshorton:
LGTM!

PR #2674, opened 2015-06-23
Adding a genconfig target to the Makefile

### LINKED ISSUE #2663: Integrate genconfig scripts with Makefile
This would make it easier to put out a new version of the config whenever we build Lantern.

xiam:
The Makefile will now run `genconfig` when preparing for a QA release.

Closes #2663

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,13 +30,11 @@ LANTERN_BINARIES_PATH ?= ../lantern-binaries
 
 GO_MOBILE_REVISION=47553f4d4275316f2610b3204f8f9ecaf3ce63a7
 
-GH_USER := getlantern
-#GH_USER := xiam
+GH_USER ?= getlantern
 
-GH_RELEASE_REPOSITORY := lantern
+GH_RELEASE_REPOSITORY ?= lantern
 
-S3_BUCKET := lantern
-#S3_BUCKET := xiam-lantern-test-1
+S3_BUCKET ?= lantern
 
 DOCKER_IMAGE_TAG := lantern-builder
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -323,7 +321,7 @@ package-darwin: require-version require-appdmg require-svgexport darwin
 
 binaries: docker genassets linux windows darwin
 
-packages: require-version require-secrets clean binaries package-windows package-linux package-darwin
+packages: require-version require-secrets clean genconfig binaries package-windows package-linux package-darwin
```
xiam: The packages target is a required step for release-qa, and it's already slow so it makes sense to use genconfig here.

Makefile
```diff
@@ -30,13 +30,11 @@ LANTERN_BINARIES_PATH ?= ../lantern-binaries
 
 GO_MOBILE_REVISION=47553f4d4275316f2610b3204f8f9ecaf3ce63a7
 
-GH_USER := getlantern
```
xiam: A bit of cleaning now that I'm here. Removing commented out value and using a conditional declaration.

Makefile
```diff
@@ -424,7 +422,7 @@ test-and-cover:
 		tail -n +2 profile_tmp.cov >> profile.cov; \
 	done
 
-android-lib: docker-golang-android
+android-lib: docker-golang-android genconfig
```
xiam: Runnin `genconfig` before building the Android's libgojni.

fffw:
Tested both `make packages` and `make android-lib`, works pretty well. Ideally we can check for the timestamp of the three generated files to avoid unnecessary rerun of `genconfig`, but as it happens rarely, I think it's ok to rerun each time. Merging.

PR #2680, opened 2015-06-25
separate local config from cloud one

fffw:
It is another attempt to resolve https://github.com/getlantern/lantern/issues/2390. It just implements our initial goal to make cloud config immutable.

### DIFF
--- a/src/github.com/getlantern/flashlight/config/cloud.go
+++ b/src/github.com/getlantern/flashlight/config/cloud.go
@@ -1,0 +1,65 @@
+package config
+
+import (
+	"compress/gzip"
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"net/http"
+	"time"
+
+	"github.com/getlantern/proxiedsites"
+
+	"github.com/getlantern/flashlight/client"
+)
+
+type cloudConfig struct {
+	Client       *client.ClientConfig
+	ProxiedSites *proxiedsites.Config // List of proxied site domains that get routed through Lantern rather than accessed directly
+	TrustedCAs   []*CA
+}
+
+const (
+	etag = "X-Lantern-Etag"
+)
+
+func cloudPollSleepTime() time.Duration {
+	return time.Duration((CloudConfigPollInterval.Nanoseconds() / 2) + rand.Int63n(CloudConfigPollInterval.Nanoseconds()))
+}
+
+func fetchCloudConfig(url string) ([]byte, error) {
+	log.Debugf("Checking for cloud configuration at: %s", url)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to construct request for cloud config at %s: %s", url, err)
+	}
+	if lastCloudConfigETag[url] != "" {
+		// Don't bother fetching if unchanged
+		req.Header.Set(ifNoneMatch, lastCloudConfigETag[url])
+	}
+
+	// make sure to close the connection after reading the Body
+	// this prevents the occasional EOFs errors we're seeing with
+	// successive requests
+	req.Close = true
+
+	resp, err := httpClient.Load().(*http.Client).Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to fetch cloud config at %s: %s", url, err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == 304 {
+		log.Debugf("Config unchanged in cloud")
+		return nil, nil
+	} else if resp.StatusCode != 200 {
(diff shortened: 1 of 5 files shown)


fffw:
Still needs to apply cloud config defaults separately. Pending.

src/github.com/getlantern/flashlight/config/cloud.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	if b == nil {
+		return
+	}
+	newCfg := emptyCloudConfig()
+	if err = newCfg.fromBytes(b); err != nil {
+		log.Errorf("Error parse cloud config: %s", err)
+		return
+	}
+	newCfg.Client.SortFrontedServers()
```
fffw: We have only 1 fronted server in default config, so I move it here, in case there are more than 1 in cloud.yaml

fffw:
Newly generated `lantern.yaml` will still contain those empty fields (`nil` or `[]`), and any modification on these fields will not take effect.

fffw:
@myleshorton Mind taking a look at this one?

myleshorton:
np @fffw. working on the ARM stuff first but will circle back around to it. Thanks!

myleshorton:
Hey @fffw I like the general direction here, although honestly I would be even a little more aggressive and make things even a little simpler. I.e. do we really even need a yamlconf Manager at all anymore? As a first pass, why not just always use whatever the cloud returns and not even worry about anything on disk -- i.e. just always overwrite what's there? 

We still need to separate out user-configurable, mutable values I think, such as whether or not to report stats, proxy all, etc, but to me those should be in a completely different "settings" system or something like that that doesn't touch the config at all. To me the above points to having a single config.go or cloud.go, something like that, that just fetches the new config and applies it repeatedly.

At a later date to me we could add a local file like cloud.yaml.overrides, or something like that, that always overrides cloud settings to allow for easier debugging, but to me we should only add that as necessary.

What do you think?

fffw:
@myleshorton Maybe I didn't understand the difference of "settings" and "configuration" here. In this PR, cloud config items is indeed separated from local mutable ones, except that they co-existed in the same `Config` object. If what you mean is to further split the two and use separate logic for them, I'd prefer to make each item totally independent, so each sub-component can retrieve only what it cares and get notified when they changes.

What `yamlconf` did and still does is to monitor the changes of mutable values in `lantern.yaml`, and to write any changes from Lantern UI to it. Once we removed `yamlconf`, we still need to do both manually.  In [my previous attempt](https://github.com/getlantern/lantern/pull/2635), I used `fsnotify` to keep track of file changes, and wrote YAML file manually. Is that the way you think?

myleshorton:
Hey sorry for the delay @fffw. Yeah I guess I'm proposing to separate out the Config object entirely -- i.e. skip any "merge" stage and just propagate the config as it's returned from the server to all relevant components. I agree it would be nice to make everything only aware of sub-components, but that strikes me as a much more involved change with a relatively limited payoff.

To me we could then treat user-configurable settings differently, with maybe a "User" or a "Settings" object that initially only has a few fields for user-configurable things. We would then save that to disk and as a first step not even worry about if it's changed on disk but just persist it so that we preserve user settings between runs.

Does that make sense?

fffw:
@myleshorton I like the idea of simplification, but might better take small
steps?

One concern is that if there's a clear dividing line between "settings" and
"configuration" to sub-components. For example, [`client.Configure`](
https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/client/client.go#L87)
cares about `MinQOS` and `ProxyAll`, as well as `TrustedCAs`. `ProxyAll` is
something user cares about, but `MinQOS`? Similar for `reportingPeriod` and
`statshubAddr` of statshub. To me, the distinction should be at the driver
level, not sub-component level.

myleshorton:
Hey sorry for the delay @fffw! It seems like we could initially just separate out:

1. Things the user can change (not much -- only used in a couple of places)
1. Things we can set in the cloud.yaml

I think we can actually eliminate MinQOS, as it's basically unused. I actually don't even see ```reportingPeriod``` in there. Where is that exactly? 

Can you describe the "driver" versus "sub-component" level a little bit more? Are you thinking of things like keeping ClientConfig? That's totally fine with me. The main thing I care about is the the config from the server always just completely overwrites whatever was there before so we eliminate some of the bugs we've been experiencing along with the associated complexity.

myleshorton:
Hey @fffw I'm wondering if the change in beta 8 to just include the version number in the config file to avoid conflicts with config files from older versions might just be enough for accomplishing what we want for now. What do you think?

fffw:
I just found that versioned config file didn't persist user changes such as "proxyall" across versions. Considering auto-update, this may make user confusing.

Anyway, this PR can be closed. We can have much simpler implementation.

PR #2683, opened 2015-06-25
Fix for issue-2682

uaalto:
Simple fix, just relying on gomobile's updates. Additionally, the double-patch trick is not necessary any longer.


myleshorton:
LGTM!

PR #2700, opened 2015-07-07
open manoto fb after first run

myleshorton:


fffw:
If this is specific to Manoto, can we add another build flag so we can tweak Makefile to build 2 binaries, one for Manoto and one for others?

myleshorton:
yeah I think that's a good idea @fffw

myleshorton:
This is maybe a little hackish but what do you think @fffw? Certainly open to better solutions, but this is at least straightforward.

sheena3:
This works fine in chrome and Firefox. Manoto fb page is not opening in IE 11 when Lantern is run for the first time.

sheena3:
Oops..I thought this is similar to the issue. I will open a separate issue.

fffw:
And, Manoto page will open each time when you click "Show Lantern", not the
first time, which disturbs user a lot. @myleshorton, is this what Manoto
asked for or unintentionally?

sheena3:
@fffw Okay. My understanding based on this issue is, manoto fb page opens up when the user installs the Lantern and runs it for the very first time. After that, we should never be showing it. 

How it works now is, manoto fb page is opening up whenever Lantern is run (not just when using "Show Lantern"). But in IE 11, it is not opening the manoto fb page during the first run alone. After that, it is fine.

PR #2701, opened 2015-07-07
Use smaller screenshots

atavism:


myleshorton:
LGTM. I need to dig into these new test failures as well.

PR #2702, opened 2015-07-09
code review suggestions

fffw:
The idea of changing to manoto url and back is good for me as it doesn't require large chunk of changes to Makefile. 

The only thing is that it confused me of how it works. I tweaked a bit so it looks more intuitive to me. If you think the tweak is good @myleshorton , I have no other comments on https://github.com/getlantern/lantern/pull/2700 .


myleshorton:
LGTM!

PR #2705, opened 2015-07-09
Add translated welcome message

atavism:


myleshorton:
Sweet @atavism! Should we add the translated string directly to fa_IR.json as well, or are you going the proper route and adding it all through transifex then syncing?

myleshorton:
Nice. Merging!

PR #2714, opened 2015-07-10
Only open Manoto page once

atavism:
@myleshorton This will make sure the Manoto page only opens a single time and not when we click "Show Lantern"


sheena3:
@atavism So manoto fb page opens only for the first time when Lantern runs, right? After that no matter how it is launched, it will just open the Lantern browser alone. Let me know if my understanding is correct.

myleshorton:
Oh good fix!!

PR #2718, opened 2015-07-11
Ui improvements

fffw:
Use fixed height, so sites will not cover Lantern logo and title on small window, with a few other tunes. Tested on OSX + Chrome and Windows 7 + IE10.


atavism:
LGTM!

PR #2732, opened 2015-07-14
Gets rid of race condition on startup closes #2719

### LINKED ISSUE #2719: IE 11 sometimes displays an error opening Lantern page after restarting
Steps:

1. Make IE as default browser
2. In Lantern browser, change the setting to Run Lantern automatically on System Startup
3. Shut down the system
4. When IE opens (WiFi is already turned ON) the Lantern browser, it says "This page can't be displayed". Same goes for the manoto TV fb page.

I tried "Show Lantern" in system tray to reopen but it didn't work either. I had to quit Lantern and reopen it. That too didn't work. It worked only when I tried for the third time.  Let me know if you need the logs. Even the normal unblocked sites didn't open (google.com, gmail.com, yahoo.com etc.). That's confusing!!!

P.S: I checked in chrome and Lantern was proxying. It is just that it doesn't work in IE.

myleshorton:
I'm not 100% sure why it seems to fix the issue, but this change removes a race condition on startup that could make the following sequence run in almost any order:

1. Lantern UI server port comes up
1. Lantern opens browser
1. Lantern proxy server comes up
1. Lantern sets the system proxy

My theory is that IE can get confused in some cases depending on the actual order those steps get executed, and in all my testing this change fixes the problem and ensures those steps always proceed as follows:

1. Lantern UI server port comes up
1. Lantern proxy server comes up
1. Lantern sets the system proxy
1. Lantern opens browser

### DIFF
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -208,12 +208,6 @@ func runClientProxy(cfg *config.Config) {
 			exit(fmt.Errorf("Unable to start UI: %v", err))
 			return
 		}
-		/*
-			if showui {
-				// Launch a browser window with Lantern.
-				ui.Show()
-			}
-		*/
 	}
 
 	applyClientConfig(client, cfg)
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -208,10 +208,12 @@ func runClientProxy(cfg *config.Config) {
 			exit(fmt.Errorf("Unable to start UI: %v", err))
 			return
 		}
-		if showui {
-			// Launch a browser window with Lantern.
-			ui.Show()
-		}
+		/*
+			if showui {
+				// Launch a browser window with Lantern.
+				ui.Show()
+			}
+		*/
 	}
 
 	applyClientConfig(client, cfg)
@@ -229,7 +231,16 @@ func runClientProxy(cfg *config.Config) {
 
 	go func() {
 		addExitFunc(pacOff)
-		client.ListenAndServe(pacOn)
+		client.ListenAndServe(func() {
+			pacOn()
+			if showui {
+				// Launch a browser window with Lantern but only after the pac
+				// URL and the proxy server are all up and running to avoid
+				// race conditions where we change the proxy setup while the
+				// UI server and proxy server are still coming up.
+				ui.Show()
+			}
+		})
 	}()
 }
 
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -225,7 +225,7 @@ func runClientProxy(cfg *config.Config) {
 
 	go func() {
 		addExitFunc(pacOff)
-		client.ListenAndServe(func() {
+		err := client.ListenAndServe(func() {
 			pacOn()
 			if showui {
 				// Launch a browser window with Lantern but only after the pac
@@ -235,6 +235,9 @@ func runClientProxy(cfg *config.Config) {
 				ui.Show()
 			}
 		})
+		if err != nil {
+			log.Errorf("Error calling listen and serve: %v", err)
+		}
 	}()
 }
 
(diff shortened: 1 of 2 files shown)


myleshorton:
Hey @fffw would you mind taking a look at this one? Or @atavism if you're free too.

src/github.com/getlantern/flashlight/ui/ui.go
```diff
@@ -92,15 +92,25 @@ func Start(addr string) error {
 // Show opens the UI in a browser. It will wait for the UI addr come up for at most 3 seconds
 func Show() {
 	go func() {
-		addr, _ := url.Parse(uiaddr)
-		if err := waitforserver.WaitForServer("tcp", addr.Host, 3*time.Second); err != nil {
-			log.Error(err)
+		addr, er := url.Parse(uiaddr)
+		if er != nil {
```
atavism: Maybe just make this ``err`` to be consistent?

src/github.com/getlantern/flashlight/ui/ui.go
```diff
@@ -92,15 +92,25 @@ func Start(addr string) error {
...
+			log.Errorf("Could not parse url `%v` with error `%v`", uiaddr, er)
 			return
 		}
-		open.Run(uiaddr)
+		if err := waitforserver.WaitForServer("tcp", addr.Host, 10*time.Second); err != nil {
```
atavism: Are we increasing this to 10 seconds now just to be on the safe side, in case there's some delay starting the UI server?

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -229,7 +225,16 @@ func runClientProxy(cfg *config.Config) {
...
+		client.ListenAndServe(func() {
+			pacOn()
+			if showui {
+				// Launch a browser window with Lantern but only after the pac
+				// URL and the proxy server are all up and running to avoid
+				// race conditions where we change the proxy setup while the
+				// UI server and proxy server are still coming up.
+				ui.Show()
+			}
+		})
```
atavism: Yeah, it makes sense to wait to launch the browser window only after we've set Lantern as the system proxy. Nice change! Works well for me on Chrome, FF, and IE.

src/github.com/getlantern/flashlight/ui/ui.go
myleshorton: Done!

src/github.com/getlantern/flashlight/ui/ui.go
myleshorton: Yeah, I basically just don't see any advantage to making it too short. If there's some unexpected delay with the the proxy starting, this will at least allow Lantern to still run.

myleshorton:
Hey one sec @atavism -- I actually just realized there might be another race in there..

myleshorton:
Hey @atavism false alarm. I was worried about those Serve calls being on go routines, but if you think about it that's not really a problem since we're already listening at that point, accepting incoming connections, and the only thing that reads from those connections is the code inside the Serve calls. So the worst case is there are some connections buffering data that doesn't get read until those Serve goroutines run, but that's fine.

I actually removed that WaitForServer line taking this into account, as we know for sure that the server is already listening at that point so don't need to double check it.

src/github.com/getlantern/flashlight/flashlight.go
atavism: The ``ListenAndServe`` callback will only be invoked if we successfully start accepting connections, but if we get an error, we really aren't accounting for it or logging it anywhere. Would be a good change to make while we're in here?

src/github.com/getlantern/flashlight/ui/ui.go
```diff
@@ -90,17 +88,22 @@ func Start(addr string) error {
 }
 
 // Show opens the UI in a browser. It will wait for the UI addr come up for at most 3 seconds
```
atavism: This comment about the 3 second wait is no longer applicable.

myleshorton: fixed!

src/github.com/getlantern/flashlight/flashlight.go
myleshorton: Done!

atavism:
ok LGTM!

myleshorton:
Awesome thanks @atavism!

Issue #2608, opened 2015-06-04
Verify Lantern continues to work if either domain fronting or chained proxies stop working


myleshorton:

xiam:
Opened one related issue for discussion: https://github.com/getlantern/lantern/issues/2622

xiam:
Another one for discussion: https://github.com/getlantern/lantern/issues/2623

xiam:
Also confirmed Lantern works with tcp port 80 closed (no chained proxies at all). Cool.

Also without port 53 udp but we should not rely on it anyways.

xiam:
We should discuss this one too: https://github.com/getlantern/lantern/issues/2624

xiam:
**Test 1**. Removing ALL chainedservers from `lantern.yaml`:

```
cloudconfig: "disabled-for-testing"
cloudconfigca: ""
chainedservers: {}
client:
  proxyall: true
  frontedservers:
    masqueradeset: cloudflare
    insecureskipverify: true
```

The rest of the `lantern.yaml` was omitted for clarity.

Result: I can browse the BBC and Facebook sites normally and even watch Youtube videos.

```
Jul 15 16:24:40.855 - DEBUG balancer: balancer.go:96 Dialing tcp://ichef.bbci.co.uk:80 with (trusted) fronted proxy at issue-2184.getiantem.org:443 using masquerade set testflare
Jul 15 16:24:40.855 - DEBUG balancer: balancer.go:96 Dialing tcp://ichef.bbci.co.uk:80 with (trusted) fronted proxy at issue-2184.getiantem.org:443 using masquerade set testflare
Jul 15 16:24:40.855 - DEBUG balancer: balancer.go:96 Dialing tcp://ichef.bbci.co.uk:80 with (trusted) fronted proxy at issue-2184.getiantem.org:443 using masquerade set testflare
Jul 15 16:24:40.856 - DEBUG balancer: balancer.go:96 Dialing tcp://ichef.bbci.co.uk:80 with (trusted) fronted proxy at issue-2184.getiantem.org:443 using masquerade set testflare
DEBUG balancer: balancer.go:96 Dialing tcp://ichef.bbci.co.uk:80 with (trusted) fronted proxy at issue-2184.getiantem.org:443 using masquerade set testflare
```

**Test 2**. Using a chained proxy that will not return a successful response.

```
  chainedservers:
    fallback-127.0.0.1:
      addr: 127.0.0.1:443
      pipelined: true
      weight: 1000000
[...]

PR #2734, opened 2015-07-15
Adding version to config file name closes #2731

### LINKED ISSUE #2731: Make sure old config files on disk aren't able to give bad config URLs

myleshorton:
This just adds the Lantern version to the yaml config file name on disk. That prevents Lantern from ever reading config files from old versions, which I think is exactly what we want.

### DIFF
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -117,7 +117,7 @@ func doMain() error {
 
 	parseFlags()
 
-	cfg, err := config.Init(packageVersion)
+	cfg, err := config.Init(version)
 	if err != nil {
 		return fmt.Errorf("Unable to initialize configuration: %v", err)
 	}
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -117,7 +117,7 @@ func doMain() error {
 
 	parseFlags()
 
-	cfg, err := config.Init()
+	cfg, err := config.Init(packageVersion)
 	if err != nil {
 		return fmt.Errorf("Unable to initialize configuration: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


myleshorton:
Hey @aranhoide what do you think of this one?

myleshorton:
Or anyone else for that matter =).

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -117,7 +117,7 @@ func doMain() error {
 
 	parseFlags()
 
-	cfg, err := config.Init()
+	cfg, err := config.Init(packageVersion)
```
atavism: Do we want to use ``version`` here instead of ``packageVersion`` -- ``packageVersion`` looks like it defaults to ``9999.99.99``, and that's to avoid auto-updating in development mode. ``version`` contains the current git revision so we'd get ``lantern-bae7606.yaml`` instead of ``lantern-9999.99.99.yaml`` which I think is more informative.

myleshorton: Yeah I initially just made the change but then reverted it. In production they'll be the same values, but I think in development the tricky thing is that the commit will continually change, so if you're actually testing using that yaml file like we are for #2608 you might get screwed up if you're development pattern involves committing locally a lot, for example.

Does that make sense?

xiam: Hello @atavism!

`packageVersion` was initially intended for autoupdates only. If no `VERSION` is passed to the Makefile it will use the git revision and when compiled in development mode it will use `9999.99.99`. I think the git revision would be more informative but I also think it would be harder to  associate it with an actual release, as we'll always have to lookup the git revision.

myleshorton: Yeah I agree @xiam. Also in production they're always the same regardless.

aranhoide:
LGTM

myleshorton:
Thanks guys!

PR #2738, opened 2015-07-15
Make fronted servers trusted by default closes #2737

### LINKED ISSUE #2737: Lantern doesn't consider fronted servers trusted by default

myleshorton:
This allows fronted servers to proxy http traffic if chained servers are blocked.


myleshorton:
Note this is also particularly an issue before downloading the cloud config file, but is important in case there's any issue getting that file.

aranhoide:
LGTM

PR #2739, opened 2015-07-15
disable CloudFront/DNSimple support

aranhoide:
Temporarily disable registering and unregistering CloudFront (and therefore DNSimple) entries in peerscanner.

### DIFF
--- a/src/github.com/getlantern/peerscanner/host.go
+++ b/src/github.com/getlantern/peerscanner/host.go
@@ -265,7 +265,8 @@ func (h *host) run() {
 			h.pause()
 			checkImmediately = true
 		case <-h.initCfrCh:
-			h.doInitCfrDist()
+			// (CloudFront/DNSimple support temporarily disabled)
+			// h.doInitCfrDist()
 		case <-pauseTimer.C:
 			log.Debugf("%v had no successful checks or resets in %v, pausing", h, pauseAfter)
 			h.pause()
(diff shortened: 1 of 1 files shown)


xiam:
Travis is showing an error:

```
--- FAIL: TestCreateAndRefresh-2 (0.16s)
	Location:	cfr_test.go:47
	Error:		No error is expected but got Processing your request will cause you to exceed the maximum number of distributions allowed.
	Messages:	Should be able to create distribution
		
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x465da0]
```

It seems like it's related to Cloudflare, do you think it has to do something with your PR?

myleshorton:
That error was actually the reason this ticket/PR was created @xiam. We chatted about this over email, but the purpose of this change is to not create so many CloudFront distributions so we don't hit that limit.

aranhoide:
To alleviate this in the meanwhile, I deleted a bunch of distributions we weren't using.

uaalto:
Ok, I couldn't find any reference to Cloudfront code that could be affected by these changes.
https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/peerscanner/main.go#L279 could be a problem, but this should never be true now: https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/peerscanner/host.go#L575
LGTM!

PR #2752, opened 2015-07-16
Bumped up timeout for checks closes #2743

### LINKED ISSUE #2743: Timeouts seems too low for some of our checks

myleshorton:
This just bumps up the timeouts for checks after failures as the existing timeouts otherwise look OK.

### DIFF
--- a/src/github.com/getlantern/balancer/dialer.go
+++ b/src/github.com/getlantern/balancer/dialer.go
@@ -131,7 +131,7 @@ func (d *dialer) defaultCheck() bool {
 			Dial: d.Dial,
 		},
 	}
-	ok, timedOut, _ := withtimeout.Do(10*time.Second, func() (interface{}, error) {
+	ok, timedOut, _ := withtimeout.Do(60*time.Second, func() (interface{}, error) {
 		resp, err := client.Get("http://www.google.com/humans.txt")
 		if err != nil {
 			log.Debugf("Error testing dialer %s to humans.txt: %s", d.Label, err)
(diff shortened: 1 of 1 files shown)


src/github.com/getlantern/balancer/dialer.go
```diff
@@ -131,7 +131,7 @@ func (d *dialer) defaultCheck() bool {
 			Dial: d.Dial,
 		},
 	}
-	ok, timedOut, _ := withtimeout.Do(10*time.Second, func() (interface{}, error) {
+	ok, timedOut, _ := withtimeout.Do(60*time.Second, func() (interface{}, error) {
```
xiam: Now that we are changing this, what do you think of making it a local constant instead of a hardcoded value? Maybe `proxyHealthTimeout` or `defaultCheckTimeout` instead of the hardcoded value that depends on the context to be understood. I think it will be easier to read and to change.

myleshorton: In general I actually think it's better to only make things constants only if they're used multiple times. In theory making something a constant makes it more descriptive, but to me there's nothing more self-documenting than where something like this is actually used in the code itself. As an illustrations of that, if this were a constant, I can't imagine ever changing it in the future without also looking at this section of code that actually uses it, in which case it just creates more indirection, would take more time, and IMHO is actually less clear and understandable.

That's my view on it, although honestly I actually don't think it's very consequential either way and actually don't care much despite the long description above!

xiam: Thanks @myleshorton. Otherwise the PR looks good to me, let's keep going and discuss this cosmetic detail any other day.

myleshorton: Sounds good @xiam thanks for the review!

Issue #2760, opened 2015-07-18
Encounter tag errors when build binaries from tag 2.0.0-beta8


yuan78:
Encounter tag errors when build binaries from tag 2.0.0-beta8:

make binaries
......
Step 35 : WORKDIR $WORKDIR
 ---> Using cache
 ---> 2bcaf8d1c7a5
Successfully built 2bcaf8d1c7a5
Generating assets...
date: invalid option -- 'j'
Try 'date --help' for more information.
date: invalid date 'tag 2.0.0-beta8 Tagger: Myles Horton <myles@getlantern.org>  Tagged 2015-07-17 20:50:34 -0700'
Icon  installer-resources/windows/lantern.ico  ID:  5
OK
Building linux/386...
date: invalid option -- 'j'
Try 'date --help' for more information.
date: invalid date 'tag 2.0.0-beta8 Tagger: Myles Horton <myles@getlantern.org>  Tagged 2015-07-17 20:50:34 -0700'
** VERSION was not set, using git revision instead (2.0.0-beta8). This is OK while in development.
Build tags:
# github.com/getlantern/flashlight
usage: 8l [options] main.8
  -1    use alternate profiling code
  -B info
        define ELF NT_GNU_BUILD_ID note
  -C    check Go calls to C code
  -D addr
        data address
  -E sym
        entry symbol
  -I interp
        set ELF interp
  -L dir
        add dir to library path
  -H head
        header type
  -K    add stack underflow checks
  -O    print pc-line tables
  -Q    debug byte-register code gen
  -R rnd
        address rounding
  -S    check type signatures
  -T addr
        text address
  -V    print version and exit
  -W    disassemble input
  -X name value
        define string data
  -Z    clear stack frame on entry
  -a    disassemble output
  -c    dump call graph
  -d    disable dynamic executable
  -extld ld
        linker to run in external mode
  -extldflags ldflags
        flags for external linker
  -f    ignore version mismatch
  -g    disable go package data checks
  -installsuffix suffix
        pkg directory suffix
  -k sym
        set field tracking symbol
  -linkmode mode
        set link mode (internal, external, auto)
  -n    dump symbol table
  -o outfile
        set output file
  -r dir1:dir2:...
        set ELF dynamic linker search path
  -race
        enable race detector
  -s    disable symbol table
  -tmpdir dir
        leave temporary files in this directory
  -u    reject unsafe packages
  -v    print link trace
  -w    disable DWARF generation
[...]

PR #2761, opened 2015-07-18
Fixing REVISION_DATE on Linux.

xiam:
Our Makefile was failing on Linux, see: https://github.com/getlantern/lantern/issues/2760


myleshorton:
LGTM.

PR #2766, opened 2015-07-19
Fix "Close Window" button

atavism:
This fixes the "Close Window" button that appears in the UI when the Lantern backend goes away.


myleshorton:
Nice! Working for me. Thanks.

Issue #2721, opened 2015-07-12
Typing an website address in the address bar of a Lantern browser returns to Lantern itself


sheena3:
I tried to open a website by typing it in the address bar in the Lantern browser. It just takes me back to http://127.0.0.1:16823/ . Is that how it is designed to work? I expected it to open the website that I wanted to visit.

sheena3:
This happens only in IE and FF.

uaalto:
What do you mean by _Lantern browser_? Do you mean that you cannot navigate to any site since the browser (IE and FF) will always redirect you to http://127.0.0.1:16823/?

uaalto:
Also, which version of Windows it is?

sheena3:
I mean the Lantern page http://127.0.0.1:16823/ and not the browser itself. I have so many tabs in the browser including the Lantern page. So, in the lantern page alone if I type any other address then it redirects me to http://127.0.0.1:16823/. In all the other tabs of the browser, it works fine.

sheena3:
I am using windows 8.1. IE 11 and FF 39.0

uaalto:
Thank you @sheena3, I'm going to try and reproduce it.

uaalto:
I can't reproduce this one. I've tried with Windows 8.1 and IE 11/FF 39.
One question, are you using a VM for testing this changes or your normal daily-use environment?

uaalto:
Windows weirdness... My test system updated and then I could reproduce the issue. But now, after I installed this one:
https://github.com/getlantern/lantern/issues/2719
I'm not able to reproduce it anymore. I'm wary of calling it a fix for this issue, but could you please check if you still get this behavior?

sheena3:
@uaalto yes, I too couldn't reproduce this in IE 11 now but FF remains the same. And I am not using VM for windows 8.1.

uaalto:
Thanks @sheena3. This is a weird error.
I indeed could reproduce it once more in FF, but now I can't after +5 tries. This error doesn't sound like something that the browser could be caching, but still it might be just gone after https://github.com/getlantern/lantern/issues/2719. In any case, I think it is related to this one: https://github.com/getlantern/lantern/issues/2741, because I got that error (exactly once) while playing with this one.
But still the question remains, why would Lantern redirect to localhost:lantern-port then?

uaalto:
[...]

Issue #2429, opened 2015-04-22
Flashlight client errors responding with OK: Unable to respond OK: WSASend tcp 127.0.0.1:8787: An existing connection was forcibly closed by the remote host.


myleshorton:
Appearing in loggly:

```
flashlight.client: Unable to respond OK: WSASend tcp 127.0.0.1:8787: An existing connection was forcibly closed by the remote host.
```

xiam:
Got a bunch of those messages too:

Apr 27 21:14:54.079 - flashlight.client: handler.go:102 Unable to respond OK: write tcp 127.0.0.1:61555: broken pipe
Apr 27 21:14:55.574 - flashlight.client: handler.go:102 Unable to respond OK: write tcp 127.0.0.1:61648: broken pipe
Apr 27 21:14:56.262 - flashlight.client: handler.go:102 Unable to respond OK: write tcp 127.0.0.1:61739: broken pipe
Apr 27 21:14:56.721 - flashlight.client: handler.go:102 Unable to respond OK: write tcp 127.0.0.1:61807: broken pipe
Apr 27 21:14:57.042 - flashlight.client: handler.go:102 Unable to respond OK: write tcp 127.0.0.1:61831: broken pipe
Apr 27 21:14:57.846 - flashlight.client: handler.go:102 Unable to respond OK: write tcp 127.0.0.1:61865: protocol wrong type for socket
Apr 27 21:14:58.003 - flashlight.client: handler.go:102 Unable to respond OK: write tcp 127.0.0.1:61887: broken pipe
Apr 27 21:14:58.384 - flashlight.client: handler.go:102 Unable to respond OK: write tcp 127.0.0.1:61896: broken pipe
Apr 27 21:15:26.995 - flashlight.client: handler.go:102 Unable to respond OK: write tcp 127.0.0.1:62089: broken pipe

uaalto:
Appear related to this issue: https://github.com/getlantern/lantern/issues/2721
Better log now, we have the line of code: 
`flashlight.client: handler.go:116`

uaalto:
xiam's log also shows a line number, probably the same after some edits.

uaalto:
_An existing connection was forcibly closed by the remote host._ is Windows parlance for _broken pipe_.

uaalto:
Should be fixed by https://github.com/getlantern/lantern/pull/2794

xiam:
I'm still having some logs on 2.0.0:

```
[...]

Issue #2741, opened 2015-07-15
IE sometimes gets into completely broken state


myleshorton:
Related to #2719. It always shows "This page can't be displayed".

When in this state, if you change the system proxy settings to not use the PAC file, it works for unblocked sites.

uaalto:
I could reproduce it just once, after +10 tries. The very nature of this error leads me to believe it is related to the race condition you've been working on @myleshorton. It also makes me think it is related to https://github.com/getlantern/lantern/issues/2721, due to its similar behavior and reproducibility.

uaalto:
I could reproduce it just once, after +10 tries. The very nature of this error leads me to believe it is related to the race condition you've been working on @myleshorton. It also makes me think it is related to https://github.com/getlantern/lantern/issues/2721, due to its similar behavior and reproducibility.

myleshorton:
I agree @uaalto -- I think #2721 could actually be a great clue. I'm going to try to reproduce that.

myleshorton:
Hey @fffw and @uaalto have either of you been checking your logs while testing this? I'm curious if some of these issues like these ```GetAddrInfoW``` errors might be related. Can you try just grepping for ERROR in your log if you're able to get this again (I haven't been able to get into this state since my changes to the startup race stuff).

uaalto:
Nope, sorry. I could reproduce all these glitches very few times, and it became especially hard after the latest changes. Among the initialization issues that we've been having, particularly this one has occurred just once or twice in my case.
What I was trying to figure out in my latest investigation is the defect that redirects the user to the same lantern home page when they write any arbitrary address in that tab. I want to see what's the exact signal that the browser receives from Lantern to produce such result. As weird as it seems, it appears to be a reload of the page, or a redirect, but I don't think this is the reason.

Anyway, this other issue I'm talking about may or may not be related to the issue in this thread. I'll do a bunch of runs now and try to get a log.

uaalto:
Has any of you been able to reproduce it after https://github.com/getlantern/lantern/issues/2719 closed?

uaalto:
15 tries, 2 reboots. I would close this as _not reproducible_. I've used the version with @myleshorton changes and my latest additions too.

myleshorton:
I know @fffw was able to get into this state even after my changes. Are you still able to reproduce it @fffw? Maybe it has something to do with actually running in a censored region?

fffw:
[...]

PR #2769, opened 2015-07-19
Issue 2429

uaalto:
[This issue](https://github.com/getlantern/lantern/issues/2429) was identified as part of the ongoing investigation of issue [2721](https://github.com/getlantern/lantern/issues/2721).

The most important thing to review and discuss here, in my opinion, concerns the new handling of the bidirectional piping client<->proxy, which now will make sure that a) starts after the succesful OK response and b) closes both directions just once, if any pipe exits. At the moment the closing is done only by the client->proxy direction, which causes the function to unblock with ongoing piping in the other direction.

It also handles all errors in flashlight/client package.


uaalto:
This code is being improved for issue #2721. Should I remove this pull request altogether or keep them separate?

myleshorton:
I like this change @uaalto. One comment coming about maybe skipping the deferred close calls in the intercept method entirely...one sec.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -69,7 +73,11 @@ func (client *Client) intercept(resp http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	defer connOut.Close()
+	defer func() {
+		if err := connOut.Close(); err != nil {
+			log.Debugf("Error closing the out connection: %s", err)
+		}
+	}()
```
myleshorton: OK, what about skipping both of these deferred close calls and instead defining your closeConns call here and both passing it to pipeData and deferring it here? That would eliminate a tad bit of duplicate code on the closes and I think would accomplish the same thing.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -93,35 +101,50 @@ func (client *Client) targetQOS(req *http.Request) int {
...
-		// A bit arbitrary, but it's rather rare now to use half closing
-		// as a way to notify server. Most application closes both connections
-		// after completed send / receive so that won't cause problem.
-		wg.Wait()
-		clientConn.Close()
-		connOut.Close()
-	}()
-
 	// Respond OK
 	err := respondOK(clientConn, req)
```
myleshorton: Just doing this at the beginning is far better -- don't know why I didn't think of that thanks.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -49,7 +49,11 @@ func (client *Client) intercept(resp http.ResponseWriter, req *http.Request) {
 		respondBadGateway(resp, fmt.Sprintf("Unable to hijack connection: %s", err))
 		return
```
myleshorton: This isn't related to your change @uaalto, but don't we need to close clientConn here?

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -93,35 +101,50 @@ func (client *Client) targetQOS(req *http.Request) int {
...
+		}
+		closeOnce.Do(closeConns)
+	}()
+
 	// Then start coyping from proxy to client
-	io.Copy(clientConn, connOut)
+	if _, err := io.Copy(clientConn, connOut); err != nil {
+		log.Debugf("Error piping data from proxy to client: %s", err)
+	}
+	closeOnce.Do(closeConns)
```
uaalto: One comment about this line too. After testing it with extra error handling I realized we were closing twice. This code would take care of that (pending in another branch):

```

// pipeData pipes data between the client and proxy connections.
func pipeData(clientConn net.Conn, connOut net.Conn, req *http.Request) {
	// Wait for both directions to be done
	done := make(chan bool)

	go func() {
		if _, err := io.Copy(connOut, clientConn); err != nil {
			log.Debugf("Error piping data from client to proxy (%v): %s", err, connOut)
		}
		done <- true
	}()

	// Then start copying from proxy to client
	if _, err := io.Copy(clientConn, connOut); err != nil {
		log.Debugf("Error piping data from proxy to client (%v): %s", err, connOut)
	}
	<-done
}
```

src/github.com/getlantern/flashlight/client/handler.go
uaalto: Sure! I'll come back to this code ASAP. I'm working on #2721, which is crazy.

src/github.com/getlantern/flashlight/client/handler.go
myleshorton: OK cool. I *think* though that we don't want to always wait for both sides to be done because one side being done is one of the main ways the other side knows to finish, i.e. we have to essentially propagate the close to the other side through actually closing that other side. Otherwise I think we could end up with the too many open files issue -- this pipeData code has been rewritten a couple of times to prevent that if you look at the history.

uaalto: That's perfectly possible. I'm touching many things in order to figure out the issue #2721, but I'll revisit this before making a case. I _believe_ I have a real clue of what's going on with that issue.

src/github.com/getlantern/flashlight/client/handler.go
myleshorton: OK awesome! I think #2721 is probably more important and hopefully could lead to or even be a fix for #2741, as I think you already mentioned.

src/github.com/getlantern/flashlight/client/handler.go
myleshorton: Exciting stuff!

src/github.com/getlantern/flashlight/client/handler.go
uaalto: I've checked what you mention in the parent comment, and I think it was originally written like this so **intercept** already handles closing connections even if **pipeData** is never reached, due to a **BadGateway** error (i.e. any return statement before reaching **pipeData**). Is this correct?

src/github.com/getlantern/flashlight/client/handler.go
uaalto: Yes I think so too. I'm prepared a commit to handle all the comments you've done in this thread.

uaalto:
This last commit proposes a way of making sure both halves are closed just once, but still being able to preemptively close the proxy->client direction if the client->proxy fails/finishes. Also makes sure connections are closed in case of error/BadGateway.

src/github.com/getlantern/flashlight/client/handler.go
myleshorton: I think that's right yeah but I would just add the closing method to intercept, so something like:

```
 44     var err error
 45
 46     // Hijack underlying connection.
 47     var clientConn net.Conn
 48
 49     var connOut net.Conn
 50
 51     // Make sure of closing connections only once
 52     var closeOnce sync.Once
 53
 54     // Force closing if EOF at the request half or error encountered.
 55     // A bit arbitrary, but it's rather rare now to use half closing
 56     // as a way to notify server. Most application closes both connections
 57     // after completed send / receive so that won't cause problem.
 58     closeConns := func() {
 59         if clientConn != nil && err := clientConn.Close(); err != nil {
 60             log.Debugf("Error closing the out connection: %s", err)
 61         }
 62         if connOut != nil && err := connOut.Close(); err != nil {
 63             log.Debugf("Error closing the client connection: %s", err)
 64         }
 65     }
 66
 67     defer closeOnce.Do(closeConns)
 68
 69     if clientConn, _, err = resp.(http.Hijacker).Hijack(); err != nil {
 70         respondBadGateway(resp, fmt.Sprintf("Unable to hijack connection: %s", err))
 71         return
 72     }
 73
 74     addr := hostIncludingPort(req, 443)
 75     // Establish outbound connection.
 76     d := func(network, addr string) (net.Conn, error) {
 77         return client.getBalancer().DialQOS("tcp", addr, client.targetQOS(req))
 78     }
 79
 80     if runtime.GOOS == "android" || client.ProxyAll {
 81         connOut, err = d("tcp", addr)
 82     } else {
 83         connOut, err = detour.Dialer(d)("tcp", addr)
 84     }
 85
 86     if err != nil {
 87         respondBadGateway(clientConn, fmt.Sprintf("Unable to handle CONNECT request: %s", err))
 88         return
 89     }
 90
 91     // Pipe data between the client and the proxy.
 92     pipeData(clientConn, connOut, req, closeConns)
```

myleshorton: Note the check for nils on the conns..

myleshorton: and passing closeConns to pipeData..

uaalto: I think your solution for keeping the code that closes both halves together is cleaner and easier to maintain. However, I don't see why you want to pass the **closeConns** function to **pipeData**, because you'll end up running that code twice (in **pipeData** and as a deferred function of **intercept**).
I've committed a version that I think cleanly handles running the closing code just once. I think it could be improved with your code closing both halves in the same function.

myleshorton: Because you still need to call it in this block:

```
121     // Start piping from client to proxy                                                                         | 15     Label string
122     go func() {                                                                                                  | 16
123         if _, err := io.Copy(connOut, clientConn); err != nil {                                                  | 17     // Weight: determines how often this Dialer is used relative to the other
124             log.Debugf("Error piping data from client to proxy: %s", err)                                        | 18     // Dialers on the balancer.
125         }                                                                                                        | 19     Weight int
126         closeOnce.Do(closeConns)                                                                                 | 20
127     }()
```

It actually should pass the full closeOnce.Do(closeConns) as a func though, so something like:

```
pipeData(clientConn, connOut, req, func() {closeOnce.Do(closeConns)})
```

Does that make sense or am I missing something? I think you still need to call that from the go routine.

myleshorton: Replacing ```closeOnce.Do(closeConns) ``` with the func call of course, so something like ```closeFunc()``` instead.

uaalto: Yes, that's a solution (passing the sync.once variable). The challenge here is to be able to call the function within the goroutine. What I proposed in the latest commit is to use channel blocking to achieve the same goal without `sync.once` or directly calling, but relying on the defer. Please, let me know what you think of this other possibility :)
https://github.com/getlantern/lantern/blob/6f9e63687a6bf16ff7099f93fe4fae30d4dddc4f/src/github.com/getlantern/flashlight/client/handler.go#L119

uaalto: Oh, by the way, I took the OK response out of the piping function to keep this last one doing just that, but that's a different topic, and I can of course put it back :)

myleshorton: Right -- I'm not suggesting passing the sync.once variable but rather a function that calls it (again ```pipeData(clientConn, connOut, req, func() {closeOnce.Do(closeConns)})```), so that would preserve the sync.once intention. 

Your changes look OK, except that we need to call the closing function if either copy method returns. I believe your change will only reach the deferred func in intercept when the client->proxy copy ends and signal <- true is called.

uaalto: Yes, that is what I understood from the comment that was in the code: the client->proxy half should be able to close both sides if finished/error, but not the other way around.
If we want both sides to be able to close the other one preemptively, then I think is best done with `once`, indeed!

myleshorton:
As far as I can tell the current code still relies on the ```signal <- true``` executing for the connections to get closed - https://github.com/getlantern/lantern/blob/issue-2429/src/github.com/getlantern/flashlight/client/handler.go#L119

Without that, intercept will block until it gets something on that channel, so none of the deferred will get called. Previously intercept itself also handled those closings in deferred calls, but it didn't block at the end so was OK. For reference:

https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/client/handler.go

uaalto:
Yes @myleshorton, this comment and commit was previous to our discussion on this code. I'll update it to reflect it.

uaalto:
Just to make sure: Do you want me to keep the old behavior of closing both connections if the client->proxy ends, as in https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/client/handler.go#L107?
Or should I make both sides close each other as you mention here https://github.com/getlantern/lantern/pull/2769/files#r35149539?

myleshorton:
So the old code *did close both both connections in both cases* -- it just did it differently through falling back to the deferred methods in intercept when the pipeData method ended. pipeData would only end when [io.Copy(clientConn, connOut)](https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/client/handler.go#L120) completed, which would return the flow of control to intercept, which would then close both of the connections.

So, yes, we want to keep the behavior where either Copy call ending triggers closing of all the connections. Make sense?

myleshorton:
> Do you want me to keep the old behavior of closing both connections only if the client->proxy ends

So just to be super clear, the above statement is incorrect as far as I can tell -- i.e. that was not the old behavior.

uaalto:
Yes, indeed. I was completely mistaken! What I don't understand is the comment:
https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/client/handler.go#L102
```
		// Force closing if EOF at the request half or error encountered.
		// A bit arbitrary, but it's rather rare now to use half closing
		// as a way to notify server. Most application closes both connections
		// after completed send / receive so that won't cause problem.
```
It got stuck in my mind that the request half forced the other to exit as soon as finish/error happened. What does this comment refer to? (I want to understand it fully so there is no detail left behind).

uaalto:
Does the _request half_ refer to both directions from the (browser) client to Lantern at the same time? If that's the case, then everything is clear to me.
What I understood before is _half_ as in _duplex_ connection direction. Thus the confusion.

fffw:
@uaalto The comment you mentioned is introduced in https://github.com/getlantern/lantern/commit/4ce2d7a5f3cf03deb2e1af5db952e38760007892 as an attempt to fix the "too many open files" issue. In fact, it's already out of date. We can remove it to avoid confusion.

myleshorton:
Gotcha thanks @uaalto and @fffw. I agree just removing that comment probably makes sense.

uaalto:
If this code looks right, we can just merge https://github.com/getlantern/lantern/pull/2794 and close this pull request. I'm sorry for these two pull requests stepping on each other, I just found one bug on my research for the other.

uaalto:
Or I can do a rebase and remove some commits.

myleshorton:
Oh no problem @uaalto -- let's just use the other one.

PR #2775, opened 2015-07-20
Separate genconfig so we commit the changes for all platforms to pick them up

myleshorton:


myleshorton:
Hey this is a slight change from before @atavism, @xiam, and @fffw in particular. I think we ideally should do both genassets and genconfig outside of ```make packages``` because we really should commit those changes for all platforms to pick them up and also so they're actually in the repo.

xiam:
@myleshorton I agree, we should do this periodically but I'm not sure what would be the best way to enforce this update without keeping it here. Maybe we can just simply take this as part of the support routine?

myleshorton:
Yeah that's an interesting idea, @xiam. I was planning on just doing it manually each time and committing the changes. The problem with the way it is now is that as far as I can tell it only affects the OS you're building on, i.e. OSX. It won't do anything for the config on other platforms because the changes aren't actually committed.

xiam:
That's right @myleshorton. I guess it would be better just to do it by hand from time to time and make sure we actually commit the modified files.

PR #2779, opened 2015-07-21
Fix UI page screen resolution issue

atavism:


src/github.com/getlantern/lantern-ui/app/_css/app.css
```diff
@@ -7602,7 +7602,7 @@ body[dir=rtl] #latestDetails dd {
 #vis-container, #vis {
   height: 100%;
   background-color: #fff;
-  overflow: scroll;
+  overflow: hidden;
```
fffw: It's not clear to me what this change for. User may not see the full content via a not maximized window, which is very common.

We can reduce the `margin-bottom` or `padding-bottom` of `#sites` to eliminate extra spaces.

fffw:
LGTM except what I noted before.

It would be better to run `UPDATE_DIST=true make genassets` and add `dist` folder to repo so others don't need to do the same.

Issue #2780, opened 2015-07-21
Check for LAN hosts and non-standard ports in PAC file


fffw:
It's useless to proxy them. Additionally, they creates lots of errors and confuses user.

PR #2781, opened 2015-07-21
not proxy traffic to LAN hosts

fffw:
Resolves https://github.com/getlantern/lantern/issues/2780

To exam ports, we have to match url with regex, which is error prone and [never perfect](https://mathiasbynens.be/demo/url-regex). It's rare to access non-standard ports on Internet, so I just skip it.

### DIFF
--- a/src/github.com/getlantern/flashlight/pac.go
+++ b/src/github.com/getlantern/flashlight/pac.go
@@ -80,8 +80,18 @@ func genPACFile() {
 	formatter :=
 		`var bypassDomains = %s;
 		function FindProxyForURL(url, host) {
-			if (host == "localhost" || host == "127.0.0.1") {
+			if (isPlainHostName(host) // including localhost
+			|| shExpMatch(host, "*.local")) {
 				return "DIRECT";
+			}
+			// only checks plain IP addresses to avoid leaking domain name
+			if (/^[0-9.]+$/.test(host)) {
+				if (isInNet(host, "10.0.0.0", "255.0.0.0") ||
+				isInNet(host, "172.16.0.0",  "255.240.0.0") ||
+				isInNet(host, "192.168.0.0",  "255.255.0.0") ||
+				isInNet(host, "127.0.0.0", "255.255.255.0")) {
+					return "DIRECT";
+				}
 			}
 			for (var d in bypassDomains) {
 				if (host == bypassDomains[d]) {


src/github.com/getlantern/flashlight/pac.go
```diff
@@ -80,9 +80,19 @@ func genPACFile() {
 	formatter :=
 		`var bypassDomains = %s;
 		function FindProxyForURL(url, host) {
-			if (host == "localhost" || host == "127.0.0.1") {
+			if (isPlainHostName(host) // including localhost
```
myleshorton: I think you mean ```!isPlainHostName(host)``` right @fffw?

fffw: No. See http://findproxyforurl.com/pac-functions/#tcompare

> This function will return true if the hostname contains no dots, e.g. http://intranet

myleshorton: Oh I see -- they're primarily referring to dots as in "internet.com" not as in "1.1.1.1" -- looks good.

src/github.com/getlantern/flashlight/pac.go
```diff
@@ -80,9 +80,19 @@ func genPACFile() {
...
 			}
+			// only checks plain IP addresses to avoid leaking domain name
+			if (/^[0-9.]+$/.test(host)) {
+				if (isInNet(host, "10.0.0.0", "255.0.0.0") ||
+				isInNet(host, "172.16.0.0",  "255.240.0.0") ||
+				isInNet(host, "192.168.0.0",  "255.255.0.0") ||
+				isInNet(host, "127.0.0.0", "255.255.255.0")) {
+					return "DIRECT";
+				}
+			}
```
myleshorton: Why wouldn't we want to go directly to any ip address @fffw? Why limit it at all?

fffw: Those are private IP addresses reserved to LAN, which is definitely not reachable from Lantern server, so we can always access them directly. But user may access an public IP address which is blocked. We should allow this right?

myleshorton: I guess I'm just thinking it's pretty unlikely users would ever enter those IPs, but I guess you're right that we might as well allow it in the off chance they do try them.

OK, merging!

fffw: Thanks @myleshorton . I just realized that some web pages can embed a few resources via IP address. The check seems meaningful.

Issue #2782, opened 2015-07-21
[ui] Layout messed up on IE 10


fffw:
![image](https://cloud.githubusercontent.com/assets/10410437/8801141/e7d6b0da-2fea-11e5-8392-68e3617a0a4e.png)

1. Logo is too large, pushes text down.
2. Share button too large.
3. Share text misplaced.

myleshorton:
Wow crazy. We should fix that =). I think we should presize all of these graphics to their actual size for the most part and avoid resizing in the UI.

PR #2783, opened 2015-07-21
make IE10 happy with image size

fffw:
This makes IE10 renders almost the same as other browser, except the footer text, which is not vertical centered and makes the footer taller. Don't merge until we find a way to fix that.

Resolves https://github.com/getlantern/lantern/issues/2782

image


myleshorton:
Working for me @fffw thanks!

Issue #2703, opened 2015-07-09
tlsdialer test code panics


fffw:
See https://travis-ci.org/getlantern/lantern/builds/70156151, it's an unexpected `x509.UnknownAuthorityError` during dialing to local test address. It's the test code triggers the panic, not production code.

--- FAIL: TestVariableTimeouts-2 (2.80s)
panic: interface conversion: x509.UnknownAuthorityError is not net.Error: missing method Temporary [recovered]
	panic: interface conversion: x509.UnknownAuthorityError is not net.Error: missing method Temporary
goroutine 29 [running]:
testing.func·006()
	/usr/local/go/src/testing/testing.go:441 +0x181
github.com/getlantern/tlsdialer.doTestTimeout(0xc208138000, 0x3ea710)
	/home/travis/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/tlsdialer/tlsdialer_test.go:219 +0x283
github.com/getlantern/tlsdialer.TestVariableTimeouts(0xc208138000)
	/home/travis/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/tlsdialer/tlsdialer_test.go:205 +0x17d

myleshorton:
Yeah this is happening way to often and is very annoying. Let's fix it.

PR #2789, opened 2015-07-22
Fix for 2703

uaalto:
TLSDialer test: make sure the error is of type net.Error before checking whether is a timeout error

https://github.com/getlantern/lantern/issues/2703


myleshorton:
LGTM thanks @uaalto!

PR #2790, opened 2015-07-22
[READY] Catch-all handler for panics that logs to loggly and runs cleanup functions

### LINKED ISSUE #2648: Add a "catch all" handler in case Lantern panics
We can try to log the panic, and reset PAC setting if possible.

uaalto:
Closes #2648 
Adds a "catch all" handler in case Lantern panics, which will log and exit (taking care of calling clean up functions like PAC uninstalling).


fffw:
This `defer` can't catch panics raised from other goroutines (I tested). If we want to do so, we have to use some trick described [here](https://groups.google.com/d/msg/golang-nuts/3QRnU0xJVXg/7NKMjeV5m-UJ) which is a bit ugly.

I think panic happens rather rarely. A good reason to implement it is if we can do both:
1. Clean up or restart itself;
2. Reliably report panic to Loggly or somewhere else for investigation.

uaalto:
Oh, I see.  That was a test I didn't do. Thanks Merlin.
I was focusing on the second point you mention, since the first one implies a bigger change and I don't know if that's a desirable behavior.
At the moment, with the simple code I introduced it does report to loggly through _golog_ package, but of course it requires handling panic in all goroutines. I can try and test the code you pointed at, that sound good?

fffw:
Current way of error reporting will not succeed if proxy stops to work, which is very likely to happen when program panics. I would not pay much attention on it unless there's an evidence that Lantern does panic sometimes. We can revisit it when we are sure that errors can be sent reliably in the future. Anyway, it's just my point.

uaalto:
Yes that's true. So, with a panic() the program stops the other threads, however, the recover() only grabs the panic() in the same thread? That means that we have to either send the log without proxying or just establish a new proxy in the recover() function, just for use in this situation. What do you think?

atavism:
@fffw Could we use something like [panicwrap](https://github.com/mitchellh/panicwrap) to make sure we monitor for and forward panics to Loggly?

fffw:
My understanding is that panic() will unwind the stack of current goroutine
until it meets recover(). Other goroutines will not be affected (except if
there's no recover(), the whole process crashes) , but it's not easy to
gracefully end a goroutine without affect the functionality of others.

I think how to reliably report error is worth a separate topic. Currently,
an error can be lost even if program doesn't panic.

uaalto:
It seems to me at this point that introducing a new dependency or spending the time on this is not worth the trouble. I think we might be better off making sure we handle all errors and log them.
The panicwrap library looks interesting though, and probably a good way to go if we want to go on a full-blown version of this.

fffw:
Agreed

uaalto:
I all agree, we can close this PR without merging.

uaalto:
Oh

uaalto:
One more thing before we close this. Would we be interested in handling panic for **uninstalling PAC**, even if we cannot log?
That's one of the things discussed here https://github.com/getlantern/lantern/issues/2648

fffw:
@uaalto There's no report from user that Lantern crashed and caused any problem. I think it's safe to ignore it for now?

uaalto:
@fffw ok! We can close this PR then. We should close the issue as well then.

myleshorton:
I actually think something like @atavism's panicwrap suggestion is a good solution. It's super simple and doesn't impact the code almost at all. To me the big problem now is that panics could be happening, and I would guess are even likely to be happening, but we have no way of knowing other than hoping users contact us.

I would vote for using panicwrap and then attempting to report directly to loggly.

uaalto:
I've done a few tests to see how panicwrap works. The PAC uninstall is not a problem, it works with a panic from any goroutine. Writing logs to a file is not a problem, but I can't make the loggly debug work without larger changes. Anything that requires access to the child process won't work, so we would need to set up loggly and the proxy specifically for this purpose _within_ the parent process.

I don't have a complete sense of the magnitude of this change yet, so I wanted to ask. Or maybe there's another way... What do you think?

myleshorton:
What happens if you just try to use Loggly from the parent process -- i.e. set it up from scratch in the parent process so it has it's own copy?

uaalto:
That's what I mean, yes. Should be proxied as well, is it correct?

myleshorton:
I would just go direct with no proxy in this case @uaalto. It seems like you must be able to reference at least some of the same packages in the outer process no? Otherwise we could just do it completely manually, copying some of the code perhaps?

myleshorton:
Hey looks like the tests are looking for tests in panicrwrap but not finding them @uaalto..

fffw:
An alternative to handle panics in one goroutine without affect others is like https://groups.google.com/forum/#!msg/golang-nuts/3QRnU0xJVXg/7NKMjeV5m-UJ. We change slightly the way to span goroutine and all set.

myleshorton:
Yeah that's interesting @fffw. Just make a public GoRecover function or similar that everything uses. Given the complications surrounding logging, I feel like that might be a better solution. What do you think @uaalto?

uaalto:
The GoRecover function looks like a wrapper that you would need to use to "decorate" each function that you want to run under its recovery mechanism. We could of course run the main() function within GoRecover, but I would need to test that this work in all cases. After the initial findings by @fffw I would say it doesn't work in the general case. Actually, that's the whole reason why paincwrap exists using that esoteric method of reading the stderr of a forked process. Panicwrap is not beautiful, but at least is almost non-intrusive at all.

Also, how should I run the tests project-wide to see errors in this particular package? (I thought only tests under testpackages.txt were run)

uaalto:
I did test this and I think it won't fix our problem (unless I'm missing something). The discussion in the thread is not about our own issue.
Check this out:

```
package main

import (
	"fmt"
	"time"
)

func goSilent(f func()) {
	go func() {
		defer func() {
			if p := recover(); p != nil {
				fmt.Println("Panic silenced:", p)
			}
		}()
		f()
	}()
}

func main() {
	goSilent(func() {
		go func() {
			panic("this is NOT recovered")
		}()
		panic("this is recovered")
	})
	time.Sleep(1000 * 1000 * 1000)
}

```

uaalto:
The question for me would be: should we go full-blown on panic logging? I would say yes, since we invested time in this already (and we actually don't know how many people are getting panics).

[This](https://github.com/getlantern/lantern/issues/2807) will help also in this regard, but adding the complete panicwrap-based implementation will add an extra layer of diagnosis that will ensure *liveness*.

The only question remaining is how big this change would be, which I'm not sure at the moment. But I will work on it if you see it worth as well.

So in summary, I've changed my position and I believe we could go for this as an extra tool.

myleshorton:
So the only way the the goSilent approach would work is if it's used every single time we create a go routine. We use some external libraries that we can't control, but not many. Is that why you're saying it wouldn't work @uaalto? Because of external libraries?

uaalto:
Yes, the goSilent approach would work only if used every single time we create a goroutine. I think that approach is verbose and error-prone, that's what I mean. I think using panicwrap is the best solution.

fffw:
Totally agree with @uaalto . Out of band way is more robust.

myleshorton:
panicwrap works for me guys!

uaalto:
Ok, I have a working proof of concept of a panic catcher that cleans up and logs to Loggly through a proxy. I now have some questions in order to tweak the code for the final version. Please see the comments in the code!

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -67,20 +67,23 @@ func Init() error {
...
 
 func Configure(addr string, cloudConfigCA string, instanceId string,
-	version string, revisionDate string) {
+	version string, revisionDate string) (done chan bool) {
 	if logglyToken == "" {
-		log.Debugf("No logglyToken, not sending error logs to Loggly")
-		return
+		//log.Debugf("No logglyToken, not sending error logs to Loggly")
+		logglyToken = "testLogglyToken"
+		//return
```
uaalto: Ok, sorry for this leftover. I'm using my own "local loggly server" (just an `nc` server) so I had this overwritten for testing. Will revert in final commit.

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -90,10 +93,14 @@ func Configure(addr string, cloudConfigCA string, instanceId string,
 
 	// Using a goroutine because we'll be using waitforserver and at this time
 	// the proxy is not yet ready.
+	done = make(chan bool, 1)
```
uaalto: I did this in order to provide optional blocking on the library client side. It doesn't change the API so it won't affect existing code.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -82,7 +84,54 @@ func init() {
...
+	time.Sleep(6 * time.Second)
+	logging.Close()
+}
+
 func main() {
+	// panicwrap works by re-executing the running program (retaining arguments,
+	// environmental variables, etc.) and monitoring the stderr of the program.
+	exitStatus, err := panicwrap.BasicWrap(
+		func(output string) {
+			logPanic("127.0.0.1:18787", output)
```
uaalto: I think we should use a different port than the user-defined one for Lantern regular use, because in case we had a panic the port might be still unusable. The way to define this probably needs some thought.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -82,7 +84,54 @@ func init() {
...
+	if err := logging.Init(); err != nil {
+		panic("Error initializing logging")
+	}
+
+	ready = logging.Configure(proxyAddr, cfg.CloudConfigCA, cfg.InstanceId, version, revisionDate)
+	<-ready
+
+	log.Error(msg)
+
+	time.Sleep(6 * time.Second)
```
uaalto: This is awful in my opinion. It turns out that the Loggly logger only flushes every 5 seconds. There is no exposed interface for manually flushing ATM. I think we should expose it and avoid this artificial wait, what do you think @myleshorton @fffw ?

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -82,7 +84,54 @@ func init() {
...
+	// Create the client-side proxy.
+	c := &client.Client{
+		// TODO: Next port in case the regular one hasn't been closed properly
+		Addr:         proxyAddr,
+		ReadTimeout:  0, // don't timeout
+		WriteTimeout: 0,
+	}
+	ready := make(chan bool)
+	go c.ListenAndServe(func() { ready <- true })
+	<-ready
```
uaalto: The alternative to these 2 `ready` blocking signals (or similar mechanism) is to wait 1 second or so.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -82,7 +84,54 @@ func init() {
...
 
+func logPanic(proxyAddr string, msg string) {
+	cfg, err := config.Init(packageVersion)
+	if err != nil {
+		panic("Error initializing config")
+	}
+
+	// Create the client-side proxy.
+	c := &client.Client{
+		// TODO: Next port in case the regular one hasn't been closed properly
```
uaalto: outdated comment.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -82,7 +84,54 @@ func init() {
...
+	<-ready
+
+	if err := logging.Init(); err != nil {
+		panic("Error initializing logging")
+	}
+
+	ready = logging.Configure(proxyAddr, cfg.CloudConfigCA, cfg.InstanceId, version, revisionDate)
+	<-ready
+
+	log.Error(msg)
```
uaalto: If the message is passed as-is, it will be logged in multiple lines (as multiple messages), because `panic` produces a ton of output. We could do `strings.Replace("\n", "\\n", -1)` to replace newlines with the escaped symbol. Anything better? Some sort of encoding? Replacing with whitespace?

src/github.com/getlantern/flashlight/flashlight.go
fffw: Totally agree to expose `Flush()` interface.

src/github.com/getlantern/flashlight/flashlight.go
fffw: It would be better if we can change logging package a bit to allow direct connection, than to set up the whole thing just for this single purpose. What do you think @uaalto ?

src/github.com/getlantern/flashlight/flashlight.go
uaalto: It requires a few (minor) intermediate changes as well, but I don't think is a problem.

src/github.com/getlantern/flashlight/flashlight.go
uaalto: That's a somewhat tactical decision in my opinion. What is the reason we are tunneling also Loggly logs? Loggly is probably not blocked, that's what I mean.
In that case, the fewer moving parts we have in the panic handler, the better.

fffw: Not absolutely required, but we decided to tunnel all traffic generated by Lantern itself to keep away from potentially being detected.

uaalto: So that's the tradeoff. We can go hard and first try with the proxy. If it fails, do it in the clear. Of course, if the proxy itself is the producer of the panic, we are left out. In any case, this whole thing is a last resort and all errors should be caught before by checks. So if we failed to do so, we better make sure to at least log.
How important is to tunnel loggly messages through our servers?

myleshorton:
What's the status of this one guys. @fffw can you review it? We're seeing some panics in 2.0.0+stable, so it would be really good to get this in!

uaalto:
@myleshorton this on is not yet ready for merging. It just needs a few details, but I'd like to discuss them (they are marked by the comments).

src/github.com/getlantern/flashlight/flashlight.go
fffw: Loggly itself does permit new lines in a message. `golog` permits them too. It's [flashlight/logging](https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/logging/logging.go#L106) that prepends timestamp to each line which is overskill and undesired I think (only affects logfiles and standard outputs). We can return a writer that just prepends timestamp for each `Write` operation there.

src/github.com/getlantern/flashlight/flashlight.go
fffw: Not sure about what other guys thinking, IMPO it's no harm to send direct message to Loggly occasionally.

fffw:
Sorry for leaving your comments unreplied @uaalto

src/github.com/getlantern/flashlight/flashlight.go
uaalto: I've realized that everywhere else we are logging with the prefixes prepended to every line, for all sorts of multiline output. I think we should change that, but in a different PR to avoid mixing concerns. Does that sound right?

fffw: Makes sense to me!

uaalto:
Thank you for your comments @fffw !

uaalto:
OK, I've added on-demand log flushing. I'd like to bring this to a meeting: should we log through proxy? If so, which local port?

src/github.com/getlantern/flashlight/flashlight.go
myleshorton: Yeah I think we should just send the messages directly to Loggly. Panics should be rare enough that we won't be generating much traffic at all, so it won't be conspicuous and will be unlikely to get Loggly blocked. We should also be able to fix panics quickly, which should make them all the more rare. Let's just go direct.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -83,7 +85,53 @@ func init() {
...
+	// environmental variables, etc.) and monitoring the stderr of the program.
+	exitStatus, err := panicwrap.BasicWrap(
+		func(output string) {
+			logPanic("127.0.0.1:18787", output)
+			exit(nil)
+		})
+	if err != nil {
+		// Something went wrong setting up the panic wrapper. This wont' be
+		// captured by panicwrap
+		log.Fatalf("Error setting up panic wrapper: %v", err)
```
myleshorton: I think this should just be an Errorf. If we can't use panicwrap, we definitely want to know that in Loggly, but we also are no worse off than we were before so probably should not exit the program (which itself could well happen before we're able to log the error to Loggly).

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -83,7 +85,53 @@ func init() {
...
+	}
+	// If exitStatus >= 0, then we're the parent process.
+	if exitStatus >= 0 {
+		os.Exit(exitStatus)
```
myleshorton: I'm similarly not sure we should just exit here @uaalto. If we're the parent process, maybe we just log the fact and keep going? What do you think?

src/github.com/getlantern/flashlight/logging/logging.go
myleshorton: I don't understand this change. You're making the last few lines of this non-blocking by default unless the caller chooses to block with the returned channel. The previous code blocked, and your new code also chooses to use the returned channel to block. So the effect of this change is that the [previous call](https://github.com/getlantern/lantern/blob/issue-2648/src/github.com/getlantern/flashlight/flashlight.go#L356) to this does not block, so Loggly will not be configured when it returns. I'm not sure what the effect of that will be, but presumably some future logging calls may not yet have Loggly configured. 

What does this buy us? Configuring Loggly costs basically nothing, so I don't understand why we'd want it to be configured asynchronously.

src/github.com/getlantern/flashlight/flashlight.go
myleshorton: I would just ditch this whole client and pass "" for the proxyAdd.

src/github.com/getlantern/flashlight/flashlight.go
myleshorton: So I think this can all just be the following right?

```go
cfg, err := config.Init(packageVersion)
if err != nil {
	panic("Error initializing config")
}
logging.Configure("", cfg.CloudConfigCA, cfg.InstanceId, version, revisionDate)
log.Error(msg)
logging.Flush()
_ = logging.Close()
```

src/github.com/getlantern/flashlight/logging/logging.go
uaalto: Previously this code didn't block. I did this so I could bock this call without affecting other code.

src/github.com/getlantern/flashlight/flashlight.go
uaalto: Ok. I was thinking on using a different address for proxy instead of the default (this is TODO), since the port might be on use and produce an error when trying to open. However, if we are not going to use Loggly through a proxy, this should just be removed.

src/github.com/getlantern/flashlight/flashlight.go
uaalto: Ok. I'll need to change some logging code, because at the moment it doesn't set the loggly writer if there is no proxy.

src/github.com/getlantern/flashlight/flashlight.go
uaalto: This is how panicwrap works, the parent process exits once it has run the child process "wrapped"
```
// If exitStatus >= 0, then we're the parent process and the panicwrap
// re-executed ourselves and completed. Just exit with the proper status.
```

src/github.com/getlantern/flashlight/logging/logging.go
uaalto: This is the line I mean: https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/logging/logging.go#L96

src/github.com/getlantern/flashlight/flashlight.go
uaalto: We would need that `logging.Configure` blocks. We also would need to call `logging.Init()`, otherwise we get a panic at print using a null pointer.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -83,7 +85,41 @@ func init() {
...
+func logPanic(msg string) {
+	cfg, err := config.Init(packageVersion)
+	if err != nil {
+		panic("Error initializing config")
+	}
+	if err := logging.Init(); err != nil {
+		panic("Error initializing logging")
+	}
+
+	<-logging.Configure("", "", cfg.InstanceId, version, revisionDate)
```
myleshorton: Nice -- didn't know you could do that <- in one line trick!

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -83,7 +85,41 @@ func init() {
...
+
+	log.Error(msg)
```
myleshorton: Actually what will that ```msg``` here typically end up being? I'm wondering if we should prepend it with something we can easily search like "PANIC" or something. We should definitely add panic searches to support.

uaalto: Yes, the output is "panic". The message passed to the `logPanic` function is exactly the output produced by a normal panic. We could just look for `panic` in the output.

myleshorton: Perfect.

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -70,24 +70,28 @@ func Init() error {
...
 
 	if version == "" {
-		log.Error("No version configured, Loggly won't include version information")
+		log.Error("No version configured, not sending error logs to Loggly")
 		return
 	}
 
 	if revisionDate == "" {
-		log.Error("No build date configured, Loggly won't include build date information")
+		log.Error("No build date configured, not sending error logs to Loggly")
```
myleshorton: It seems like we should include ```success <- false``` in all of these cases too right @uaalto?

myleshorton:
Aside from that potential success <- false change this all looks good to me @uaalto!

src/github.com/getlantern/flashlight/logging/logging.go
uaalto: Yes!! This mistake could have caused a lot of pain in the future, I'm sorry. Thank you for catching it on time!!

myleshorton:
LGTM!

PR #2796, opened 2015-07-22
Removing TAG variable. Use VERSION instead. Closes #2795

### LINKED ISSUE #2795: Make TAG in Makefile equivalent to VERSION
Or in other words, remove TAG.

This will help us avoiding replacing releases accidentally.

xiam:
Hey @myleshorton,

I think this can help us to have less noise in the releases creation cycle. Could you take a look?


myleshorton:
Yes I like it thanks @xiam! I'm also going to clean up all those ugly manoto targets in there -- I'll ticket that actually. I'll merge as soon as the tests pass CI.

PR #2801, opened 2015-07-23
Port mobile support to Gomobile

uaalto:
This is finally done. It addresses some topics:

* Ready to support iOS
* Android 5.0 PIE support (depends on Go ultimately), see https://github.com/getlantern/firetweet/issues/142
* Ease of maintenance. When Go 1.5 is released we can even remove the special docker image altogether.
* Everything necessary now by Android is now released as an **aar** package, which is a zip that contains everything necessary for Firetweet or the test app without need for generating and copying binding files around.

IMPORTANT
------------------
These changes will affect what is expected by Firetweet, so it should be merged along [this PR](https://github.com/getlantern/firetweet/pull/144) as well.


Makefile
```diff
@@ -462,21 +463,22 @@ test-and-cover:
 		tail -n +2 profile_tmp.cov >> profile.cov; \
 	done
 
-android-lib: docker-golang-android genconfig
+android-lib-dev: docker-mobile
```
xiam: This `android-lib-dev` target looks a lot like `android-lib`, the only difference I can tell is this line:

```
if [ -d "$(FLASHLIGHT_ANDROID_TESTER_DIR)" ]; then \
```

compared to

```
if [ -d "$(FIRETWEET_DIR)" ]; then \
```

I think we can just leave one `android-lib` target and use environment variables, we'll have to do something like this:

```
FIRETWEET_DIR=../flashlight-android-tester make android-lib
```

for the current `android-lib-dev` or:

```
make android-lib
```

for the default `android-lib` case.

xiam:
Looks good @uaalto, I just have a comment on the Makefile and I think we also need to update the `README.md` accordingly (we are not going to generate a `libgojni.so` anymore, right?).

myleshorton:
Just wanted to say this is exciting stuff @uaalto!

src/github.com/getlantern/lantern-mobile/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+ENV GOPATH /
+
+ENV PATH $PATH:$GOROOT/bin
+ENV WORKDIR /lantern
+
+# Updating system.
+RUN dnf -y update && dnf clean all
+
+# Requisites for building Go.
+RUN dnf install -y git tar patch gzip curl hostname pcre-devel && dnf clean all
```
xiam: While building the image I found some logs like this:

```
./test.bash: line 8: which: command not found
```

Can we add the which command here as well?

xiam:
I tested the new `android-lib` target @uaalto, cool stuff!

I'd just add a note in the docs for the new `.aar` file that replaces our old `libgojni.so`.

```
find . | grep .aar
./src/github.com/getlantern/lantern-mobile/libflashlight.aar
```

Thanks!

Makefile
uaalto: Yes you are right. They converged to the point where they are very similar. The most important difference however is their dependencies. I made android-lib-dev at first to avoid running `genconfig`.

uaalto: So, I still would have two separate tasks, but DRY by one depending on the other. How does this sound?

uaalto:
Thank you!
Yes I have to update the README, good catch!

src/github.com/getlantern/lantern-mobile/Dockerfile
uaalto: I will add this as well, thank you! Tests pass anyway, but it is healthier for the environment.

Makefile
xiam: Yeah, sounds good. Maybe `android-lib` for development and `android-lib-dist` for distribution?

uaalto: Perfect!

src/github.com/getlantern/lantern-mobile/Dockerfile
uaalto: It is not caused by a missing `which` utility, but rather by `which` not finding a file.
I tried to improve this, but I think is not worth fiddling with it: the issue seems to be with their testing scripts. They test the existence of executable which is part of the test suite. Anyway these error recover, and are internal to the test suite.
We can also avoid running the test suite at all.
There were other errors before caused by missing programs, but I already fixed them before the pull request.

xiam:
Looks good @uaalto, but today the target is failing:

```
make android-lib
2015/07/24 16:05:53 Failed: exit status 1
INFO[0244] The command [/bin/sh -c cd $GOROOT/src && ./all.bash] returned a non-zero code: 1
make: *** [docker-mobile] Error 1
```

This was working yesterday, do you think this is due to go1.5 being unstable?

uaalto:
That is a very weird error. I guess "reproducible builds" is just an overstatement. I did run this several times today.
We can just avoid tests and run just make.bash. You can also try removing the Docker image and rebuilding it from scratch. Surprisingly, this works sometimes.

xiam:
Yeah, right now I'm trying `make clean` and removing the docker image. Let's see if that works.

xiam:
Nope, this is still not working,

This error is just above the other one I already posted:

```
--- FAIL: TestCPUProfileMultithreaded (0.26s)
	pprof_test.go:165: total 12 CPU profile samples collected
	pprof_test.go:181: runtime/pprof_test.cpuHog1: 0
	pprof_test.go:181: runtime/pprof_test.cpuHog2: 11
	pprof_test.go:195: runtime/pprof_test.cpuHog1 has 0 samples out of 11, want at least 1, ideally 5
FAIL
FAIL	runtime/pprof	6.538s
--- FAIL: TestTraceStress (1.42s)
	trace_test.go:226: failed to parse trace: p 4 is running before start (offset 613, time 103442316)
--- FAIL: TestTraceFutileWakeup (0.24s)
	trace_test.go:433: failed to parse trace: g 161 is not runnable before start (offset 78284, time 42728926)
FAIL
FAIL	runtime/trace	2.5
```

I'll try to track down this issue...

uaalto:
The `failed to parse trace: g 161 is not runnable before start` issues have appeared sometimes with many different versions of Go I've tried.
It is definitely an issue stemming from the instability of the Go version. But happened with many versions, just sometimes happens and some other doesn't, as far as I can tell from your results.

xiam:
After a few tries it worked. We need to test again once go1.5 is marked as stable.

uaalto:
Yes indeed. I tried many versions, and I will keep trying for more stable ones. It should be approaching beta.

xiam:
Exciting stuff. Thanks @uaalto.

uaalto:
Thank you! Let's keep an eye on Go 1.5 ;)

PR #2805, opened 2015-07-23
Adding shortcut closes #2804

### LINKED ISSUE #2804: Add desktop shortcut for opening Lantern
Several users have asked about this, and I agree it's unclear now.

myleshorton:


myleshorton:
Hey @uaalto or @xiam you mind taking a look at this? It's a one-line change =).

xiam:
Sure!

xiam:
@myleshorton is there any way we can remove the shortcut icon automatically after uninstalling Lantern? There is also a Lantern icon that remains in the start menu.

screen shot 2015-07-23 at 1 20 15 pm

xiam:
Otherwise looks and install fine :-).

myleshorton:
Oh right darn I forgot about the uninstall. Thanks @xiam will add! Good catch.

myleshorton:
OK I just added the uninstall @xiam.

xiam:
Great @myleshorton, this working fine now. Thanks!

uaalto:
@myleshorton @xiam I tried it as well, on Win 8.1. It works, install and uninstall, Desktop icon included.
I noticed that when I run the installer, it didn't ask me for confirmation or a destination folder. Given that this is normal behavior for installers, do you think that it may be good to add it as well?

xiam:
@uaalto I actually think not bothering the users with technical details is better, when I install something in Windows I just click the "Next" button until the thing gets installed. I also think choosing a destination folder may be handy when the program to be installed is huge but this is not the case with Lantern.

myleshorton:
Yeah I tend to agree @xiam. I thought about what you're saying @uaalto, but it's also super easy for users to just delete a desktop icon if they don't want it -- it's comparably quick to checking a checkbox really.

uaalto:
I actually was referring to the whole process, not the icon. When I run the installer, it just showed the progress bar directly from the double-click, no *next* button or anything.

uaalto:
This is what happens just after double-click on the lantern installer (no confirmation dialog)

screen shot 2015-07-23 at 11 25 43 pm

xiam:
> I actually was referring to the whole process, not the icon. When I run the installer, it just showed the progress bar directly from the double-click, no next button or anything.

Sorry, I was referring to a standard installation on Windows, clicking "Next" is what I'd typically do, but I would rather not bother clicking or reading anything to install a program, that's why I like the current Lantern installation on Windows.

uaalto:
I really don't have a strong position on this topic at all, but I just raise it as it could be potentially seen by users as *too aggressive* (i.e. from double-click straight to installed in your system without confirmation dialog).

Issue #2813, opened 2015-07-25
Multicast listening causing user warnings on OSX and Windows


myleshorton:
We used to handle this through code signing on OSX and adding a firewall rule on Windows, but that's let's punt on that until the next release. Actually the code signing approach might still be working on OSX not sure.

uaalto:
@myleshorton Do we have old code doing this for reference? I would like to get to this to make Lantern Pi available.

uaalto:
This will be apparently solved by code signing, right?

myleshorton:
On OSX it will be solved by code signing, but not on windows. Our old code that handled this one Windows is at:

https://github.com/getlantern/lantern-java/blob/master/install/win/Lantern.bat

We'll need to integrate something similar for give mode lantern in general actually. Would you mind ticketing that @uaalto?

uaalto:
Done. I'll be adding more information there if necessary. I keep track of that issue as part of the P2P work.

myleshorton:
Awesome!

On Fri, Aug 14, 2015 at 12:45 AM, Ulysses Aalto <notifications@github.com>
wrote:

> Done. I'll be adding more information there if necessary. I keep track of
> that issue as part of the P2P work.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/getlantern/lantern/issues/2813#issuecomment-131008849>
> .
>



-- 
--
President
Brave New Software Project, Inc.
https://www.getlantern.org
pgp A998 2B6E EF1C 373E 723F A813 045D A255 901A FD89

myleshorton:
This will be an issue again with IPFS.

myleshorton:
Closing since we don't have this issue now.

PR #2814, opened 2015-07-25
Temporarily disable localdiscovery #2813

myleshorton:
Note this doesn't address the actual issue in #2813 but rather temporarily disables localdiscovery so we can release 2.0.0 without worrying about this issue.

### DIFF
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -9,7 +9,6 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
-	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -25,7 +24,6 @@ import (
 	"github.com/getlantern/flashlight/client"
 	"github.com/getlantern/flashlight/config"
 	"github.com/getlantern/flashlight/geolookup"
-	"github.com/getlantern/flashlight/localdiscovery"
 	"github.com/getlantern/flashlight/logging"
 	"github.com/getlantern/flashlight/proxiedsites"
 	"github.com/getlantern/flashlight/server"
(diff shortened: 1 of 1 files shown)


myleshorton:
Can you take a quick look at this @uaalto? It's pretty trivial so I might just merge, but basically the issue is that OSX and Windows require special handling when opening listening ports that are available beyond localhost, particularly opening a windows firewall rule on Windows and just not requiring a user prompt on OSX. Code signing might actually already solve the OSX issue (I need to verify whether the user gets prompted to allow incoming connections to Lantern with an actually installed version), but the Windows issue remains.

myleshorton:
Hey I'm just going to merge this @uaalto in order to get 2.0.0 into QA, but please give it a look if you can as well as at #2813. Have you ever tried the localdiscovery stuff on windows by the way? I would think the firewall would block. I should have also tested that in my review but of course did not.

uaalto:
The code looks good, it is safely merged. The tests are still run, but I guess that's ok.

I'm surprised about OSX requirements, since I'm developing on this system. How can I reproduce it?

I'm not surprised about Windows, I didn't test under that platform, to be honest. I'll test it.

myleshorton:
If you have your firewall activated you should see an exception in there
for Lantern and or flashlight. Without that you should get a prompt to
allow incoming connections or not with the default config, and for some
reason for me it only appears at shutdown.

uaalto:
I'm not able to reproduce it on OSX a the moment. Neither of the behaviors you mention with/without the firewall. Should this happen as soon as you join the multicast group?

myleshorton:
> Should this happen as soon as you join the multicast group?

On windows yeah, on OSX for some reason it only seems to happen on shutdown.

Issue #2740, opened 2015-07-15
[ui]IE and Firefox can't show sites completely on small window


fffw:
From top to bottom: Chrome, IE and Firefox. Scrolled to bottom. IE and Firefox still can't show completely. The impact is relatively small though.

![image](https://cloud.githubusercontent.com/assets/10410437/8710808/2c8bc866-2b7f-11e5-91a7-9ee1ce27ea8f.png)

myleshorton:
I think this is worth fixing along with the screen resolution issue.

myleshorton:
This one's all set @atavism right?

myleshorton:
I think this is really an issue of the div the scrollbar is assigned to, as mentioned in #2800.

atavism:
@myleshorton I think we can just fix the footer at the bottom and set the overflow to hidden without having to assign the scrollbar to another div. Do you want to check if https://github.com/getlantern/lantern/pull/2816 fixes it?

PR #2816, opened 2015-07-25
Fix footer at bottom of UI landing page

atavism:
To resolve https://github.com/getlantern/lantern/issues/2740


fffw:
Cool! Works on IE/FF/Chrome for both Windows and Darwin. Merging.

PR #2818, opened 2015-07-26
Fixed loopback check to consider localhost to be loopback closes #2817

### LINKED ISSUE #2817: Lantern not waiting for localhost proxy to come up before issue http requests
The check at https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/util/http.go#L85 to see if the requester is attempting to proxy through localhost only works for localhost IP addresses, i.e. it does not consider "localhost" to be a loopback address. That's problematic because we use localhost for autoupdates, analytics, and possibly more, so the localhost proxy could very well not be available yet by the time those services request it.

myleshorton:
This is a pretty big bug. Basically on startup, we wouldn't actually wait for the localhost proxy to be available before trying to use it. This would result in errors like:

```
Jul 25 20:12:54.692 - 0m3s ERROR autoupdate: autoupdate.go:97 Problem checking for update: Problem fetching update: Post https://update.getlantern.org/update: http: error connecting to proxy http://localhost:8787: dial tcp 127.0.0.1:8787: ConnectEx tcp: No connection could be made because the target machine actively refused it.
```

### DIFF
--- a/src/github.com/getlantern/flashlight/util/http.go
+++ b/src/github.com/getlantern/flashlight/util/http.go
@@ -80,24 +80,37 @@ func httpClient(rootCA string, proxyAddr string, persistent bool) (*http.Client,
 			proxyAddr = host + proxyAddr
 		}
 
-		var ip net.IP
-		if ip = net.ParseIP(host); ip != nil {
-			if ip.IsLoopback() {
-				log.Debugf("Waiting for loopback proxy server to came online...")
-				// Waiting for proxy server to came online.
-				err := waitforserver.WaitForServer("tcp", proxyAddr, 60*time.Second)
-				if err != nil {
-					// Instead of finishing here we just log the error and continue, the client
-					// we are going to create will surely fail when used and return errors,
-					// those errors should be handled by the code that depends on such client.
-					log.Errorf("Proxy never came online at %v: %q", proxyAddr, err)
-				}
+		if isLoopback(host) {
+			log.Debugf("Waiting for loopback proxy server to came online...")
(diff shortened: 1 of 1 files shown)


myleshorton:
Hey can one of you guys take a look at this one too when you have a chance @fffw @atavism, @uaalto, @aranhoide or @xiam? 

Just stumbled across it in the user's logs that Mo sent in, and it's a fairly significant bug.

atavism:
LGTM. Really nice find! I'm surprised we didn't catch this sooner. It could explain several issues we were seeing previously.

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -80,24 +80,37 @@ func httpClient(rootCA string, proxyAddr string, persistent bool) (*http.Client,
...
+			log.Debugf("Waiting for loopback proxy server to came online...")
+			// Waiting for proxy server to came online.
+			err := waitforserver.WaitForServer("tcp", proxyAddr, 60*time.Second)
+			if err != nil {
+				// Instead of finishing here we just log the error and continue, the client
+				// we are going to create will surely fail when used and return errors,
+				// those errors should be handled by the code that depends on such client.
+				log.Errorf("Proxy never came online at %v: %q", proxyAddr, err)
 			}
+			log.Debugf("Connected to proxy on localhost")
```
atavism: Oh, I already merged, but a quick note: it's a little weird that if the proxy never comes online, we log the error (with the promise of failing later on) followed by a debug message saying we are connected. But not really a big deal.

PR #2823, opened 2015-07-26
Adding more detailed error info in response bodies closes #2822

### LINKED ISSUE #2822: Include more information in enproxy server error responses
This will be helpful for debugging, for example, whether or not errors are from fronting providers or from enproxy itself.

myleshorton:
This combined with more detailed client side logging should help narrow down some of these issues.

### DIFF
--- a/src/github.com/getlantern/enproxy/proxy.go
+++ b/src/github.com/getlantern/enproxy/proxy.go
@@ -348,11 +348,15 @@ func clientIpFor(req *http.Request) string {
 }
 
 func badGateway(resp http.ResponseWriter, msg string) {
-	log.Errorf("Responding Bad Gateway: %s", msg)
-	resp.WriteHeader(http.StatusBadGateway)
+	respond(http.StatusBadGateway, resp, fmt.Sprintf("Enproxy server responding Bad Gateway: %s", msg))
 }
 
 func forbidden(resp http.ResponseWriter, msg string) {
-	log.Errorf("Responding Forbidden: %s", msg)
-	resp.WriteHeader(http.StatusForbidden)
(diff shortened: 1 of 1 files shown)


myleshorton:
@fffw @atavism, @uaalto, @aranhoide or @xiam can you guys take a peek at these logging PRs when you have a chance as well? Thanks all!

src/github.com/getlantern/enproxy/proxy.go
```diff
@@ -348,11 +348,15 @@ func clientIpFor(req *http.Request) string {
...
 func forbidden(resp http.ResponseWriter, msg string) {
-	log.Errorf("Responding Forbidden: %s", msg)
-	resp.WriteHeader(http.StatusForbidden)
+	respond(http.StatusForbidden, resp, fmt.Sprintf("Enproxy server responding Forbidden: %s", msg))
+}
+
+func respond(status int, resp http.ResponseWriter, msg string) {
+	log.Errorf(msg)
+	resp.WriteHeader(status)
+	resp.Write([]byte(msg))
```
uaalto: This is the main difference with the previous approach: your reply now contains the same text that is being logged in the server. Is that supposed to be logged in Client's side as well?

fffw: Oh I just wanted to add this after found where the 403 forbidden came from! Adding log at client side is better, but I think it's easier to add after we actually have production servers return `msg` back.

myleshorton: Right yeah that's the idea @fffw. I added logging on the client side separately. I think those response bodies will actually show a CloudFlare message body, but this will allow us to be sure it's not the 403 we're returning ourselves.

uaalto:
Merging!

myleshorton:
Thanks @uaalto!

PR #2834, opened 2015-07-28
Restore rotator tests; Sort test testpackages.txt

uaalto:
Fixed `rotator` tests (they were failing), and add them to testpackages.txt. Sorted the lines of `testpackages.txt` in alphabetical order.

If `rotator` tests were deactivated for any other reasone besides failing, they could be kept ready to pass anyway.


myleshorton:
LGTM!

PR #2835, opened 2015-07-28
Handle all errors in github.com/getlantern (Closes #2807)

### LINKED ISSUE #2807: Full-project `errcheck` pass and add error checks where necessary

uaalto:
Closes #2807 
This seems to introduce different failures in tests, since the errors are caught earlier on.


src/github.com/getlantern/analytics/analytics.go
```diff
@@ -77,7 +77,11 @@ func Configure(trackingId string, version string, proxyAddr string) {
 			return
 		}
 		// Store new session info whenever client proxy is ready
-		sessionEvent(trackingId, version)
+		if status, err := sessionEvent(trackingId, version); err != nil {
+			log.Debugf("Unable to store new session info: %v", err)
```
myleshorton: I would log this as an error actually

src/github.com/getlantern/buuid/buuid.go
```diff
@@ -58,7 +62,9 @@ func (id ID) Write(b []byte) error {
 // ToBytes returns a 16-byte representation of this ID
 func (id ID) ToBytes() []byte {
 	b := make([]byte, EncodedLength)
-	id.Write(b)
+	if err := id.Write(b); err != nil {
+		log.Debugf("Unable to write as 16-bytes representation: %v", err)
```
myleshorton: I think think should be an Error too --- unlikely to ever happen, but if it does it's concerning I think.

src/github.com/getlantern/buuid/buuid.go
```diff
@@ -70,6 +76,8 @@ func FromString(s string) (ID, error) {
 // String() returns the string-encoded version like in uuid.UUID.
 func (id ID) String() string {
 	b := uuid.UUID(make([]byte, EncodedLength))
-	id.Write([]byte(b))
+	if err := id.Write([]byte(b)); err != nil {
+		log.Debugf("Unable to write as string: %v", err)
```
myleshorton: Same to me here.

src/github.com/getlantern/flashlight/settings/settings.go
```diff
@@ -110,5 +112,8 @@ func read() {
 			}
 			return nil
 		})
+		if err != nil {
+			log.Debugf("Unable to update settings: %v", err)
```
myleshorton: I think this should probably be logged as an error too, although I would imagine settings logs it internally most likely.

src/github.com/getlantern/idletiming/idletiming_conn.go
```diff
@@ -132,14 +139,18 @@ func (c *IdleTimingConn) Write(b []byte) (int, error) {
...
+			}
 			n, err := c.conn.Write(b)
 			c.markActive(n)
 			totalN = totalN + n
 			return totalN, err
 		} else {
 			// Use our own deadline
-			c.conn.SetWriteDeadline(maxDeadline)
+			if err := c.conn.SetWriteDeadline(maxDeadline); err != nil {
+				log.Debugf("Unable to set write deadline: %v", err)
```
myleshorton: I'm torn about these ones. Maybe error too? I actually think they fail all the time though because enproxy doesn't support these deadline funcs, but it's actually still unclear to me why that's not a real problem (or maybe it actually is a real bug, not sure).

src/github.com/getlantern/lantern-mobile/client.go
```diff
@@ -68,7 +68,9 @@ func newClient(addr, appName string) *mobileClient {
 		fronter: hqfd.NewDirectDomainFronter(),
 		appName: appName,
 	}
-	mClient.updateConfig()
+	if err := mClient.updateConfig(); err != nil {
+		log.Debugf("Unable to update config: %v", err)
```
myleshorton: Maybe error?

src/github.com/getlantern/lantern-mobile/client.go
```diff
@@ -137,7 +139,9 @@ func (client *mobileClient) pollConfiguration() {
 			return
 		case <-pollTimer.C:
 			// Attempt to update configuration.
-			client.updateConfig()
+			if err := client.updateConfig(); err != nil {
+				log.Debugf("Unable to update config: %v", err)
```
myleshorton: Here too?

src/github.com/getlantern/lantern-mobile/interface.go
```diff
@@ -2,7 +2,9 @@ package client
 
 // StopClientProxy stops the proxy.
 func StopClientProxy() error {
-	defaultClient.stop()
+	if err := defaultClient.stop(); err != nil {
+		log.Debugf("Unable to stop client: %v", err)
```
myleshorton: Seems like an error too, although not clear we'd be able to log it.

src/github.com/getlantern/yamlconf/yamlconf_disk.go
```diff
@@ -36,7 +36,9 @@ func (m *Manager) reloadFromDisk() (bool, error) {
 
 	if m.cfg != nil && m.cfg.GetVersion() != cfg.GetVersion() {
 		log.Trace("Version mismatch on disk, overwriting what's on disk with current version")
-		m.writeToDisk(m.cfg)
+		if err := m.writeToDisk(m.cfg); err != nil {
+			log.Debugf("Unable to write to disk: %v", err)
```
myleshorton: Error here too?

myleshorton:
Hey looks great @uaalto! A few that seem like they should be logged as errors but otherwise all good.

src/github.com/getlantern/idletiming/idletiming_conn.go
uaalto: It's a significant change in behavior if this fails. I would make it an error, and see if it appears in Loggly.

src/github.com/getlantern/lantern-mobile/client.go
uaalto: It should be then, as in the desktop version.

src/github.com/getlantern/lantern-mobile/interface.go
uaalto: I'll set it as an error too for correctness. But it probably won't log to loggly, just locally.

src/github.com/getlantern/chained/dialer.go
```diff
@@ -40,9 +40,9 @@ func (d *dialer) Dial(network, addr string) (net.Conn, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Unable to dial server: %s", err)
 	}
-	err = d.sendCONNECT(network, addr, conn)
-	if err != nil {
-		conn.Close()
+	if err := d.sendCONNECT(network, addr, conn); err != nil {
+		// We discard this error, since we are only interested in sendCONNECT
+		_ = conn.Close()
```
atavism: If we're just discarding the error, you could just do ``conn.Close()`` here, right?

src/github.com/getlantern/chained/server.go
```diff
@@ -50,7 +50,11 @@ func (s *Server) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 		fmt.Fprintf(resp, "Unable to dial %s : %s", address, err)
 		return
 	}
-	defer connOut.Close()
+	defer func() {
+		if err := connOut.Close(); err != nil {
+			log.Debugf("Unable to close connection: %v", err)
```
atavism: I think this should be logged as an error too? Same with the three below.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -95,30 +95,34 @@ func main() {
...
 	}
 
 	// Schedule cleanup actions
 	handleSignals()
-	addExitFunc(func() { logging.Close() })
+	addExitFunc(func() {
+		if err := logging.Close(); err != nil {
+			log.Debugf("Error closing log: %v", err)
+		}
+	})
```
atavism: This code to close logging and check the result appears twice (also at 102-105 above). Might be advantageous/cleaner to just move it to a separate function and use that instead?

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -190,17 +192,15 @@ func parseFlags() {
 	}
 	// Note - we can ignore the returned error because CommandLine.Parse() will
 	// exit if it fails.
-	flag.CommandLine.Parse(args)
+	_ = flag.CommandLine.Parse(args)
```
atavism: Same as above. If we're ignoring the error, why do we need the underscore at all? Just to be explicit we're ignoring it?

src/github.com/getlantern/chained/server.go
uaalto: ok!

src/github.com/getlantern/flashlight/flashlight.go
uaalto: Yes, I agree. There might be an advantage to placing error checking close to the offending function though. Namely, specialized error messages.

src/github.com/getlantern/flashlight/flashlight.go
uaalto: No if we want errcheck to pass. We should explicitly discard errors from now on.

src/github.com/getlantern/chained/dialer.go
uaalto: We need to be explicit that we are ignoring the error for errcheck to pass, which I think is good as well. It reflects what the comment states.

myleshorton:
I just reverse-merged valencia into this @uaalto, which should hopefully get the tests passing too.

myleshorton:
Weird -- it didn't merge testpackages.txt. I wonder what's up with that. Investigating.

myleshorton:
OK merging!

uaalto:
Great! Thank you!
We now could use the git hooks, just removing temporarily everything but the errcheck tests.

PR #2841, opened 2015-07-29
Stopping Lantern when port 8787 is not available. Closes #2840

### LINKED ISSUE #2840: If port 8787 is used by another application Lantern never starts
Run an application that binds at port 8787.

```
# server.py

import BaseHTTPServer as bhs, SimpleHTTPServer as shs

s = bhs.HTTPServer(("127.0.0.1", 8787), shs.SimpleHTTPRequestHandler)
s.serve_forever()
```

```
python server.py
```

And then try to start Lantern. A new process will spawn waiting forever for the port `8787` to become a proxy.

xiam:
See https://github.com/getlantern/lantern/issues/2840

### DIFF
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -253,9 +253,9 @@ func runClientProxy(cfg *config.Config) {
 	watchDirectAddrs()
 
 	go func() {
-		addExitFunc(pacOff)
 		err := client.ListenAndServe(func() {
 			pacOn()
+			addExitFunc(pacOff)
 			if showui && !*startup {
 				// Launch a browser window with Lantern but only after the pac
 				// URL and the proxy server are all up and running to avoid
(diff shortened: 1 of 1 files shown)


src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -253,9 +253,9 @@ func runClientProxy(cfg *config.Config) {
 	watchDirectAddrs()
 
 	go func() {
-		addExitFunc(pacOff)
 		err := client.ListenAndServe(func() {
 			pacOn()
+			addExitFunc(pacOff)
```
xiam: Even when `ListenAndServe()` failed the `pacOff()` function was called anyway. It's better to move it here after `pacOn()`.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -267,7 +267,7 @@ func runClientProxy(cfg *config.Config) {
 			}
 		})
 		if err != nil {
-			log.Errorf("Error calling listen and serve: %v", err)
+			exit(fmt.Errorf("Error calling listen and serve: %v", err))
```
xiam: We were logging a message instead of exiting.

myleshorton:
LGTM!

PR #2842, opened 2015-07-29
golog to print stack when PRINT_STACK=true

fffw:
Together with TRACE=true, it's very handy for debugging, especially running test.

### DIFF
--- a/src/github.com/getlantern/golog/golog.go
+++ b/src/github.com/getlantern/golog/golog.go
@@ -252,11 +252,11 @@ func (l *logger) doPrintStack() {
 		if funcForPc == nil {
 			break
 		}
-		file, line := funcForPc.FileLine(pc)
 		name := funcForPc.Name()
 		if strings.HasPrefix(name, "runtime.") {
 			break
 		}
+		file, line := funcForPc.FileLine(pc)
 		fmt.Fprintf(buf, "\t%s\t%s: %d\n", name, file, line)
 	}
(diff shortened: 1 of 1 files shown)


uaalto:
I have two comments about the change:

* I noticed that with `PRINT_STACK=true`, flashlight will output filenames with full path, which I think clutters the log a bit. Does this happen to you? Is it intentional?
* This one is minor and not caused by your change, but I noticed that these environment variables are set to `true` also if `TRACE=1` or `PRINT_STACK=1`, not just with `=true`. I think we should reflect this in the comments, even if it's fairly simple, but currently we assert that _they are only written if the program is run with environment variable "TRACE=true"_

uaalto:
Ok, I see that the full filepaths are only for the stacktrace. What do you think about shortening them to the filename only, or relative paths?

fffw:
@uaalto Sorry for the late response. I didn't notice the clutter caused, but I used to copy and paste the full path to open a file in another vim window (better method exists though).

fffw:
`ParseBool`'s comment says

> ParseBool returns the boolean value represented by the string. It accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False. Any other value returns an error.

I think less options make things clear to reader. We accept `true` and any other value is unspecified.

uaalto:
Ok to both! I would still reword the comment, because as it stands "only if" is not true!

Thank you for this change @fffw , it is indeed very useful!

PR #2845, opened 2015-07-29
Git hook (closes #2787)

### LINKED ISSUE #2787: Add errcheck, go vet, go xxx -race, etc to git hooks

uaalto:
Closes https://github.com/getlantern/lantern/issues/2787.
It is not perfect yet, and it won't allow any code to be pushed at the moment since our code doesn't pass through all the tests yet. However, as I progress in this code sanitization, it will become practical to use.

It tries to keep running time minimal by testing only packages that have changed since last commit.

I will keep informed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,11 @@ Install by copying it into the local `.git/hooks/` directory, with the `pre-push
 file name if you want to run it before pushing. Alternatively, you can name it
 `pre-commit` to run it before each commit..
 
+**Important notice**
+
+If you *must* commit without running the hooks, you can run git with the
+`--no-verify` flag.
+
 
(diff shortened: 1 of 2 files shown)


uaalto:
One option that I propose is to keep only `errcheck` activated at the moment, and as I progress on the other checks we can activate them. At the moment golint and go test -race fail and seem to require a fair chunk of attention.

myleshorton:
LGTM @uaalto! Technically we should probably merge #2835 first, but we also don't need to start using this right away, so merging.

uaalto:
Yes, exactly. I will allocate time for the other tests to pass too. Thank you!

PR #2851, opened 2015-07-30
Closes #2846

### LINKED ISSUE #2846: lantern server does not want to accept ^C and refuses to die
Launched a headless Lantern server with the latest valencia code:

```
./lantern_linux_amd64 -role=server -addr=:443
```

After a while I sent `^C` expecting the process to die:

```
^CJul 29 17:58:45.463 - 0m5s DEBUG flashlight: flashlight.go:404 Calling exit func
DEBUG flashlight: flashlight.go:404 Calling exit func
DEBUG flashlight: flashlight.go:407 No exit func remaining, exit now
^C^C^C^C^C^C^C^C
```

Lantern server ignored my signal and kept working; I had to send it another signal to actually terminate it:

```
killall ./lantern_linux_amd64 -KILL
```

I suspect this is related to the changes we made to catch termination signals, but the odd thing is that it does not happen in client mode.

uaalto:
Closes #2846 
Run the server in a goroutine, so it plays nicely with cleanup/signal-handling code.

### DIFF
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -155,9 +155,9 @@ func doMain() error {
 	log.Debug("Running proxy")
 	if cfg.IsDownstream() {
 		// This will open a proxy on the address and port given by -addr
-		runClientProxy(cfg)
+		go runClientProxy(cfg)
 	} else {
-		runServerProxy(cfg)
+		go runServerProxy(cfg)
 	}
 
 	return waitForExit()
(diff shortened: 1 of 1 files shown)


xiam:
@uaalto We have this `waitForExit()` function that blocks until something writes to `exitCh` in `flashlight.go`, I was thinking if there is any strong reason to spawn a goroutine only for the `ListenAndServe()` blocks on both client and server code instead of making `runClientProxy()` and `runServerProxy()` run inside goroutines. Something like this:

```
  log.Debug("Running proxy")
  if cfg.IsDownstream() {
    // This will open a proxy on the address and port given by -addr
    go runClientProxy(cfg)
  } else {
    go runServerProxy(cfg)
  }

  return waitForExit()
```

I think that would make this part of the code more clear on the blocking nature of `waitForExit()`, and if `runServerProxy()` ever needs to block itself for some reason (like after calling a `srv.ListenAndServe()`) it still can do it without affecting code around it.

What do you think?

uaalto:
I like the idea! I've committed the required changes for that.

xiam:
Thanks @uaalto!

Issue #2776, opened 2015-07-21
IE can't open any site after logging off Windows with Lantern is running


fffw:
Found while testing https://github.com/getlantern/lantern/issues/2741. Can be reproduced as following:

1. Open Lantern and de-select "Run on system start" option.
2. Restart Windows, or logout and login again.
3. Launch IE to visit any site.

That's because Lantern didn't have a chance to call `pacOff()` during system logout. Solution might be similar to https://github.com/getlantern/lantern/issues/2648, but with signal handler.

Alternatively we can find a way to let IE ignore bad PAC url, if any, as Chrome and Firefox can handle this situation.

myleshorton:
Interesting. I suspect @uaalto's changes for #2669 might do the trick too. That branch should be merged soon.

fffw:
It may be a bit different on Windows. Windows never triggers `SIGTERM`, it sends `WM_QUIT` instead.  Currently, the only place we handle Windows messages is in `systray`.

fffw:
Ref https://msdn.microsoft.com/en-us/library/xdkz3x12.aspx#code-snippet-2

xiam:
Bummer, still happening after #2669 

![screen shot 2015-07-22 at 11 39 08 am](https://cloud.githubusercontent.com/assets/385670/8831188/4c559a0a-3066-11e5-9ed2-e10c9be0dcdd.png)

uaalto:
@xiam Maybe you can try with this https://github.com/getlantern/lantern/commit/8bcc8431c493a331d44847049d290655e871f5d0, or even a slightly more aggressive code that runs the `exit(nil)` even if recover doesn't return `nil`?

xiam:
@uaalto The problem seems to be that the application is just killed without any kind of warning.

I am testing with a very simple snippet apart from Lantern:

```go
package main

import (
        "log"
        "os"
        "os/signal"
        "syscall"
)

func main() {
        log.Println("Waiting...")
        ch := make(chan os.Signal)
        signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
        log.Println(<-ch)
}
```

When logging off the expected signal is not received (on Windows).

xiam:
@fffw do you think we can handle this within the systray code?

fffw:
@xiam Yes it's the only way I can think of for now. `systray` calls back to go code to clean up gracefully in the `WM_DESTROY` handler.

myleshorton:
[...]

PR #2861, opened 2015-07-31
Adding -clear-proxy-settings as workaround for #2776

xiam:
This is a workaround that relies on Lantern running at Windows start up with an specific flag that was introduced only to clear system proxy settings.

See #2776

### DIFF
--- a/installer-resources/windows/lantern.nsi
+++ b/installer-resources/windows/lantern.nsi
@@ -83,6 +83,10 @@ Section
     # This is a bad registry entry created by old Lantern versions.
     DeleteRegKey HKCU "Software\Microsoft\Windows\CurrentVersion\Run\value"
 
+    # Add a registry key to set -clear-proxy-settings. See https://github.com/getlantern/lantern/issues/2776
+    WriteRegStr HKCU "Software\Microsoft\Windows\CurrentVersion\Run" \
+                     "Lantern" "$\"$INSTDIR\lantern.exe$\" -clear-proxy-settings"
+
     # Launch Lantern
     ShellExecAsUser::ShellExecAsUser "" "$INSTDIR\lantern.exe"
 
--- a/src/github.com/getlantern/launcher/launcher_windows.go
+++ b/src/github.com/getlantern/launcher/launcher_windows.go
@@ -25,14 +25,13 @@ func CreateLaunchFile(autoLaunch bool) {
 			log.Errorf("Could not get Lantern directory path: %q", err)
 			return
 		}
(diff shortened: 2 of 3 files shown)


installer-resources/windows/lantern.nsi
```diff
@@ -83,6 +83,10 @@ Section
     # This is a bad registry entry created by old Lantern versions.
     DeleteRegKey HKCU "Software\Microsoft\Windows\CurrentVersion\Run\value"
 
+    # Add a registry key to set -clear-proxy-settings. See https://github.com/getlantern/lantern/issues/2776
```
xiam: Setting this registry key is a task of the installer. By default Lantern will only clear the system's proxy settings and quit unless it is asked to run at Windows startup.

xiam:
Hey @fffw, @myleshorton,

Could you help me reviewing this one? It should help closing #2776. 

Thanks.

myleshorton:
LGTM @fffw! Good stuff.

src/github.com/getlantern/launcher/launcher_windows.go
```diff
@@ -17,24 +18,24 @@ var (
...
+		startupCommand = fmt.Sprintf(`"%s" -startup`, lanternPath)
 	} else {
-		// We just set the value to the empty string because the windows registry
-		// library we're using doesn't support RegDeleteValue
-		err = gowin.WriteStringReg("HKCU", runDir, "Lantern", "")
-		if err != nil {
-			log.Errorf("Error removing Lantern auto-start registry key: %q", err)
-		}
+		// Just clear stored proxy settings and quit.
+		startupCommand = fmt.Sprintf(`"%s" -clear-proxy-settings`, lanternPath)
```
fffw: Oh @xiam it has a potential problem. If user want to stop using Lantern for a period and set proxy manually for their own purpose, Internet setting will be cleared each time computer started up, and user will never know the reason as Lantern already quit!

fffw: We may need be more conservative to clear PAC setting only if the PAC string is exactly what we set.

myleshorton: Oh good catch @fffw! I like the idea to do more careful checking in pacOff, but that's a bigger change too.

fffw: Yes let's add a separate ticket.

xiam: Good catch indeed @fffw. I guess we can modify the pac.exe to backup and restore original settings instead of destroying them.

fffw: Hmm... Seems we don't have a good place to backup the original settings.

PR #2864, opened 2015-07-31
Do not test cfr package, because it fails due to CloudFront deactivation

uaalto:
It came up in Travis

### DIFF
--- a/testpackages.txt
+++ b/testpackages.txt
@@ -23,7 +23,6 @@ github.com/getlantern/lantern-mobile
 github.com/getlantern/multicast
 github.com/getlantern/pathreflect
 github.com/getlantern/peerscanner/cfl
-github.com/getlantern/peerscanner/cfr
 github.com/getlantern/peerscanner/dsp
 github.com/getlantern/proxiedsites
 github.com/getlantern/proxy


xiam:
@uaalto I have seen this test failing often, do we know the reason of the failure? Would it be safe to skip it?

myleshorton:
It fails because we've hit our cloudfront distribution limit. I've
requested an increase but its still pending approval. In general I think
it's important to test CloudFlare, but I believe Alan has a pending PR for
disabling the unused CloudFront part.

uaalto:
I reviewed the code and found no issues in the code or potential dependencies that could be affected by that removal, so I merged it. It is a minimal change. This PR fixes the tests that failed due to deactivation, because they expect this code to be active.

This should be part of that same PR, but I didn't test this particular package because it is a 'subpackage' of peerscanner and didn't realize it had its own separate tests until I saw it on Travis. That's why it's an independent PR.

myleshorton:
Oh right sorry @uaalto I didn't realize CloudFront had it's own package. Looks great!

uaalto:
thank you! I didn't realize either @myleshorton  :)

PR #2871, opened 2015-08-03
Adds support for configuration embedded in packages 2857

myleshorton:


src/github.com/getlantern/flashlight/packaged/packaged.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		log.Debugf("Error opening file %v", err)
+		// This typically means the file doesn't exist. If that's
+		// the case, and we're hard coded here to open a URL, we need
+		// to write the yaml file so future auto-updated versions will
+		// also have the URL we're trying to open.
+		//writeUrlToFile()
+		return "", &PackagedSettings{}, err
+	}
+	data := make([]byte, 2000)
```
xiam: Why the 2k byte limit?

aranhoide: That's an inicial capacity, not a limit, right?

myleshorton:
OK this should finally be all set @xiam. I had to fix an issue on Windows where we were trying to open the empty string or nil, and that would cause Lantern to open a windows explorer window. Should be all set now though.

xiam:
OK @myleshorton , let's take a look.

src/github.com/getlantern/flashlight/ui/ui.go
```diff
@@ -112,13 +112,31 @@ func Show() {
...
+
+// openExternalUrl opens an external URL of one of our partners automatically
+// at startup if configured to do so. It should only open the first time in
+// a given session that Lantern is opened.
+func openExternalUrl() {
+	if openedExternal {
+		log.Debugf("Not opening external URL again")
+		return
+	}
+	defer func() {
```
xiam: Strictly speaking this `openedExternal` variable will be set to `true` even if the page could not be opened and we'll also have this 4s period on which `Show()` will continue opening browser windows by calling `openExternalUrl()`. That's not a big deal but maybe we could try setting `openedExternal` to `true` immediately after calling `openExternalUrl()` and only set if to `false` if something failed.

```
func openExternalUrl() {
  if openedExternal {
    log.Debugf("Not opening external URL again")
    return
  }

  openedExternal = true

  path, s, err := packaged.ReadSettings()
  if err != nil {
    openedExternal = false
    log.Errorf("Could not read yaml from %v: %v", path, err)
    return
  }

  time.Sleep(4 * time.Second)
  err = open.Run(s.StartupUrl)
  if err != nil {
    openedExternal = false
    log.Errorf("Error opening external page to `%v`: %v", uiaddr, err)
  }
}
```

xiam:
@myleshorton each one of the `Makefile` targets does something, except this new `manoto` target, it only sets a variable, so if we call it directly we'll se something like:

```
make manoto
make: `manoto' is up to date.
```

It would be probably better to do something like

```
MANOTO=1 make packages
```

The `Makefile` would look for the `MANOTO` env variable and set the current `MANOTO` (which we probably should rename to something more general like `PACKAGE_SETTINGS`) based on the presence of the `MANOTO` variable. We are already doing something similar with the `HEADLESS` variable.

```
HEADLESS=1 make packages
```

This `HEADLESS` is checked in `define build-tags` and the `BUILD_TAGS` variable is modified based on the presence of that value.

That makes more sense to me, but what do you think? I could take a stab at that detail.

myleshorton:
Do you want to merge this one as soon as it passes CI @xiam?

xiam:
Almost there. Thanks @myleshorton!

src/github.com/getlantern/flashlight/ui/ui.go
myleshorton: Hey I agree this should be cleaned up -- I can do that on some other pass @xiam.

PR #2874, opened 2015-08-04
Issue2863

### LINKED ISSUE #2863: Turn off PAC only if PAC url is the same as what we set
It's related to https://github.com/getlantern/lantern/pull/2861 but should be applied in general.

fffw:
Resolves #2863


src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -217,6 +206,16 @@ func runClientProxy(cfg *config.Config) {
 		exit(err)
 	}
 
+	if *clearProxySettings {
```
fffw: Move `clearProxySettings` here, as we require `cfg.UIAddr`

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -217,6 +206,16 @@ func runClientProxy(cfg *config.Config) {
...
+		// This is a workaround that attempts to fix a Windows-only problem where
+		// Lantern was unable to clean the system's proxy settings before logging
+		// off.
+		//
+		// See: https://github.com/getlantern/lantern/issues/2776
+		doPACOff("http://" + cfg.UIAddr + "/proxy_on.pac")
```
fffw: hardcode here violates DRY, but I didn't find an easy way to eliminate it.

src/github.com/getlantern/flashlight/pac.go
```diff
@@ -142,27 +142,27 @@ func pacOn() {
...
 	if atomic.CompareAndSwapInt32(&isPacOn, 1, 0) {
 		log.Debug("Unsetting lantern as system proxy")
-		doPACOff()
+		doPACOff(pacURL)
 		log.Debug("Unset lantern as system proxy")
 	}
 }
 
-func doPACOn() {
+func doPACOn(pacURL string) {
```
fffw: Passing `pacURL` as parameter seems more natural than using package level variable.

src/github.com/getlantern/pac-cmd/windows.c
```diff
@@ -96,15 +97,24 @@ int togglePac(bool turnOn, const char* pacUrl)
   options.pOptions[1].dwOption = INTERNET_PER_CONN_AUTOCONFIG_URL;
   if (turnOn) {
     options.pOptions[0].Value.dwValue = PROXY_TYPE_AUTO_PROXY_URL;
-    wchar_t url[150];
```
fffw: `url` is not used anywhere.

src/github.com/getlantern/pac-cmd/windows.c
```diff
@@ -96,15 +97,24 @@ int togglePac(bool turnOn, const char* pacUrl)
...
   else {
+    if (strlen(pacUrl) == 0) {
+      goto turnOff;
+    } 
+    if(!InternetQueryOption(NULL, INTERNET_OPTION_PER_CONNECTION_OPTION, &options, &dwBufferSize)) {
+      reportWindowsError("Querying options");
+      goto cleanup;
+    }
+    // we turn pac off only if the option is set and pac url equals what provided
+    if ((options.pOptions[0].Value.dwValue & PROXY_TYPE_AUTO_PROXY_URL) != PROXY_TYPE_AUTO_PROXY_URL
```
fffw: Checks the option as well as pac url to be more strict.

src/github.com/getlantern/pac/pac2go.sh
```diff
@@ -18,11 +18,13 @@ BINPATH=../pac-cmd/binaries
```
fffw: reorder the commands by OS to reduce the impact in case of error.

fffw:
Tested on Windows and Ubuntu Linux 32/64 bit. It will turn off PAC only if the proxy option is set to "automatic proxy setting" and the address is exactly the same as what Lantern set. It also works on Mac but didn't test via installer.

The only remaining work is to `codesign` on Mac and add to repo.

fffw:
Oh @uaalto added binary for linux arm, but I didn't have arm platform to build and test new binary.

myleshorton:
Great thanks @fffw! Hey @atavism can you review this baby?

atavism:
@myleshorton Sure!

src/github.com/getlantern/flashlight/flashlight.go
atavism: Not a big deal, but we seem to use the ``fmt.Sprintf("http://%s/proxy_on.pac", cfg.UIAddr)`` convention elsewhere, and it's cleaner to me.

atavism: We have a check below whether ``cfg.UIAddr`` is supplied or not (lines 227-229) that we probably want to come before this block, right?

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -252,13 +251,13 @@ func runClientProxy(cfg *config.Config) {
...
-		go func() {
+		      Temporarily disabling localdiscover. See:
+		      https://github.com/getlantern/lantern/issues/2813
+		      // Continually search for local Lantern instances and update the UI
+		      go func() {
 			addExitFunc(localdiscovery.Stop)
 			localdiscovery.Start(!showui, strconv.Itoa(tcpAddr.Port))
-		}()
+		      }()
 	*/
```
atavism: Looks like this was here before, but maybe we just remove this actually? IMO, it's rarely helpful keeping commented out code around as its just creates unnecessary clutter.

src/github.com/getlantern/pac-cmd/linux.c
```diff
@@ -26,17 +27,26 @@ int togglePac(bool turnOn, const char* pacUrl)
     }
   }
   else {
+    if (strlen(pacUrl) != 0) {
+      // clear pac setting only if it's equal to pacUrl
+      char* old_mode = g_settings_get_string(setting, "mode");
+      char* old_pac_url = g_settings_get_string(setting, "autoconfig-url");
+      if (strcmp(old_mode, "auto") != 0 || old_pac_url == NULL || strcmp(old_pac_url, pacUrl) != 0 ) {
```
atavism: Does ``g_settings_get_string`` return an empty string if a key is undefined, i.e. could ``old_mode`` ever be NULL here? If so, do we need to account for that before passing it to ``strcmp``? Just confirming.

src/github.com/getlantern/flashlight/flashlight.go
fffw: Oh I didn't notice the check below. It's already checked in [`Config.ApplyDefaults()`](https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/config/config.go#L215), so the below check is extra I think.

src/github.com/getlantern/flashlight/flashlight.go
fffw: The issue is open. With the comment as a clue, we may revisit it afterwards.

atavism:
OK cool, everything looks fine to me, Merlin! Let's go ahead and merge this then.

src/github.com/getlantern/pac-cmd/linux.c
fffw: Good catch @atavism . No official document found but from [this thread](https://mail.gnome.org/archives/commits-list/2010-September/msg01431.html) `g_settings_get_string` will never return NULL, also confirmed from [source code](https://github.com/GNOME/glib/blob/6cd1f8b40ff66509eedfce9c3c46cbef799a7ffe/gio/gsettings.c#L1236). I'll remove following checking against NULL.

atavism: Ah, cool. I already merged these changes so if you just want to put that into a separate PR I'll go ahead and merge it!

myleshorton:
This is great guys thanks! Do we still need to code sign the OSX and windows binaries?

fffw:
Need to sign OSX binary. Windows already signed.

myleshorton:
OK I'll sign it!

myleshorton:
Just signed it @fffw. Do I also need to rebuild the go files that embed this? What's that process again? Or does the Makefile handle that?

fffw:
The last 2 line of `pac/pac2go.sh` is to sign and embed Mac binary to go
file.

myleshorton:
Ok, so what's the sequence? When should that be run?

fffw:
Just run it, check in the changed `pac-cmd/binaries/darwin/pac` and
`pac_bytes_darwin.go`, then all set. Sorry my Internet connection is flaky
in the recent hours.

PR #2877, opened 2015-08-05
sleep unconditionally on every iteration of register loop

### LINKED ISSUE #2539: Lantern server pegs one CPU
I don't know when this started happening, but I noticed it in latest Valencia (1d38bf3d1ea582).

If I launch the same executable in a client role instead, CPU consumption is very low, barely noticeable.

I built Lantern by just using `go build` in the flashlight repo in Ubuntu 15.04 amd64.

aranhoide:
Fixes #2539.

If `registerat` flag was empty, this would busy wait forever, hogging one CPU.

### DIFF
--- a/src/github.com/getlantern/flashlight/server/server.go
+++ b/src/github.com/getlantern/flashlight/server/server.go
@@ -236,9 +236,9 @@ func (server *Server) register(updateConfig func(func(*ServerConfig) error)) {
 				if err == nil {
 					resp.Body.Close()
 				}
-				time.Sleep(registerPeriod)
 			}
 		}
+		time.Sleep(registerPeriod)
 	}
 }
 


fffw:
Good catch @aranhoide ! It's really hard to spot.

To prevent similar things from happen again, is it good to extract the registration process to separate method, and rename original one to, say, `runRegisterLoop`?

myleshorton:
Yes great catch @aranhoide. It looks like if [baseUrl is empty](https://github.com/getlantern/lantern/blob/issue2539/src/github.com/getlantern/flashlight/server/server.go#L201) the idea behind this code is that if the config changes at runtime, it will pick up those changes and keep chugging along. Is that correct? I'd be again tempted to simplify it and to scrap the fancy runtime config changes altogether, as it complicates the code and seems unlikely to ever be used in practice.

myleshorton:
But maybe this is actually fine for now, since it fixes the issue. Maybe no reason to "fix" things beyond the actual issue at hand?

myleshorton:
I think this is fine for now - going to go ahead and merge!

aranhoide:
@fffw @myleshorton:  Sorry, I was out.  I think both suggestions make sense.

The refactoring suggested by @fffw was a smaller one and worth sneaking in while we were at it, but I don't suppose it's worth a separate PR.  Sorry I missed it.

@myleshorton I don't suppose I'd ever change this particular setting without restarting the server just in case, so yeah, I agree the value of this functionality is limited.  Now, we have runtime reconfigurability all across the board, and disabling that ability for this particular area would introduce an exception to that (the only one? I can't think of any other from the top of my head).  It sounds like that could creep into more of an ambitious rethinking.  So I think it's good that we left this alone at this moment.

Issue #2791, opened 2015-07-22
Lantern opens automatically after restart even when "Run Lantern automatically on Startup" is disabled - Windows 8.1, chrome


sheena3:
I tried two different scenarios. One with Lantern running while restarting and the other is after quitting Lantern. 

In both these scenarios, Lantern started automatically after restart.

atavism:
Thanks, @sheena3. Which OS is this happening on?

sheena3:
This is in Windows 8.1. I haven't checked in my Mac and Linux systems.

xiam:
Interesting, this does not happen to me on Windows, in fact that's the main reason of #2776.

sheena3:
@xiam I forgot to mention the browser. I saw this when the default browser is chrome. I will check with other browsers now.

sheena3:
More on this issue...

@xiam I tested all the browsers and all behaves in the same manner. 

And one more thing that I noticed is, fb manoto tv page is not opening while restarting (I made sure that the "Run Lantern automatically on system startup" is disabled). Could this be because of that?

FYI - I downloaded the beta 8 version from getlantern.org and tested this.

myleshorton:
Thanks @sheena3! The Lantern version on the web site doesn't open Manoto.

myleshorton:
I've got this one guys -- our code for deleting the registry key is just wrong.

myleshorton:
This should be all set @sheena3 but please test when this goes to QA.

sheena3:
This doesn't work as expected.

Lantern is still starting automatically after restarting. I have disabled 'Run automatically on system startup' option.

myleshorton:
I think this is because you have old settings in your windows registry, @sheena3. I'll make a change to the installer that just deletes the old value. Thanks!

sheena3:
Should I download an updated version to retest?

sheena3:
@myleshorton Tried with the new Lantern. Still not working in windows but works in Mac.

myleshorton:
Only the one I emailed about last week -- should say in the bottom:

```Lantern 2.0.0 (20150729.035956)```

sheena3:
Yeah - that is the version I used to retest this issue and reopened.

myleshorton:
Oh I think I just realized what's happening -- that version doesn't include a fix I added to overwrite old Lantern settings. Next one should fix it.

sheena3:
Do we have a new version now that I can use for retesting?

[...]

PR #2890, opened 2015-08-06
Fixed registry from installer and windows and auto-start by default closes #2802

### LINKED ISSUE #2802: Don't open web browser on automatic startup

myleshorton:


src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -298,7 +299,8 @@ func (cfg *Config) applyClientDefaults() {
 
 	if cfg.AutoLaunch == nil {
 		cfg.AutoLaunch = new(bool)
-		*cfg.AutoLaunch = false
+		*cfg.AutoLaunch = true
+		launcher.CreateLaunchFile(*cfg.AutoLaunch)
```
myleshorton: We want two things here. First, I think it makes sense to make auto-launch the default, especially because users have had trouble finding Lantern once it's installed. Second, we want to reflect whatever the starting state is on startup, as otherwise we won't set the registry setting or plist until the user actually interacts with the UI. Come to think of it, though, that call should be outside of the check for nil -- fixing now.

myleshorton: Actually I think the above is what we want after all, as the only other way AutoLaunch should ever get populated is from the user, in which case CreateLaunchFile is called.

myleshorton:
Hey @xiam can you take a peek at this one?

xiam:
I confirmed that this change removes the "Lantern" and the "value" string values from the `Run` key.

xiam:
This will help closing https://github.com/getlantern/lantern/issues/2791, thanks @myleshorton

PR #2893, opened 2015-08-06
*really* remove CloudFront and DNSimple support from peerscanner

aranhoide:
A lot was missing.

Sorry about the previous PR.  I thought I couldn't properly test this without bringing down the production peerscanner because we didn't have enough spare domains, and that's why it
went untested.  But since we're removing CloudFront and hence DNSimple support, we only need one domain.


fffw:
All CloudFront and DNSimple related calls and functions in package `main` were commented out. Ideally we can extract an interface from the implementations, so we can change one line of code to remove CloudFront+DNSimple, or potentially add other CDNs in the future, but it's not trivial.

@aranhoide It will not add any domain by just running a peerscanner without registering a new server right? I tried to run it for a while to verify but it will deregister and register a few servers due to timeouts testing them, so I skip it to avoid affecting production system.

Overall, it LGTM. Thanks!

aranhoide:
@fffw thanks!  To test it, you'd use the `-cfldomain` command line flag, which specifies where to register/unregister servers.  We have the test domain flashlightproxy.com for this purpose, so you'd say `./peerscanner -cfldomain flashlightproxy.com`.  Also, for any flashlight server to register to your test peerscanner you'd have to call it with `./flashlight -registerat https://yourserverurl.org`.

I have [this test peerscanner](https://cloud.digitalocean.com/droplets/4467475) to test stuff.  `ps-test.getiantem.org` points to it.  It's normally turned off.  Whenever you want to test anything peerscanner related, feel free to log into it, copy over a new peerscanner binary, and start it.

myleshorton:
Awesome. @aranhoide can you add those notes to the peerscanner readme?

aranhoide:
Done.

myleshorton:
Nice thanks!

PR #2895, opened 2015-08-07
Provide TLS Config to failing tests that randomly reach handshaking and fail unproperly

### LINKED ISSUE #2858: tlsdialer_test.go:224: x509: certificate signed by unknown authority

uaalto:
Closes #2858 
I've run the tests on Travis 5 times. This error is not easily reproducible, it is a matter of times run. In case it fails again, we will reduce the timeout from 5000 microseconds to something like 3000-4000. But if it works with this tweak, is best to keep it at 5000.


uaalto:
I'm running this non-stop today and didn't fail once. Looks good for now.

myleshorton:
OK let's try it! Thanks @uaalto.

PR #2909, opened 2015-08-10
Parallel access to improve detour performance

### LINKED ISSUE #2470: Improve detour logic to work well on flaky networks
Forgot to log an issue. Basically, we can use parallel access to strip out the arbitrary 3s timeout.

fffw:
Resolves #2470 

Finally get this rewritten version work. Sorry it takes so much longer than it should be.

The behaviour of this version is much more dynamic than previous one. Though passed all tests, it might still be buggy. Any of your suggestions are welcome!


fffw:
When used in Lantern, it leaks connections in long run and leads to "too many open files" error finally.

fffw:
Those increasing CLOSE_WAIT connections which caused "too many open files" error might be caused by chained dialer didn't return `idletiming.Conn`. Create https://github.com/getlantern/lantern/issues/2927 to handle it.

xiam:
Awesome @fffw, let's carefully review this one.

myleshorton:
Nice @fffw. You could try just implementing the #2927 change on your branch
to see if it fixes it too.

On Tue, Aug 11, 2015 at 11:29 PM, fffw <notifications@github.com> wrote:

> Those increasing CLOSE_WAIT connections which caused "too many open files"
> error might be caused by chained dialer didn't return idletiming.Conn.
> Create #2927 <https://github.com/getlantern/lantern/issues/2927> to
> handle it.
>
> —
> <https://github.com/getlantern/lantern/pull/2909#issuecomment-130185524>.
>



-- 
--
President
https://www.getlantern.org

src/github.com/getlantern/detour/detour.go
```diff
@@ -43,325 +38,327 @@ import (
...
+							_ = c.Close()
+							return
+						}
+						log.Tracef("Feed detour connection to %s to read/write op", dc.addr)
+						dc.chDetourConn <- c
+					}
+				case <-t.C:
+					// still no connection made
+					chAnyConn <- false
+					return
```
fffw: An potential issue here. On extremely slow network, connection made after this timeout will have no where to go but remain open.

src/github.com/getlantern/detour/detour.go
```diff
@@ -43,325 +38,326 @@ import (
...
-		return
-	}
-	// Hijacked content is usualy encapsulated in one IP packet,
-	// so just check it in one read rather than consecutive reads.
-	if dc.inState(stateDirect) && detector.FakeResponse(b) {
-		log.Tracef("%s still content hijacked, add to whitelist so will try detour next time", dc.addr)
-		AddToWl(dc.addr, false)
-		return
+func (dc *Conn) followupRead(b []byte) (n int, err error) {
+	if !dc.withValidConn(func(c conn) { c.FollowupRead(b, dc.chRead) }) {
```
fffw: Both direct and detour connection can be available at this point. If we don't want to read/write on both connections, we should keep one and close another at first read. It's definitely a bug.

fffw:
Model it as a state machine may make the code more understandable and less error prone, though the change is not small.

src/github.com/getlantern/detour/direct_conn.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func dialDirect(network string, addr string, ch chan conn) {
+	go func() {
+		log.Tracef("Dialing direct connection to %s", addr)
+		conn, err := net.DialTimeout(network, addr, TimeoutToConnect)
+		detector := blockDetector.Load().(*Detector)
+		if err == nil {
+			if detector.DNSPoisoned(conn) {
+				_ = conn.Close()
```
xiam: What's the purpose of the blank identifier here?

src/github.com/getlantern/detour/direct_conn.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		ch <- ioResult{n, err, dc}
+	}()
+	return
+}
+
+func (dc *directConn) Write(b []byte, ch chan ioResult) {
+	go func() {
+		n, err := dc.Conn.Write(b)
+		defer func() { ch <- ioResult{n, err, dc} }()
```
xiam: Why using `defer` here? The function has already ended at this point and there is nothing left to do.

src/github.com/getlantern/detour/detour.go
```diff
@@ -43,325 +38,326 @@ import (
...
 		}
-		log.Tracef("Dial %s to %s succeeded", dc.stateDesc(), addr)
-		if !whitelisted(addr) {
-			log.Tracef("Add %s to whitelist", addr)
-			AddToWl(dc.addr, false)
+		// handle dialing result
+		go func() {
+			t := time.NewTimer(TimeoutToConnect)
+			defer t.Stop()
+			for i := 0; i < 2; i++ {
```
xiam: Could we have a comment describing the purpose of this loop?

src/github.com/getlantern/detour/detour.go
```diff
@@ -43,325 +38,326 @@ import (
...
+						chAnyConn <- true
+					} else {
+						if c.ConnType() == connTypeDirect {
+							// Could happen if direct route is much slower.
+							log.Debugf("Direct connection to %s established too late, close it", dc.addr)
+							_ = c.Close()
+							return
+						}
+						log.Tracef("Feed detour connection to %s to read/write op", dc.addr)
+						dc.chDetourConn <- c
```
xiam: When a connection is received and it is not direct, we'll write to this channel and then the loop will exit, because `i` is now `1`. It would probably be safer to end this loop explicitly with `return`. My reasoning is that if `i` would be allowed to grow to `3` or more, then this loop will continue one more time at least, the problem is we already have what we wanted.

src/github.com/getlantern/detour/detour.go
```diff
@@ -43,325 +38,326 @@ import (
...
 }
 
-// SetWriteDeadline() implements the function from net.Conn
+// SetWriteDeadline implements the function from net.Conn
 func (dc *Conn) SetWriteDeadline(t time.Time) error {
-	dc.writeDeadline = t
-	if err := dc.getConn().SetWriteDeadline(t); err != nil {
-		log.Debugf("Unable to set write deadline", err)
-	}
+	log.Trace("SetWriteDeadline not implemented")
```
xiam: Why logging it instead of returning a "SetWriteDeadline not implemented" error?

src/github.com/getlantern/detour/detour.go
```diff
@@ -369,60 +365,15 @@ var nonIdempotentMethods = [][]byte{
...
-// ref section 9.1.2 of https://www.ietf.org/rfc/rfc2616.txt.
-// checks against non-idemponent methods actually,
-// as we consider the https handshake phase to be idemponent.
-func (dc *Conn) isIdempotentRequest() bool {
-	dc.muLocalBuffer.Lock()
-	defer dc.muLocalBuffer.Unlock()
-	b := dc.localBuffer.Bytes()
+// Ref section 9.1.2 of https://www.ietf.org/rfc/rfc2616.txt.
+// We consider the https handshake phase to be idemponent.
+func isNonIdempotentHTTPRequest(b []byte) bool {
```
xiam: I am not sure if we can just assume this non-idempotence, as far as I can tell idempotence depends on the application. According to this function GET, DELETE and PUT are considered equally idempotent, but that may not be the case.

For instance, some applications may do things like:

GET /posts/delete?id=123

Which is a GET and by this check would be considered idempotent, nothing prevents a developer for doing that. Is there any way we can stay in the safe side? It seems like this could cause side effects that may be difficult to track back to Lantern.

src/github.com/getlantern/detour/detour_conn.go
```diff
@@ -0,0 +1,82 @@
...
+			return
+		}
+	}()
+	return
+}
+
+func (dc *detourConn) Write(b []byte, ch chan ioResult) {
+	go func() {
+		n, err := dc.Conn.Write(b)
+		defer func() { ch <- ioResult{n, err, dc} }()
```
xiam: The function is already about to end at this point, I think the defer is not necessary.

xiam:
Hey @fffw, this PR looks great. I just added a few comments on it, what do you think? I'll take another close look afterwards.

src/github.com/getlantern/detour/direct_conn.go
fffw: Just want to ignore the errors when closing `conn`. Add a debug message would be better. Thanks for pointing out.

src/github.com/getlantern/detour/direct_conn.go
fffw: Ha yes good catch!

src/github.com/getlantern/detour/detour.go
fffw: Good point, thanks.

src/github.com/getlantern/detour/detour.go
fffw: Agreed. Explicit return will express the intention more clearly.

src/github.com/getlantern/detour/detour.go
fffw: I just followed what `enproxy` doing. Returning an error makes the code more deterministic indeed.

src/github.com/getlantern/detour/detour.go
fffw: `GET /posts/delete?id=123` is still idempotent as long as `ID` is an actual ID representing an unique resource instead of a sequence number. Something like `GET /posts/create` is the problem.

Since we can't enumerate all applications behaviour, the least thing we can do is to follow standard. Just like browser not prompting user when refreshing a GET request, we can do the same I think.

fffw:
Thanks for your insights @xiam !

src/github.com/getlantern/detour/detour.go
xiam: Good catch, would it be possible to identify and close such connection? It sounds like a way to run out of file descriptors.

src/github.com/getlantern/detour/detour.go
xiam: I think you're right. Then I'd only add PUT to the `nonIdempotentMethods` blacklist, since some PUT requests can include a body.

src/github.com/getlantern/detour/detour.go
fffw: Direct connection is dialed with timeout, so it will always return before the timer runs out. There's no timeout dialing detour connection because balancer doesn't have `DialTimeout`.

Follow that path, fronted dialers return idletiming connections, I also added 1 hour idletiming to chained connections in this PR (https://github.com/getlantern/lantern/commit/fdad34209c99393ceb4bde98a200ef9e53409f52). In practice it won't cause too many dangling connections though we should fix here anyway.

xiam:
This PR looks good to me but our tests are failing due to an apparently unrelated error. I'm going to take a closer look to that.

xiam:
Yep, unrelated errors. Merging! thanks a lot @fffw.

myleshorton:
Do you think this is ready for production @fffw?

fffw:
Though not 100% sure, but it should work better than previous one. The "use of closed connection" error is not specific to this branch, so it should be caused somewhere else. I also spotted that detour connection was made quicker than direct ones in some cases, should be caused by pre-dialed connections of `connpool`, but let's see what will happen.

PR #2912, opened 2015-08-10
Temporary workaround for Firefox issue caused by reconnection with multiple open tabs

### LINKED ISSUE #2799: Lantern UI can show "Waiting for Lantern" on launch indefinitely
I was seeing this on our Windows VPS. You want to take a peek @atavism? I launched from the Start menu.

uaalto:
Closes #2799 

I've confirmed the bug is caused by Firefox, by creating a simplified test case to trigger it without Lantern's code. Bug opened at Firefox's bugzilla [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1192773).

This workaround is just for Firefox, and just shows the localized "Lantern Gone" message when the websocket is disconnected. The code is temporary, so I kept the code minimal. Let's hope it's fixed soon in Firefox.


src/github.com/getlantern/lantern-ui/app/js/app.js
```diff
@@ -122,15 +122,20 @@ var app = angular.module('app', [
...
-            $window.location.reload();
-          });
-        }, WS_RECONNECT_INTERVAL);
+        // Temporary workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1192773
+        if (navigator.userAgent.toLowerCase().indexOf('firefox') > -1) {
+          $rootScope.backendIsGone = true;
+          $rootScope.$digest()
+        } else {
+          // Try to reconnect indefinitely when the websocket closes
+          $interval(function() {
```
xiam: It seems like a new interval will be created each time the onClose event is triggered. I don't see this interval being saved to anywhere nor being interrupted later in the code.

src/github.com/getlantern/lantern-ui/app/js/app.js
```diff
@@ -122,15 +122,20 @@ var app = angular.module('app', [
...
-        // when the websocket closes
-        $interval(function() {
-          console.log("Trying to reconnect to disconnected websocket");
-          ds = $websocket('ws://' + document.location.host + '/data');
-          ds.onOpen(function(msg) {
-            $window.location.reload();
-          });
-        }, WS_RECONNECT_INTERVAL);
+        // Temporary workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1192773
+        if (navigator.userAgent.toLowerCase().indexOf('firefox') > -1) {
```
xiam: Does this mean that if we are using firefox, the browser will not try to reconnect in case the connection is dropped?

src/github.com/getlantern/lantern-ui/app/js/app.js
```diff
@@ -256,22 +261,27 @@ var app = angular.module('app', [
...
-            $log.debug("Time since last connect", timeSinceLastConnected);
-            if (timeSinceLastConnected > MILLIS_UNTIL_BACKEND_CONSIDERED_GONE) {
-              // If it's been more than 10 seconds since we last connect,
-              // treat the backend as gone
-              console.log("Backend is gone");
-              $rootScope.backendIsGone = true;
-            } else {
-              $rootScope.backendIsGone = false;
+        // Temporary workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1192773
+        if (navigator.userAgent.toLowerCase().indexOf('firefox') > -1) {
```
xiam: If we are on firefox, then the timeout won't be executed, is that expected?

src/github.com/getlantern/lantern-ui/app/js/app.js
uaalto: Yes

src/github.com/getlantern/lantern-ui/app/js/app.js
uaalto: The interval-triggered function is only executed in browsers that are not Firefox. That is, when we are on Firefox, we won't reconnect and it will show the localized "Lantern Gone" message.

src/github.com/getlantern/lantern-ui/app/js/app.js
xiam: I've experimented that after a while (between 40 and 50s), Firefox attempts to connect again and succeeds, in each one of the tabs. Will this prevent Firefox to ever reconnect?

src/github.com/getlantern/lantern-ui/app/js/app.js
xiam: Yes. My observation was oriented towards the fact that this interval will be created each time this event is triggered and this interval is not saved anywhere so, as far as I understand, it cannot be stopped, if Lantern gets disconnected five times, then we will be running five different intervals checking for the same thing.

It seems like you didn't introduce this so maybe we can follow this problem on a different issue, I think it won't cause I'll effects unless we're running hundreds of timeouts but it may cause some hard to catch bugs.

src/github.com/getlantern/lantern-ui/app/js/app.js
uaalto: Yes, that 40-50 secs. delay is the bug. It grows if you add more tabs, and there is a "memory" of opened tabs, so even if you close them it happens.
The only way I found the "Waiting for Lantern" message can be avoided when you run Lantern (if the bug has been triggered) is by preventing Firefox to reconnect automatically. So yes, the intention of this workaround is that once lantern is shut down and then restarted, then you need to open a new page or reload manually.

src/github.com/getlantern/lantern-ui/app/js/app.js
uaalto: Yes, I see what you mean, and is one of the things I thought could be the cause of this bug. Unfortunately other tests lead me to the conclusion that what triggers the bug is _previously opened connections_, not many of them trying to open at the very same time (I tried that by heavily randomizing this timeout among other ways).

However, this interval is not going to be created each time, because it effectively overrides the onOpen callback, which will reload the web page and reinitialize the whole thing. A sure way of starting from scratch each time :)

src/github.com/getlantern/lantern-ui/app/js/app.js
uaalto: Yes, it's intentional. This timeout is what shows the message "Lantern is Gone" after 11 seconds in the "waiting for lantern" status.

src/github.com/getlantern/lantern-ui/app/js/app.js
xiam: Oh, gotcha!

xiam:
Hello @uaalto,

This is what happens on Firefox/Win using your branch:

After closing Lantern the "Waiting for Lantern" message appears.

screen shot 2015-08-12 at 11 16 07 am

A while we see the "Not running anymore" message:

screen shot 2015-08-12 at 11 16 15 am

And after restarting Lantern and waiting about a minute Lantern goes back to normal:

screen shot 2015-08-12 at 11 20 07 am

But the difference is not too clear to me (as an user) yet, because the very same thing happened before the change, as an user I have no need to open a new page or to do anything by hand (which I actually think is good).

uaalto:
Thanks @xiam. The web shouldn't connect at all again once it has disconnected. There is no other code doing producing a reload AFAIK. You need to reload the page manually. That's the expected behavior with this workaround.

The first two states of the images you posted are to be expected now. The third isn't. Are you saying that if you leave the old page alone, that was showing the message "Lantern appears not to be running anymore", and you run Lantern again, after a minute or so the page will reload __automatically__?

xiam:
> Are you saying that if you leave the old page alone, that was showing the message "Lantern appears not to be running anymore", and you run Lantern again, after a minute or so the page will reload automatically?

That's right, I'll send you a video.

uaalto:
Thank you @xiam. It may be about [this line](https://github.com/getlantern/lantern/blob/issue-2799/src/github.com/getlantern/lantern-ui/app/js/app.js#L196). I'll have to take a look then.

xiam:
@uaalto I confirm this was a problem with my testing. And the result does not actually look confusing at all as I was supposing :-).

xiam:
Thanks @uaalto

uaalto:
Thank you!

PR #2917, opened 2015-08-11
fix git hook

fffw:

### DIFF
--- a/git-hook
+++ b/git-hook
@@ -28,8 +28,8 @@ which golint >/dev/null || (echo "Unable to find golint, please install it: \`go
 
 echo "*** Running \033[0;34mErrcheck\033[0m ***" && \
 for dir in $MODIFIED_DIRS; do
-  errcheck github.com/getlantern/$dir || \
-    (echo "\033[0;31merrcheck returned error analyzing the package '$dir'\033[0m" && \
+  errcheck github.com/getlantern/$dir || (\
+    echo "\033[0;31merrcheck returned error analyzing the package '$dir'\033[0m" && \
     echo "Please, fix and run again\n" && \
     exit 1)
 done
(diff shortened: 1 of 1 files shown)


git-hook
```diff
@@ -27,16 +27,16 @@ which errcheck >/dev/null || (echo "Unable to find errcheck, please install it:
 which golint >/dev/null || (echo "Unable to find golint, please install it: \`go get -u github.com/golang/lint/golint\`" && exit 1)
 
 echo "*** Running \033[0;34mErrcheck\033[0m ***" && \
-for dir in *; do
+for dir in $MODIFIED_DIRS; do
   errcheck github.com/getlantern/$dir || \
-    echo "\033[0;31merrcheck returned error analyzing the package '$dir'\033[0m" && \
+    (echo "\033[0;31merrcheck returned error analyzing the package '$dir'\033[0m" && \
```
fffw: Explicitly group statements, as in Bash,  || and && are the same precedence.

git-hook
```diff
@@ -27,16 +27,16 @@ which errcheck >/dev/null || (echo "Unable to find errcheck, please install it:
 which golint >/dev/null || (echo "Unable to find golint, please install it: \`go get -u github.com/golang/lint/golint\`" && exit 1)
 
 echo "*** Running \033[0;34mErrcheck\033[0m ***" && \
-for dir in *; do
+for dir in $MODIFIED_DIRS; do
```
fffw: Only check modified directions.

fffw:
git hook is super useful to catch errors. Thanks @uaalto ! It would be good if you can take a look at this quick fix.

fffw:
Oh I just noticed that it can only act as pre-commit hook after my change, as when `git push`, staging area is clean. But it will take too long to errcheck and test race for all code, pre-commit hook with only the changed directory might be more useful.

git-hook
uaalto: wow, this was definitely meant to be! Thank you @fffw !

git-hook
uaalto: Thank you! A very tiny detail, though. Could you please place the `(` as in the other cases? (or change the other cases if you prefer). Even though is a tiny style detail, Bash is already a quirky language, so uniformity pays off in my opinion.

fffw: Strange enough I didn't notice the brackets below! Will correct.

fffw:
Updated @uaalto .

uaalto:
Awesome! LGTM.

PR #2919, opened 2015-08-11
Prepend timestamp only on first line of a multiline output

### LINKED ISSUE #2914: Append log prefix only once in multiline output.
This reduces log size and makes it easier to use (copy&paste).

uaalto:
Closes #2914


myleshorton:
Looks good @uaalto -- certainly simpler than the LinePrepender too!

PR #2923, opened 2015-08-11
Override the IP address sent to analytics closes #2922

### LINKED ISSUE #2922: Override the IP address in analytics data
All requests are currently considered to be in the country of the proxy server. See

https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide?hl=en#using-a-proxy-server

myleshorton:
This is necessary to get accurate geo data in analytics.

### DIFF
--- a/src/github.com/getlantern/flashlight/analytics/analytics.go
+++ b/src/github.com/getlantern/flashlight/analytics/analytics.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 
 	"github.com/getlantern/flashlight/config"
+	"github.com/getlantern/flashlight/pubsub"
 	"github.com/mitchellh/mapstructure"
 
 	"github.com/getlantern/analytics"
@@ -27,12 +28,15 @@ var (
 func Configure(cfg *config.Config, version string) {
 
 	if cfg.AutoReport != nil && *cfg.AutoReport {
-		analytics.Configure(TrackingId, version, cfg.Addr)
+		pubsub.Sub(pubsub.IP, func(ip string) {
+			log.Debugf("Got IP %v -- starting analytics", ip)
+			analytics.Configure(ip, TrackingId, version, cfg.Addr)
 
-		err := StartService()
-		if err != nil {
-			log.Errorf("Error starting analytics service: %q", err)
-		}
+			err := StartService()
+			if err != nil {
+				log.Errorf("Error starting analytics service: %q", err)
+			}
+		})
 	}
 }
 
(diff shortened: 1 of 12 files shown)


myleshorton:
Hey @uaalto you mind taking a look at this one?

src/github.com/getlantern/flashlight/analytics/analytics.go
```diff
@@ -27,12 +28,15 @@ var (
 func Configure(cfg *config.Config, version string) {
 
 	if cfg.AutoReport != nil && *cfg.AutoReport {
-		analytics.Configure(TrackingId, version, cfg.Addr)
+		pubsub.Sub(pubsub.IP, func(ip string) {
+			log.Debugf("Got IP %v -- starting analytics", ip)
+			analytics.Configure(ip, TrackingId, version, cfg.Addr)
```
myleshorton: I'm a little bit concerned about this snippet only because the UI could try to call the analytics service before it's up. That might have also been possible before not sure, and the consequences are probably very minimal, but just wanted to flag it.

uaalto: If the `pubsub.IP` topic takes long to be published, it could happen. But I see an interface in analytics that only allows for starting/stopping the service, so how could the UI actively call the service?

uaalto:
geolookup test need to be updated for the new function prototype too:
```
geolookup/geolookup_test.go:13: assignment count mismatch: 2 = 3
geolookup/geolookup_test.go:29: assignment count mismatch: 2 = 3
```

uaalto:
In order to test the analytics code, are you using any kind of mock server, test account or a specific way of making sure it works before moving to production?

src/github.com/getlantern/flashlight/analytics/analytics.go
myleshorton: Right -- I'm not familiar with how it works on the lantern-ui side. Does
that only make the call to analytics once the backend sets it up? If so
then we're golden.

myleshorton:
Will fix!

myleshorton:
I was testing using the realtime analytics with a different account number.

myleshorton:
OK just fixed that test thanks @uaalto. I also added that package to the automated tests.

uaalto:
The only point I see bad things could happen is in the PubSub part, but is clear code. Golint produces many errors, but they come from previous code. It still have to do the full Golint project pass for that to be make sense.
The PR code LGTM! Merging.

PR #2942, opened 2015-08-13
Report and use IP and country from CloudFlare closes #2941

### LINKED ISSUE #2941: Use CloudFlare's geo IP lookup in flashlight servers
This will avoid using geoIP at all. We can simply return the country in a custom HTTP header (or just re-use CloudFlare's) and can also return the user's IP in a custom header.

myleshorton:
I just went ahead and implemented the server side of this since it's really straightforward. You want to take a look @aranhoide?

### DIFF
--- a/src/github.com/getlantern/enproxy/proxy.go
+++ b/src/github.com/getlantern/enproxy/proxy.go
@@ -130,6 +130,9 @@ func (p *Proxy) Serve(l net.Listener) error {
 
 // ServeHTTP: implements the http.Handler interface
 func (p *Proxy) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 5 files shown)


aranhoide:
LGTM, tests pass in my machine.  The only nit is that IIRC we've been using the "X-Lantern-" prefix for headers instead of "Lantern-", but I won't lose any sleep over this.  Feel free to either change that or merge this PR as is.

myleshorton:
OK merging since the X- is an old convention that's no longer recommended.

myleshorton:
Can you deploy this when you have a chance @aranhoide?

aranhoide:
@myleshorton done.

PR #2948, opened 2015-08-14
use versioned dmg title fixes #2947

### LINKED ISSUE #2947: Use versioned title for Lantern dmg
Confusion will happen if we opened multiple dmgs.

1. Open a Lantern dmg, say beta8, it will add a volume named "Lantern" and show the install window.
2. Open a dmg with different version, it will add a volume with the same name, and the background image is from the previous dmg! I guess the Lantern.app it contains is also from previous dmg.

Use versioned title will resolve this.

fffw:


myleshorton:
Nice @fffw! LGTM.

PR #2951, opened 2015-08-14
Fix issue where welcome screen text disappears

atavism:
This fixes an issue where the Welcome text disappears after ten seconds. That text is hidden when the backend goes away and the text "Lantern appears not to be running anymore"


myleshorton:
LGTM thanks @atavism!

PR #2952, opened 2015-08-14
Another attempt to fix analytics closes #2788

### LINKED ISSUE #2788: Client analytics not working?
We've actually seen a drop in our client analytics over the last few days even as downloads have gone up significantly. It seems like something's not working correctly.

You mind taking a look @atavism? I can give it a try too.

myleshorton:
This is another attempt at this, as nothing seems to be working @atavism. This just tries to strip things down to make them as simple as possible, including going back to just using the javascript stuff on the UI side since that's the only thing we've seen consistently work.

What do you think?


src/github.com/getlantern/lantern-ui/app/js/constants.js
```diff
@@ -35,7 +35,7 @@ var DEFAULT_LANG = 'en_US',
       es: {dir: 'ltr', name: 'español'},
       ar: {dir: 'rtl', name: 'العربية'}
     },
-    GOOGLE_ANALYTICS_WEBPROP_ID = 'UA-21815217-12',
+    GOOGLE_ANALYTICS_WEBPROP_ID = 'UA-21815217-13',
```
myleshorton: Just using separate IDs for the UI and backend now.

myleshorton:
Hey just wanted to ping you on this @atavism. Our analytics is still totally broken, so I'd really like to get this in production ASAP!

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -222,7 +221,7 @@ func (cfg *Config) ApplyDefaults() {
 	}
 
 	if cfg.InstanceId == "" {
-		cfg.InstanceId = hex.EncodeToString(uuid.NodeID())
+		cfg.InstanceId = uuid.New()
```
myleshorton: We're now using this as the client ID (cid) @atavism. From the discussion [here](https://groups.google.com/forum/#!topic/google-analytics-measurement-protocol/LN8WfJAPla8) it sounds like the whole cid=555 thing might be the reason we've seen data for a day or so with a lot of these but then somehow have seen that data totally stop.

I changed this to a version 4 UUID because that's what analytics expects, and there's no reason not to use version 4 for our normal InstanceId.

atavism:
@myleshorton Looking over your changes now.

myleshorton:
Awesome thanks @atavism! You'll see I just skip the entire library and interaction with the frontend just to make it as simple as possible in hopes of getting this darn thing working!

src/github.com/getlantern/flashlight/config/config.go
atavism: Yeah, the comment about '555' being used and initially seeing a few hits than none is definitely similar to what I've encountered. It sounds like if you don't vary the client id, future requests end up getting blocked.

src/github.com/getlantern/lantern-ui/app/js/services.js
```diff
@@ -205,30 +205,9 @@ angular.module('app.services', [])
...
-                if (gaObj.hasOwnProperty(key)) {
-                    gaObj[key] = fields[name];
-                }
-              }
-              DataStream.send('Analytics', gaObj);
-          }
-      }
+    function trackPageView() {
+			console.log("SENDING PAGE VIEW")
+      ga('send', 'pageview');
```
myleshorton: I just went back to this approach because it's the only thing we've seen consistently work!

src/github.com/getlantern/flashlight/config/config.go
myleshorton: Yeah exactly -- that sounds exactly like what I've seen too, both with that last pass with this new ID and with our damn emails.

src/github.com/getlantern/flashlight/settings/settings.go
```diff
@@ -92,14 +91,6 @@ func read() {
...
 
 			if autoReport, ok := settings["autoReport"].(bool); ok {
-				// turn on/off analaytics reporting
-				if autoReport {
-					if err := analytics.StartService(); err != nil {
-						log.Debugf("Unable to start analytics service: %v", err)
-					}
-				} else {
-					analytics.StopService()
-				}
```
atavism: There is the case where if the user enables auto-reporting, after having it disabled, we don't immediately track the session until Lantern restarts, but since auto-reporting is the default now that shouldn't be much of a problem anyway.

atavism:
LGTM! Simplifying by removing the analytics service makes the code clearer, and it's working fine for me.

myleshorton:
Awesome thanks @atavism!

PR #2956, opened 2015-08-17
Include OS version in Loggly messages

### LINKED ISSUE #2252: detect os version in go
We need it for error reporting to Loggly.
@oxtoacart has a great reference  https://groups.google.com/forum/#!topic/golang-nuts/NzEPGqlHIYQ

uaalto:
Closes #2252 

The OS version detection code is custom and tries to be minimal and avoid opening files when possible. Currently supports Linux, OS X, Windows and Android. Each one uses a different technique. iOS has been tried but is not mature enough yet (more specifically, can't use cgo code ATM, and fails to compile with some features).

`GetString()` provides a major.minor.patch string, `GetSemanticVersion()` uses this string to construct a semantic version object. `GetHumanReadable()` provides a string with useful OS information, which depends on the OS and the available features.


uaalto:
An update to some of Gomobile's dependencies was required for Android support.

src/github.com/getlantern/osversion/cmd/main_android.go
```diff
@@ -0,0 +1,26 @@
...
+		if err != nil {
+			log.Println("Error")
+		}
+		log.Println(str)
+
+		str, err = osversion.GetHumanReadable()
+		if err != nil {
+			log.Println("Error")
+		}
+		log.Println(str)
```
myleshorton: Android supports GetSemanticVersion() too right? Not a big deal since it looks like this is basically demo code, but any reason to leave it out here?

myleshorton: Oh maybe it doesn't support GetSemanticVersion() huh?

uaalto: Oh, yes it does. That's common code (you can find it in osversion.go, without _<tag> qualifiers). I didn't add it because it is not platform-dependent.
I will add it in the next commit!

uaalto: Done.

myleshorton:
OK LGTM! Thanks @uaalto

PR #2959, opened 2015-08-17
Register and report User-Agents used in Lantern session

### LINKED ISSUE #2925: Detect and report user-agent in loggly

uaalto:
Closes #2925

### DIFF
--- a/src/github.com/getlantern/flashlight/logging/useragents.go
+++ b/src/github.com/getlantern/flashlight/logging/useragents.go
@@ -1,7 +1,9 @@
 package logging
 
 import (
+	"bytes"
 	"fmt"
+	"regexp"
 	"sync"
 )
 
@@ -10,6 +12,7 @@ type agentsMap map[string]int
 var (
 	userAgents  = make(agentsMap)
 	agentsMutex = &sync.Mutex{}
+	reg         = regexp.MustCompile("^Go.*package http$")
 )
 
 // registerUserAgent tries to find the User-Agent in the HTTP request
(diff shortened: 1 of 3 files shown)


src/github.com/getlantern/flashlight/logging/useragents.go
```diff
@@ -0,0 +1,50 @@
+package logging
+
+import (
+	"bytes"
+	"fmt"
+	"regexp"
+	"sync"
+)
+
+type agentsMap map[string]int
```
myleshorton: Is this typedef really useful? Looks like it's just used in one place.

src/github.com/getlantern/flashlight/logging/useragents.go
```diff
@@ -0,0 +1,50 @@
...
+				userAgents[agent] = n + 1
+			} else {
+				userAgents[agent] = 1
+			}
+		}
+	}()
+}
+
+// getSessionUserAgents returns the
+func GetSessionUserAgents() string {
```
myleshorton: nit: I think this method can be package-private.

myleshorton:
A couple nits but otherwise LGTM!

src/github.com/getlantern/flashlight/logging/useragents.go
uaalto: I tend to use typedefs for long-ish definitions to clarify what they actually mean. In this case I agree it doesn't help much.

src/github.com/getlantern/flashlight/logging/useragents.go
uaalto: Sure! The reason I did it public is because of its symmetry with the other method, but is not going to be used outside logging in principle.

myleshorton:
LGTM!

PR #2964, opened 2015-08-17
Issue2470 suggestions

xiam:
Hey @fffw, just a few trivial fixes for your detour changes. Care to take a look?

Thanks.

### DIFF
--- a/src/github.com/getlantern/detour/whitelist.go
+++ b/src/github.com/getlantern/detour/whitelist.go
@@ -19,7 +19,9 @@ var (
 func AddToWl(addr string, permanent bool) {
 	muWhitelist.Lock()
 	defer muWhitelist.Unlock()
-	whitelist[addr] = wlEntry{permanent}
+	if add != "" {
+		whitelist[addr] = wlEntry{permanent}
+	}
 }
 
 //RemoveFromWl removes an addr from whitelist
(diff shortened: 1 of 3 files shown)


src/github.com/getlantern/detour/whitelist.go
```diff
@@ -19,7 +19,9 @@ var (
 func AddToWl(addr string, permanent bool) {
 	muWhitelist.Lock()
 	defer muWhitelist.Unlock()
-	whitelist[addr] = wlEntry{permanent}
+	if add != "" {
```
xiam: A small safety check. I realised some getParentDomain() parameters could return "" in some specific cases.

fffw: A bit unclear to me. `getParentDomain()` is used by `whitelisted()` but not here. `addr` will  never be "" except passed in intentionally.

myleshorton: This also doesn't compile right?

xiam: > `addr` will never be "" except passed in intentionally.

That's what I am trying to prevent, with this check the function will take care for itself, instead of relying on the context.

xiam: > This also doesn't compile right?

Right! fixed. Thanks.

fffw:
Great, thanks @xiam !

PR #2965, opened 2015-08-17
use new ip and country headers closes #2943

### LINKED ISSUE #2943: Use country and IP headers returned from ordinary flashlight servers

myleshorton:
Can someone have a look at this, maybe you @uaalto? It just skips our whole geolookup althogether, although now I'm realizing I need to add this in more places -- still work to do! Don't review yet!

### DIFF
--- a/src/github.com/getlantern/flashlight/geolookup/geolookup.go
+++ b/src/github.com/getlantern/flashlight/geolookup/geolookup.go
@@ -31,9 +31,15 @@ var (
 	service  *ui.Service
 	client   atomic.Value
 	cfgMutex sync.Mutex
-	country  atomic.Value
-	ip       atomic.Value
+	country  = atomicString()
+	ip       = atomicString()
 )
+
+func atomicString() atomic.Value {
+	var val atomic.Value
+	val.Store("")
+	return val
+}
 
 func GetIp() string {
 	return ip.Load().(string)
--- a/src/github.com/getlantern/flashlight/geolookup/geolookup.go
+++ b/src/github.com/getlantern/flashlight/geolookup/geolookup.go
@@ -36,19 +36,11 @@ var (
 )
 
 func GetIp() string {
-	c := ip.Load()
-	if c == nil {
-		return ""
-	}
-	return c.(string)
+	return ip.Load().(string)
 }
 
 func GetCountry() string {
-	c := country.Load()
-	if c == nil {
-		return ""
-	}
-	return c.(string)
+	return country.Load().(string)
 }
 
 // Configure configures geolookup to use the given http.Client to perform
@@ -58,6 +50,10 @@ func GetCountry() string {
 func Configure(newClient *http.Client) {
 	cfgMutex.Lock()
 	defer cfgMutex.Unlock()
+
+	// Avoid annoying checks for nil later.
+	ip.Store("")
+	country.Store("")
 
 	client.Store(newClient)
 
--- a/src/github.com/getlantern/flashlight/geolookup/geolookup.go
+++ b/src/github.com/getlantern/flashlight/geolookup/geolookup.go
@@ -96,9 +96,7 @@ func lookupIp(httpClient *http.Client) (string, string, error) {
 	var req *http.Request
 	var resp *http.Response
 
-	lookupURL := "http://nl.fallbacks.getiantem.org"
-
-	if req, err = http.NewRequest("HEAD", lookupURL, nil); err != nil {
+	if req, err = http.NewRequest("HEAD", "http://nl.fallbacks.getiantem.org", nil); err != nil {
 		return "", "", fmt.Errorf("Could not create request: %q", err)
 	}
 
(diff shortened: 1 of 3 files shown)


uaalto:
Sure, I will review it! Let me know here when it's ready for review.

myleshorton:
OK I think this is all set for now @uaalto. I think ideally we'd pass the geo data for all servers from the config side as well, which would eliminate all lookups of IPs other than our own. I opened that ticket at #2973

The last change (https://github.com/getlantern/lantern/commit/e49a4dadf42d604eabdf75ecaeb00c5d9efb65d2) to server.go is pretty important so the server side doesn't constantly do unnecessary lookups too -- should speed things up a bit on flashlight actually but also requires redeploying all the flashlight servers @aranhoide.

myleshorton:
Any luck with this one @uaalto?

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -31,23 +31,24 @@ var (
...
+	ip       atomic.Value
 )
 
-func GetLocation() *geolookup.City {
-	l := location.Load()
-	if l == nil {
-		return nil
+func GetIp() string {
+	c := ip.Load()
+	if c == nil {
```
uaalto: Can we avoid this check each time if we set the `ip` atomic value as "" from the beginning?

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -31,23 +31,24 @@ var (
...
 	}
-	return l.(*geolookup.City)
+	return c.(string)
 }
 
 func GetCountry() string {
-	loc := GetLocation()
-	if loc == nil {
+	c := country.Load()
+	if c == nil {
```
uaalto: (same here)

myleshorton:
Good question! Let me play with it.

myleshorton:
How's this @uaalto?

uaalto:
Great! :)
I'll continue reviewing this code.

myleshorton:
ooh -- this has issues from CI -- investigating!

uaalto:
Looks like a panic from a conversion precisely in the atomic values.

myleshorton:
OK this should fix it @uaalto thanks!

uaalto:
I think also better to initialize it there.

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -295,6 +295,11 @@ func (server *Server) checkForBannedCountry(req *http.Request) error {
 }
 
 func (server *Server) lookupCountry(req *http.Request) (string, error) {
+	// Use the country CloudFlare gives us if it's available.
+	cf := req.Header.Get("Cf-Ipcountry")
```
uaalto: Just making sure -- this could be done independently of the country/ip lookup above, right?
Here we are just trying to get the information via Cloudflare's filter, as a shortcut.

myleshorton:
Agreed, although it's interesting other code is trying to get that data
before its there -- seems like we might want to have everything listen for
the event.

uaalto:
Ok, I've done a full pass and looks good to me. But I'd like to give it a second pass tomorrow first thing in the morning with a fresh mind if that's ok.

myleshorton:
Sounds good!

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -88,6 +91,45 @@ func registerService() error {
...
+	if resp.StatusCode != http.StatusOK {
+		body := "body unreadable"
+		b, err := ioutil.ReadAll(resp.Body)
+		if err == nil {
+			body = string(b)
+		}
+		return "", "", fmt.Errorf("Unexpected response status %d: %v", resp.StatusCode, body)
+	}
+
+	ip := resp.Header.Get("Lantern-Ip")
```
uaalto: We are 100% positive that Enproxy will respond with these headers right? Even though a bit pedantic, wouldn't a test to make sure be a good idea? One rename or refactor could sneak empty strings here.

src/github.com/getlantern/flashlight/server/server.go
uaalto: The main reason I ask this is because I was thinking on potential issues with the encoding used for countries in CF-IPCountry. Are we confident both Cloudflare and us are using the same format? Cloudflare is using ISO 3166-1 Alpha 2 format, I checked that.

uaalto:
Ok, I've done a couple observations on the code, but I'm ready to merge when you take a look!

src/github.com/getlantern/flashlight/geolookup/geolookup.go
myleshorton: I added them in #2941 but yeah I agree a test would be good. I'll add it.

src/github.com/getlantern/flashlight/server/server.go
myleshorton: > this could be done independently of the country/ip lookup above, right?

It's actually always done independently -- i.e. this code is only currently run on flashlight servers. As far as I know there's only one standard for two letter country codes, so I think we're fine there. The prior lookup, however, also returns codes that are ultimately from the "Cf-Ipcountry" header (i.e. they're returned from Flashlight servers that are using that header), so there's no way for them to conflict.

myleshorton:
OK @uaalto I just added a test and comments.

myleshorton:
I'm going to go ahead and merge with that test now in there @uaalto, as it would be great to get this into 2.0.1!!

PR #2969, opened 2015-08-18
Add exponential backoff timeout to test

### LINKED ISSUE #2968: Test fail: An error is expected but got nil. There should have been a problem dialing
This is somewhat expected from changes to fix https://github.com/getlantern/lantern/issues/2858. The tests need to take into account that sometimes connections will succeed. The current approach of a random timeout will continually need tuning otherwise.

uaalto:
Closes #2968 

Adapts the test to reduce the timeout within a lower boundary to keep it reasonable. Without this, the test depend on the system, which will require continuous tuning and/or complicate reproducibility.

### DIFF
--- a/src/github.com/getlantern/tlsdialer/tlsdialer_test.go
+++ b/src/github.com/getlantern/tlsdialer/tlsdialer_test.go
@@ -205,11 +205,11 @@ func TestVariableTimeouts(t *testing.T) {
 	}
 
 	for i := 0; i < 500; i++ {
-		// The 5000 microseconds limit is arbitrary. In some systems this may be too high,
+		// The 3000 microseconds limit is arbitrary. In some systems this may be too high,
 		// leading to a successful connection and thus a failed test. On the other hand,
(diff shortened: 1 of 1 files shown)


myleshorton:
For some reason I thought the issue with the test was with the cert itself. Does it just appear that way but is actually an issue with the timeout @uaalto?

uaalto:
The previous error was caused by the connection succeeding and then trying to do the TLS handshake. Because we didn't provide any certificates in the test code (assuming it wasn't going to reach that stage), it was failing to use a proper certificate.
Once this was fixed, it turns out that _sometimes_ the the 5000 microseconds timeout is too high, and the connection actually succeeds. The test relies on the connections timing out. Since I couldn't replicate the issue after 1000 runs in my machine, I assumed that it depends a lot on the machine, so I propose a way to adapt to faster machines.
The code will just reduce the max value of the timeout and repeat the test. It will do this until a connection succeeded in a suspiciously low time, which will be logged for further investigation.

myleshorton:
Gotcha -- OK let's give it a shot!

PR #2978, opened 2015-08-18
don't try to respond with bad gateway after OK closes #2977

### LINKED ISSUE #2977: Data race in client/handler.go between responding with OK and bad gateway
In practice we'll pretty much always respond with OK first.

```
==================
WARNING: DATA RACE
Write by goroutine 122:
  github.com/getlantern/flashlight/client.(*Client).intercept()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/client/handler.go:103 +0x942
  github.com/getlantern/flashlight/client.(*Client).ServeHTTP()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/client/handler.go:30 +0x21e
  net/http.serverHandler.ServeHTTP()
      /Users/afisk/go_native/src/net/http/server.go:1703 +0x1f6
  net/http.(*conn).serve()
      /Users/afisk/go_native/src/net/http/server.go:1204 +0x1087

Previous write by goroutine 139:
  github.com/getlantern/flashlight/client.func·008()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/client/handler.go:83 +0xd6

Goroutine 122 (running) created at:
  net/http.(*Server).Serve()
      /Users/afisk/go_native/src/net/http/server.go:1751 +0x3cd
  github.com/getlantern/flashlight/client.(*Client).ListenAndServe()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/client/client.go:81 +0x618
  main.runClientProxy()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/flashlight.go:312 +0x79c

Goroutine 139 (finished) created at:
  github.com/getlantern/flashlight/client.(*Client).intercept()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/client/handler.go:90 +0x71c
  github.com/getlantern/flashlight/client.(*Client).ServeHTTP()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/client/handler.go:30 +0x21e
  net/http.serverHandler.ServeHTTP()
      /Users/afisk/go_native/src/net/http/server.go:1703 +0x1f6
  net/http.(*conn).serve()
      /Users/afisk/go_native/src/net/http/server.go:1204 +0x1087
==================
```

myleshorton:
Can you also take a look at this one @uaalto?

### DIFF
--- a/src/github.com/getlantern/flashlight/client/handler.go
+++ b/src/github.com/getlantern/flashlight/client/handler.go
@@ -100,10 +100,6 @@ func (client *Client) intercept(resp http.ResponseWriter, req *http.Request) {
 	} else {
 		connOut, err = detour.Dialer(d)("tcp", addr)
 	}
-	if err != nil {
-		respondBadGateway(clientConn, fmt.Sprintf("Unable to handle CONNECT request: %s", err))
-		return
-	}
 
 	if <-success {
 		// Pipe data between the client and the proxy.


uaalto:
This change is correct. I think this change fixes the situation, which currently has the case where a badGateway is responded after the OK. That is definitely a bug. We shouldn't respond BadGateway after the OK.
Alternatively, we can also check for a failed OK and reply with badGateway if that's the case.

myleshorton:
> Alternatively, we can also check for a failed OK and reply with badGateway if that's the case.

I *think* the only time the OK would fail in practice is if the client is no longer there though, which would of course mean the bad gateway would fail too. Does that make sense?

uaalto:
I agree that it is very unlikely that the OK would fail. Maybe with some setup like lossy local WiFi where another device uses Lantern as a proxy. This is the sort of situation we talked recently for mobile phones in local networks.
But in any case, what I think could happen is that we fail in `connOut, err = detour.Dialer(d)("tcp", addr)`, and it happens before the OK. At the moment we are not retrying, so we could answer badGateway in that situation and cancel the connection.

On the other hand, to simplify things, I think it would be ok to respond OK and close the connection. This would leave the browser/user-agent to deal with it.

uaalto:
So, if you agree with last point, I merge =)

myleshorton:
Yeah I agree it's theoretically possible, but I don't think it's quite worth going to the trouble. We could create another sync.Once method that responds and maybe preferences one response code over the other with a flag, but in practice I think the OK would only fail if the connection were actually lost and that in any case it would almost always beat the bad gateway call. So think the fact that we close the connection even after the OK should be enough for the browser.

uaalto:
I agree. LGTM, merging!

myleshorton:
Thanks @uaalto!

PR #3031, opened 2015-08-25
Report connection errors to InfluxDB closes #2985

### LINKED ISSUE #2985: Report failures connecting each chained server
As another step of https://github.com/getlantern/lantern/issues/2884, we can anonymously send failures to InfluxDB, so we can have a sense on which server is performing bad. Sometimes, server looks good but just doesn't serve traffics well.

fffw:
Closes https://github.com/getlantern/lantern/issues/2985

Idealy we should report periodically, say, 1 minute, but I'd like to postpone it to when we also want to tracks connections, delay and throughput.


fffw:
Sample graphs at https://dashboard.getlantern.org/dashboard/db/lantern-client-errors . We should not expose server ip addresses unless we make the page password protected, but it's ok for now.

myleshorton:
Awesome@fffw! Just one thing. I think it might make sense to make this even simpler and to specifically report errors from line 99 in balancer.go -- https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/balancer/balancer.go#L99 

Those are specifically errors even connecting.

fffw:
@myleshorton It's great to have `measured` to be plugged in a single point, but two concerns prevents me from doing it.

1. We only have (much longer) labels instead of server addresses at `balancer.go`. InfluxDB is not good at string manipulation at this moment.

2. `balancer` is an independant library that is not logically related to `measured`. `flashlight/chained` and `flashlight/fronted` is dedicated to our application, which is good to use any library.

src/github.com/getlantern/flashlight/measured/measured.go
```diff
@@ -0,0 +1,43 @@
...
+// and reports them to statshub via HTTPS post
+func Configure(cfg *Config, httpClient *http.Client) {
+	measured.Stop()
+	measured.Reset()
+	measured.AddReporter(reporter.NewInfluxDBReporter(cfg.InfluxURL,
+		cfg.InfluxUsername,
+		cfg.InfluxPassword,
+		"lantern",
+		httpClient))
+	country := geolookup.GetCountry()
```
myleshorton: GetCountry is unlikely to be populated at the time this is called. You could use the pubsub package along the lines of https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/analytics/analytics.go#L28 however.

myleshorton:
> @myleshorton It's great to have measured to be plugged in a single point, but two concerns prevents me from doing it.
> 
> We only have (much longer) labels instead of server addresses at balancer.go. InfluxDB is not good at string manipulation at this moment.
> 
> balancer is an independant library that is not logically related to measured. flashlight/chained and flashlight/fronted is dedicated to our application, which is good to use any library.

OK that makes sense @fffw. I'm mostly worried about noise, and I wonder if initially it might make sense to either only included data for chained or to completely separate out chained from fronted data. Both can get things like EOF or timeouts, but we're already quite aware of a lot of the fronted errors but have less visibility I think into the chained errors. Would it be possible to separate them?

src/github.com/getlantern/flashlight/measured/measured.go
fffw: Good catch @myleshorton !

fffw:
We can filter out fronted errors by the special server name "*.fallbacks.getiantem.org"

fffw:
Modified as per your advice @myleshorton, would you mind take a look again?

myleshorton:
Hey I mostly want to just run Lantern more with this one @fffw, but hopefully this will go out along with the detour changes in 2.0.3.

PR #2989, opened 2015-08-19
Remove HTTP config server closes #2988

myleshorton:
This just simplifies things a bit more. I think @uaalto you and I are the main people primarily focused on the Lantern go backend side right now, so assigning over to you!

PR #3033, opened 2015-08-25
Fix dsp and yamlconf test

fffw:


src/github.com/getlantern/yamlconf/yamlconf_test.go
```diff
@@ -214,69 +213,6 @@ func TestCustomPoll(t *testing.T) {
 	}, updated, "Custom polled config should contain correct data")
 }
 
-func TestConfigServer(t *testing.T) {
```
fffw: The code this test targets was removed at https://github.com/getlantern/lantern/pull/2989

src/github.com/getlantern/peerscanner/dsp/dsp_test.go
```diff
@@ -39,6 +40,8 @@ func TestRegister(t *testing.T) {
 		assert.NotNil(t, rec, "A new record should have been returned")
 	}
 
+	// give dnssimple enough time to complete the operation
+	time.Sleep(100 * time.Millisecond)
```
uaalto: I see that you've tried with 100 millis and 1 sec. In order to make it resilient, could we somehow detect success on the DNSSimple side to continue the test? Or maybe don't consider it failed if the fail was on DNSSimple side?

fffw: Ha I forgot to try run the test locally before submit code. Line 40 already verified that a new record is created, 100 millis is for DNSimple to sync internal state which should be very short I think.

uaalto:
LGTM! Thanks @fffw, having clean tests finally!

PR #3042, opened 2015-08-26
add projecttalk link

manukall:
this adds a badge link to the ProjectTalk message board for lantern.

i think this could help make the issues more useful by moving a lot of the discussion that is not development related to the message board.

additionally you can use the "Moved to ProjectTalk" label on issues that should be discussed outside of github. those will automatically get imported to ProjectTalk and a comment with a link will be posted to the issue.


uaalto:
This is interesting @manukall, thank you!
We've been receiving a lot of *support-related* issues on github, which is distracting. It seems to have stopped for now though. The problem of this is keeping track of yet another site with issue threads will bring more overhead to the team I think. 
I'll discuss it with the team. Thanks again!

manukall:
@uaalto thanks. it's great to hear you like the idea.

ProjectTalk is very much just at  the beginning. Right now, [Semantic-UI](https://github.com/Semantic-Org/Semantic-UI) is the only larger project actively using it. We're still figuring out what works best, but I believe that in general it will be very helpful to remove some of the distraction you mentioned.

One idea behind moving the support-related issues to a message board is that it will also encourage other users to GIVE support. While issues on Github can feel more directed towards the projects maintainers, people understand message boards as community-driven. This way, we hope to decrease the overhead for maintainers.

Anyway, as I said, we're right at the beginning. So I'm grateful for any feedback you have. If you can think of a way of improving ProjectTalk, please let me know.

PR #3065, opened 2015-08-29
Adding support for docker-machine since b2d is deprecated

JulianGindi:
The `default` docker-machine instance is the one the automated installer creates for you on install, so I thought I'd stick with that. We can also make that a variable at the top of the Makefile to make changing it easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ RUBY := $(shell which ruby 2> /dev/null)
 APPDMG := $(shell which appdmg 2> /dev/null)
 SVGEXPORT := $(shell which svgexport 2> /dev/null)
 
-BOOT2DOCKER := $(shell which boot2docker 2> /dev/null)
+DOCKERMACHINE := $(shell which docker-machine 2> /dev/null)
 
 GIT_REVISION_SHORTCODE := $(shell git rev-parse --short HEAD)
 GIT_REVISION := $(shell git describe --abbrev=0 --tags --exact-match 2> /dev/null || git rev-parse --short HEAD)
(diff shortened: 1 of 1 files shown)


JulianGindi:
What is the best way to test this locally? Can I test within a docker container? Hard to debug this build issue without such a system.

xiam:
Hello @JulianGindi,

`make binaries` should be enough to test if everything is working OK with the docker machines we use.

JulianGindi:
Awesome, thanks for the tip. I was able to build all the binaries successfully...was wondering if anyone had more information about why this build might be failing on travis.

JulianGindi:
Seems like a few other PRs from personal branches (not the main repo), are running into the same travis build issue I am. Maybe something to look into. Is this the expected way for people to contribute, or should I be sending patches to the mailing list? Am more than happy to help out in any way.

uaalto:
Hi @JulianGindi 
I guess the reason why you are getting this issue is because we rely on private keys for some part of the process, and those need to be specifically set up per account on Travis.
I don't see an immediate solution to this apart of letting the tests fail or deactivate the offending ones. You can probably remove https://github.com/getlantern/lantern/blob/valencia/.travis.yml#L10 and deactivate tests that rely on this.

I'll think of ways this can be improved in the future, we have however many priorities now in Lantern that may require immediate attention.

Now, about your PR, what is the error/issue you are actually solving in this commit?

Thank you very much for contributing!

uaalto:
Closing for now, as there is no response and/or apparent reason for this change, which seems to break builds.

PR #3067, opened 2015-08-30
Adding missing "yum clean all" to prevent saving unnecessary cache fi…

xiam:
…les across layers.


myleshorton:
LGTM thanks @xiam!

PR #3068, opened 2015-08-30
use tip, lastest version of go

bitdop:
I have also added allow_failures via matrix for tip, so that failures building on tip are considered allowed failures that way you can still see and anticipate issues, but breaking unstable changes on tip don't mess up other builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: go
 go:
 - 1.4.2
+- tip
+matrix:
+  allow_failures:
+    - go: tip
 install:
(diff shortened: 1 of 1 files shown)


uaalto:
Thank you @bitdop 
We are currently using 1.4.2 for development (and looking forward moving to 1.5), why would we test with tip? The aim of testing is to get an environment as close as possible to users, and we release based on the 1.4.2 runtime.

uaalto:
Closing. This can bring some issues into the testing environment.

PR #3070, opened 2015-08-30
Don't stick to single masquerade

fffw:
Main changes of this PR:

1. Check masquerades using direct domain fronting against http://geo.getiantem.org/lookup, instead of full domain fronting against Google. This prevents the 403 error. Ideally we can setup a static page as test target which is always cached by CF, though for now, geo-server's load is very light.

2. Instead of using fixed http.Client, subcomponents will request a new direct domain fronter each time. This prevents a failed masquerade from keep failing.

fronted_test.go is failed due to 403 errors. I choose to leave it as a indicator to tell us if CF recovered or not.

With this change, Lantern can update config 1 minutes after it starts up. geolookup still not work, as it access `http://nl.fallbacks.getiantem.org` to gain ip and country. We can't move it to `geo.getiantem.org/lookup` unless it add a field of requester ip address in the response.

### DIFF
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -32,7 +32,7 @@ import (
 	"github.com/getlantern/flashlight/statreporter"
 	"github.com/getlantern/flashlight/statserver"
 	"github.com/getlantern/flashlight/ui"
-	"github.com/getlantern/flashlight/util"
+	//"github.com/getlantern/flashlight/util"
 
 	"github.com/mitchellh/panicwrap"
 )
@@ -358,9 +358,9 @@ func applyClientConfig(client *client.Client, cfg *config.Config) {
 	} else {
 		// Give everyone their own *http.Client that uses the highest QOS dialer. Separate
 		// clients for everyone avoids data races configuring those clients.
-		config.Configure(hqfd.NewDirectDomainFronter())
-		geolookup.Configure(hqfd.NewDirectDomainFronter())
-		statserver.Configure(hqfd.NewDirectDomainFronter())
+		config.Configure(hqfd)
+		geolookup.Configure(hqfd)
+		statserver.Configure(hqfd)
 		// Note we don't call Configure on analytics here, as that would
 		// result in an extra analytics call and double counting.
 	}
@@ -425,12 +425,13 @@ func runServerProxy(cfg *config.Config) {
 }
 
 func updateServerSideConfigClient(cfg *config.Config) {
-	client, err := util.HTTPClient(cfg.CloudConfigCA, "")
+	// temporarily comment it as we didn't have fronted dialer on server side
+	/*client, err := util.HTTPClient(cfg.CloudConfigCA, "")
 	if err != nil {
 		log.Errorf("Couldn't create http.Client for fetching the config")
 		return
 	}
-	config.Configure(client)
+	config.Configure(client)*/
 }
 
 func useAllCores() {
--- a/src/github.com/getlantern/fronted/dialer.go
+++ b/src/github.com/getlantern/fronted/dialer.go
@@ -41,7 +41,7 @@ var (
 type Dialer interface {
 	proxy.Dialer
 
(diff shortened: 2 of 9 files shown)


fffw:
@myleshorton This is my proposal to overcome the CF 403 issue.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -358,9 +358,9 @@ func applyClientConfig(client *client.Client, cfg *config.Config) {
 	} else {
 		// Give everyone their own *http.Client that uses the highest QOS dialer. Separate
 		// clients for everyone avoids data races configuring those clients.
```
aranhoide: Doesn't the change below obsolete the comment above?

src/github.com/getlantern/fronted/dialer.go
```diff
@@ -183,14 +183,15 @@ func (d *dialer) Close() error {
...
-	enproxyConfig := d.enproxyConfigWith(func(addr string) (net.Conn, error) {
-		return d.dialServerWith(masquerade)
-	})
-
 	return &http.Client{
-		Transport: &http.Transport{
-			Dial: func(network, addr string) (net.Conn, error) {
-				return enproxy.Dial(addr, enproxyConfig)
+		Transport: &DirectDomainTransport{
+			Transport: http.Transport{
```
aranhoide: [not important] I stay away from the lantern repo code for a while and I forget about the relationship between clients, dialers, transports etc.- transports have transports inside?  :S

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -425,12 +425,13 @@ func runServerProxy(cfg *config.Config) {
 }
 
 func updateServerSideConfigClient(cfg *config.Config) {
-	client, err := util.HTTPClient(cfg.CloudConfigCA, "")
+	// temporarily comment it as we didn't have fronted dialer on server side
+	/*client, err := util.HTTPClient(cfg.CloudConfigCA, "")
```
myleshorton: We don't need a fronted dialer on the server side though do we @fffw?

src/github.com/getlantern/fronted/masquerade.go
```diff
@@ -138,7 +138,7 @@ func (vms *verifiedMasqueradeSet) doVerify(masquerade *Masquerade) bool {
 	go func() {
 		start := time.Now()
 		httpClient := vms.dialer.HttpClientUsing(masquerade)
-		req, err := http.NewRequest("HEAD", "http://www.google.com/humans.txt", nil)
+		req, err := http.NewRequest("HEAD", "http://geo.getiantem.org/lookup", nil)
```
myleshorton: Check out my changes in https://github.com/getlantern/lantern/pull/3073 @fffw

It turns out it's not the URL that triggers the 403 but the enproxy headers!

src/github.com/getlantern/flashlight/flashlight.go
myleshorton: I would just delete the comment above probably.

myleshorton:
Thanks @fffw I somehow missed you had submitted this PR until today, but I think our PRs are almost entirely complementary. I would vote for merging yours in followed by mine, likely with a conflict or two.

src/github.com/getlantern/flashlight/flashlight.go
fffw: No we don't need it. As I changed `config.Configure` to accept `fronted.Dialer` as parameter, the call to it here becomes invalid. We can pass the `http.Client` as another parameter, but it makes code less understandable.

Extract those code to a common call like `nextHTTPClient` would be good to hide the difference.

src/github.com/getlantern/fronted/dialer.go
fffw: It is what `NewDirectDomainFronter()` doing previously. `DirectDomainTransport` is a simple wrapper of `http.Transport`

myleshorton:
I think we can actually close this one after all right @fffw? Feel free to re-open if you think this is still useful.

PR #3079, opened 2015-09-01
Add cache control header and fix storing of etag closes #3076

### LINKED ISSUE #3076: config: not store etag until yaml sucessfully parsed
In rare circumstance, config-server returned yaml might be malformed, which triggers below error:

```
Unable to unmarshal YAML for update: YAML error: line 2: mapping values are not allowed in this context
```

As etag already set, successive requests will always 304 not modified, hence leaves the default config unchanged.

myleshorton:
Slightly less risky version of this change.


myleshorton:
here's a pared-down version @aranhoide

myleshorton:
Hey I'm just going to merge this one in @aranhoide  and @fffw to get this version out in the field!!

Issue #3077, opened 2015-09-01
Lantern doesn't run on XP (sometimes)


uaalto:
Seems caused by **panicwrap** package. Getting rid of it seems to fix it. There are two options here:

* Remove panicwrap usage under Windows XP (not very satisfactory given the amount of users that run this OS).
* Try to fix panicwrap itself (may be due to an OS limitation or a Go bug, or very time-consuming to catch).

fffw:
I built two versions of 2.0.0 with and without `-Hwindowsgui` respectively, and run them on an XP SP3 vm. The one without console always starts up and works well, will another one never starts up. I didn't see latter in task manager.

Both binaries can run on Windows 7 without issue. Binary without `panicwrap` works well on XP.

It's not easy to spot the error returned by `panicwrap` though.

uaalto:
I have made some progress on this issue. Apparently is two different issues at once. Panicwrap is not the cause. I'll open another issue for the bug I've identified, and try to verify the second part.

uaalto:
I forgot to tell that the bug can be triggered with and without panicwrap and/or windowsgui flag.

uaalto:
The first issue, referenced in #3086 was:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x53633b]

goroutine 484 [running]:
sync/atomic.LoadUint64(0x117acec4, 0x53aa1a, 0xc)
        /go/src/sync/atomic/asm_386.s:165 +0xb
github.com/getlantern/detour.(*Conn).anyDataReceived(0x117aceb0, 0x117c8f40)
        /lantern/src/github.com/getlantern/detour/detour.go:199 +0x32
github.com/getlantern/detour.func┬╖008()
        /lantern/src/github.com/getlantern/detour/detour.go:145 +0xf2
created by github.com/getlantern/detour.func┬╖010
        /lantern/src/github.com/getlantern/detour/detour.go:150 +0x63f

goroutine 1 [syscall, locked to thread]:
syscall.(*Proc).Call(0x1162d980, 0x115dff74, 0x2, 0x2, 0x7565a0, 0x8d3134, 0x0, 0x0)
        /go/src/syscall/dll_windows.go:132 +0x27b
syscall.(*LazyProc).Call(0x11557a60, 0x115dff74, 0x2, 0x2, 0x0, 0x0, 0x0, 0x0)
[...]

PR #3087, opened 2015-09-02
Fix atomic uint64 alignment issue that occurs on x86-32 and ARM

### LINKED ISSUE #3086: Sporadic nil pointer dereference in detour
```
Sep 02 18:28:17.872 - 0m3s ERROR flashlight: flashlight.go:101 panic: runtime error: invalid memory address or nil pointer dereference

[signal 0xc0000005 code=0x0 addr=0x0 pc=0x537ebb]

goroutine 538 [running]:
sync/atomic.LoadUint64(0x11d85784, 0x53c59a, 0xc)
        /go/src/sync/atomic/asm_386.s:165 +0xb
github.com/getlantern/detour.(*Conn).anyDataReceived(0x11d85770, 0x11d838c0)
        /lantern/src/github.com/getlantern/detour/detour.go:200 +0x32
github.com/getlantern/detour.func·008()
        /lantern/src/github.com/getlantern/detour/detour.go:146 +0xf2
created by github.com/getlantern/detour.func·010
        /lantern/src/github.com/getlantern/detour/detour.go:151 +0x63f

```

uaalto:
Closes #3086 

This bug was discovered as part of the on-going investigation for https://github.com/getlantern/lantern/issues/3077.
The rationale can be found here: https://golang.org/pkg/sync/atomic/#pkg-note-BUG

### DIFF
--- a/src/github.com/getlantern/detour/detour.go
+++ b/src/github.com/getlantern/detour/detour.go
@@ -56,6 +56,11 @@ var (
 // Conn implements an net.Conn interface by utilizing underlie direct and
 // detour connections.
 type Conn struct {
+	// Keeps track of the total bytes read from this connection, atomic
+	// Due to https://golang.org/pkg/sync/atomic/#pkg-note-BUG it requires
+	// manual alignment. For this, it is best to keep it as the first field
+	readBytes uint64
+
 	// The underlie connections, uses buffered channel as ring queue to avoid
 	// locking. We have at most 2 connetions so a length of 2 is enough.
 	conns chan conn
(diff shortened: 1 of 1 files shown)


atavism:
LGTM.. So unless that readBytes field is the first member of the struct, it isn't aligned on a 64-bit boundary which can cause that atomic LoadUint64 operation to fail on ARM, right? Nice debugging.

uaalto:
Yes. It happens in x86 32 bits too. Actually is not necessary that is the first member of the struct, but that is aligned. The easiest way is to put it first, so it dosn't break if someone adds new fields and doesn't realize of this.

PR #3091, opened 2015-09-03
[important] Workaround for issue #3077

uaalto:
Issue #3077 is confirmed to be triggered by panicwrap trying to fork a GUI app. This seems to offend Windows XP, due to invalid standard file descriptors. Unfortunately we didn't see any log of this because panicwrap fails before logging is configured. Additionally, this bug isn't detected because it doesn't happen if the application is run in console mode (which is generally used for development).

A better solution could be researched further, although I'm not certain it is viable yet, but I wanted to get this ASAP in valencia, since it could potentially affect all Windows XP users.

### DIFF
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -113,11 +113,15 @@ func main() {
 	if err != nil {
 		// Something went wrong setting up the panic wrapper. This won't be
 		// captured by panicwrap
+		// At this point, continue execution without panicwrap support. There
+		// are known cases where panicwrap will fail to fork, such as Windows
+		// GUI app
 		log.Errorf("Error setting up panic wrapper: %v", err)
-	}
-	// If exitStatus >= 0, then we're the parent process.
-	if exitStatus >= 0 {
-		os.Exit(exitStatus)
+	} else {
+		// If exitStatus >= 0, then we're the parent process.
+		if exitStatus >= 0 {
+			os.Exit(exitStatus)
+		}
 	}
 
 	parseFlags()


fffw:
Consider the amount of users this bug affected, I'd suggest to merged this
fix into 2.0.2 and update the installers. How do you think @uaalto
and @myleshorton ?

uaalto:
Yes, and [this issue](https://github.com/getlantern/lantern/issues/3086) is also critical, it produces a random segfault.

myleshorton:
Yeah I think we should probably get this in, but keep in mind XP is only around 5% of our users (at least of the visitors to our web site) so not as urgent as it might otherwise be. I think the general connectivity issue is much more pressing right now.

uaalto:
Oh, I thought the number of XP users was higher! This OS version gives a lot headaches, the firewall thing was an example too.

fffw:
https://github.com/getlantern/lantern/issues/3086 is triggered by new
detour which is not in any release, so that won't be a problem.

I do see several Chinese user complaints that Lantern 2.0.2 can run for the
first time (started by installer) but can't run again anymore. All of the
reported OS (if reported any) is XP. I verified it just now. It should be
caused by the `panicwrap` issue.

myleshorton:
Hey I think we might as well put this in 2.0.2. Would you mind doing a PR to the release-2.0.1 branch @uaalto?

uaalto:
Done. I'm very happy I could finally use cherry-pick for a perfect case like this :)
I think we still have to merge this one, right? If it is a specific release branch, even if they are the same commit it needs to be merged back in valencia.

myleshorton:
Yeah either way @uaalto. The release branch will be merged back into valencia will a bunch of things again.

PR #3093, opened 2015-09-03
datacenter based on locale and better tests for config copying closes #3092

### LINKED ISSUE #3092: Lantern not working well for Iranian users.

myleshorton:


fffw:
Great catch @myleshorton  LGTM. I'm testing on Iran VPS to see if it actually resolve the always `time out' issue.

myleshorton:
Cool. Are you able to reproduce the timeout issue on the VPS? It was working fine for me on the Windows one, but I haven't checked the others -- that's a very good idea though I should.

fffw:
From testing on Iran VPS, Lantern can get `nl` fronted server, but all masquerade tests, all fronted and chained servers are still timed out, though the `nl` server can even be verified from the Chinese box. Checking what is happening.

myleshorton:
Interesting @fffw. Thanks. That's definitely what users are seeing too.

myleshorton:
You're using one of the ubuntu boxes right @fffw?

fffw:
Yes ubuntu box @myleshorton . Code from this PR still get same result, that's unexpected.

myleshorton:
Yeah weird. I just sent it to a user and heard the same back. At least we're able to reproduce it now though. That's huge.

myleshorton:
OK @aranhoide would you mind taking a quick look at this? I'm just stripping out the embedded fallbacks entirely for this one.

myleshorton:
I'm actually not sure this is working at all -- still verifying.

myleshorton:
Weird. So these IPs work in genconfig but not when actually running Lantern. Very odd. Still investigating.

aranhoide:
Why are you skipping fallbacks?  In order to test this, or do you want to release this without default chained servers?

myleshorton:
Yeah that's what I was thinking @aranhoide. They appear to be blocked in Iran and overloaded when they're not blocked. So the thought is to use DDF to get your config and use DF prior to that.

myleshorton:
Sorry I was thinking of releasing without default chained servers at all.

myleshorton:
I think I might have to change genconfig to do exactly what we're actually doing in the client. Odd there would be a difference, but it almost looks like there might be some DPI going on that's catching something unique.

aranhoide:
Yes, well, apparently genconfig doesn't try and domain front at all, just connect and grab certs?

myleshorton:
Yeah exactly. Maybe it's using SNI?

aranhoide:
I'm not sure why that matters.  IIUC, the point is that genconfig should try and proxy (or DDF) through these domains, to make sure they're actually Cf-backed and frontable.

aranhoide:
Taking a step back... *if* fronting is blocked by any means other than blacklisting the IPs we're using then the IP scanner won't help.

OTOH, if fronting depends on "secret" knowledge (that is, the set of masquerades that work), then it offers no blocking advantages over chained proxying (other than them looking different DPI wise).

myleshorton:
Hey @aranhoide, @uaalto, @fffw can one of you take a look at this PR so we can merge and get 2.0.3 out the door? Thanks!

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -309,6 +313,27 @@ func (cfg *Config) ApplyDefaults() {
...
+		localeTerritory = "us"
+	}
+	log.Debugf("Locale territory: %v", localeTerritory)
+	return defaultRoundRobinForTerritory(strings.ToLower(localeTerritory))
+}
+
+// defaultDataCenter customizes the default data center depending on the user's locale.
+func defaultRoundRobinForTerritory(localeTerritory string) string {
+	datacenter := ""
+	if localeTerritory == "cn" {
```
uaalto: I saw in the tests you used uppercase and here you are only testing for lowercase. Is there any possibility that the input is uppercase?

myleshorton: Yeah it'll actually generally be upper case from ```jibber_jabber.DetectTerritory()```

We lower case it in defaultRoundRobin in ```return defaultRoundRobinForTerritory(strings.ToLower(localeTerritory))``` but you're right we really should lowercase it in the method itself -- I'll make that change thanks!

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -98,31 +100,33 @@ func majorVersion(version string) string {
 // copyNewest is a one-time function for using older config files in the 2.x series.
 // from 2.0.2 forward, Lantern will consider all major versions to be compatible and
 // will name them accordingly, as in "lantern-2.yaml".
-func copyNewest(file string) {
+func copyNewest(file string, existsFunc func(file string) (string, bool)) string {
```
uaalto: So, the main purpose of passing this function as a parameter is testability, am I correct?

src/github.com/getlantern/flashlight/genconfig/README.md
```diff
@@ -0,0 +1,5 @@
+# Updating the global configuration
```
uaalto: Oh, I was writing this from scratch. Don't worry, I'll merge them later on.

src/github.com/getlantern/flashlight/config/config.go
myleshorton: yes exactly

myleshorton:
I just made the lower case change @uaalto

src/github.com/getlantern/flashlight/genconfig/README.md
myleshorton: Feel free to delete mine or just do whatever you like -- I just threw it in

uaalto:
Great! I think is good to make sure.
I finished reviewing the first part. About the genconfigWithVerify: is mostly based on genconfig, with the intention of *NOT* checking against custom Root CAs,  right? I can see that you commented out the code checking certificates and generates only masquerades.

myleshorton:
Oh, I would actually just ignore that file. We're not using it for anything right now, but it should be useful in the future for sketching out a more decoupled way of doing direct domain fronting.

The purpose of it was to actually verify proxying through the IPs we're able to connect to, but it turns out that's currently not necessary.

myleshorton:
We could actually delete it, but we want a version around somewhere for future refactoring. I could just keep it locally...

uaalto:
OK! I'm not sure what's best actually. I guess is good not to clutter if we are not going to use it for now, but maybe can be kept with a comment saying that it's WIP or something. 
The rest LGTM. I'm ready to merge it, unless you want to do something about the genconfigWithVerify program.

uaalto:
oh, by the way. Did we run errcheck, and the tests? Did we also test this code live somehow (I missed this part)?

uaalto:
I think [this commit](https://github.com/getlantern/lantern/commit/4cc077a1639950e2ab2841df2ed19c982be2ca0d) aranhoide just did is based on genconfigWithVerify, so maybe we can remove it now.

myleshorton:
Yup I'll just delete it @uaalto!

myleshorton:
I didn't run errcheck or the tests @uaalto but have been testing live a bunch. I'll run those

uaalto:
ok LGTM! Thanks @myleshorton

myleshorton:
Thanks for the review @uaalto!

PR #3095, opened 2015-09-04
Windows XP issue PR for 2.0.1

### LINKED ISSUE #3077: Lantern doesn't run on XP (sometimes)

uaalto:
Closes #3077


fffw:
@uaalto This PR is still trying to merge to `valencia`, rather than `release-2.0.1`.

PR #3116, opened 2015-09-09
balancer: not exclude used dialers when retry

fffw:
It's another small fix  in addition to yours @myleshorton to prevent "No dialers left to try" error when only 1 dialer available.

### DIFF
--- a/src/github.com/getlantern/balancer/dialer.go
+++ b/src/github.com/getlantern/balancer/dialer.go
@@ -109,8 +109,7 @@ func (d *dialer) start() {
 }
 
 func (d *dialer) isActive() bool {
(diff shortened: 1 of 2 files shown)


src/github.com/getlantern/balancer/dialer.go
```diff
@@ -109,8 +109,7 @@ func (d *dialer) start() {
 }
 
 func (d *dialer) isActive() bool {
-	//return atomic.LoadInt32(&d.active) == 1
-	return true
+	return atomic.LoadInt32(&d.active) == 1
```
fffw: Disable `isActive` check in balancer.go to make the intention more explicit.

myleshorton:
Nice -- I like it @fffw!

PR #3117, opened 2015-09-09
use p12 file for package-windows which is already in secrets repo

fffw:
The previous one, `bns.pfx`, doesn't exist in secrets folder. Changed to `bns_cert.p12` which is the same format.


myleshorton:
That one is the one that's signed with a sha-1

PR #3121, opened 2015-09-10
Issue3106 fix linux

fffw:
@myleshorton Can you have a look?

It fixes a few issues that prevents fresh installed Lantern from starting, on all platforms I think.


installer-resources/chained.yaml
```diff
@@ -1,3 +1,4 @@
+chainedservers:
```
fffw: Without this line, `len(servers.ChainedServers)` will always be zero when reading from chained.yaml

installer-resources/linux/lantern.sh
```diff
@@ -38,6 +41,7 @@ if [ ! -f $LANTERN_USER_BINARY ]; then
   mkdir -p $LANTERN_USER_DIRECTORY/bin
   cp $LANTERN_SOURCE_BINARY $LANTERN_USER_BINARY
   cp $LANTERN_SOURCE_YAML $LANTERN_USER_DIRECTORY
+  cp $LANTERN_CHAINED_YAML $LANTERN_USER_DIRECTORY
```
fffw: We forgot to copy the new file.

installer-resources/chained.yaml
myleshorton: We don't actually use the checked in copy - the copy we built with has that
line.

src/github.com/getlantern/flashlight/client/bootstrap.go
```diff
@@ -147,9 +146,11 @@ func bootstrapPath(fileName string) (string, error) {
 			yamldir = dir + "/../Resources"
 		}
 	} else if runtime.GOOS == "linux" {
-		yamldir = dir
+		yamldir = dir + "/../"
```
fffw: On linux, executable is started from `~/.lantern/bin`, and the config file is placed at `~/.lantern`

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -144,6 +144,9 @@ func copyGoodOldConfig(configDir, configPath string) {
 	}
 
 	for _, file := range files {
+		if file.IsDir() {
```
fffw: Skip directories to avoid noise.

installer-resources/chained.yaml
fffw: Ok I see. So my PR only fixes the startup issue on linux.

myleshorton:
LGTM!

fffw:
Thanks @myleshorton ! I'll fix the failed tests later.

PR #3128, opened 2015-09-10
Cfr masquerades

aranhoide:
This branch disables everything related to Cloudflare domain fronting and introduces a bunch of CloudFront masquerades.  We have a scanner running so we'll get lots more of these.

These masquerades will be used to hit https://config.getiantem.org/chained.yaml.gz via DDF, if and when nothing else works.

This PR is for early review.  I wouldn't merge this until we have the whole thing working.


src/github.com/getlantern/flashlight/genconfig/genconfig.bash
```diff
@@ -1,2 +1,2 @@
 #!/usr/bin/env bash
-go run genconfig.go -blacklist="blacklist.txt" -domains="domains.txt" -proxiedsites="proxiedsites" -fallbacks="fallbacks.json"
+go run genconfig.go -blacklist="blacklist.txt" -masquerades="masquerades.txt" -proxiedsites="proxiedsites" -fallbacks="fallbacks.json"
```
myleshorton: Much clearer - that's been annoying me for awhile!

src/github.com/getlantern/flashlight/genconfig/genconfig.go
```diff
@@ -32,8 +33,8 @@ const (
 
 var (
 	help            = flag.Bool("help", false, "Get usage help")
-	domainsFile     = flag.String("domains", "", "Path to file containing list of domains to use, with one domain per line (e.g. domains.txt)")
-	blacklistFile   = flag.String("blacklist", "", "Path to file containing list of blacklisted domains, which will be excluded from the configuration even if present in the domains file (e.g. blacklist.txt)")
+	masqueradesFile = flag.String("masquerades", "", "Path to file containing list of pasquerades to use, with one space-separated 'ip domain' pair per line (e.g. masquerades.txt)")
```
myleshorton: +1

src/github.com/getlantern/flashlight/genconfig/genconfig.go
```diff
@@ -221,37 +222,46 @@ func loadTemplate(name string) string {
...
 		if blacklisted {
 			log.Tracef("Domain %s is blacklisted, skipping", domain)
 			continue
 		}
-		log.Tracef("Grabbing certs for domain: %s", domain)
+		log.Tracef("Grabbing certs for IP %s, domain %s", ip, domain)
 		cwt, err := tlsdialer.DialForTimings(&net.Dialer{
 			Timeout: 10 * time.Second,
-		}, "tcp", domain+":443", false, nil)
+		}, "tcp", ip+":443", false, &tls.Config{ServerName: domain})
```
myleshorton: The addition of ServerName here definitely does not trigger the use of SNI correct? Just want to make double sure!

aranhoide: TBH I'm not positive.  What's your concern here exactly?  We could always use `InsecureSkipVerify` and manually verify the cert afterwards.

myleshorton: Just that SNI is theoretically easier to block, although it probably doesn't matter too much in this case.

myleshorton:
Hey @aranhoide and @fffw I just switched the config on this branch to cloudfront directly and am just sketching out a WIP in config_test.go that I'll move to wherever it should go once it's working.

myleshorton:
Hey @aranhoide I think we do need https://d2wi0vwulmtn99.cloudfront.net/cloud.yaml.gz to be available over http as well after all since otherwise it won't get X-Forwarded-For through normal chained proxies!

myleshorton:
I actually changed it to HTTP and HTTPS in the distribution, but it looks like it needs revalidation, and that's taking forever.

Either that or the config server origin is actually returning 403 for some reason...let me check that.

myleshorton:
OK @aranhoide it's actually working fine with HTTP now, but I think there was an existing issue with the bootstrap server setup and auth tokens again -- I'll fix that next. Now the direct domain fronting config fetching should be working on this branch but not the chained config fetching!

myleshorton:
@fffw, @xiam, or @aranhoide if you could look at the existing changes that would be awesome though. Should be much simpler to use DDF now. The one thing I want to do is randomize the order of the masquerade checks and to eventually also store which ones aren't working, but I don't think we need to do the latter now.

src/github.com/getlantern/fronted/direct.go
```diff
@@ -0,0 +1,122 @@
...
+// successful response.
+func (d *Direct) Response(url string, ms []*Masquerade) ([]byte, error) {
+	for _, m := range ms {
+		client := d.NewHttpClient(m)
+
+		if resp, err := client.Get(url); err != nil {
+			continue
+		} else {
+			defer resp.Body.Close()
+			if 200 != resp.StatusCode {
```
fffw: Do we want to accept ETag and expect 304, like previous implementation?

myleshorton: Oh good catch! Yeah I'll add handling for that @fffw thanks.

myleshorton:
OK the auth token issue with loading the bootstrap config should be solved, but now there's a weird issue where general proxying starts to fail after visiting a page or two.

Otherwise there's the 127.0.0.1 issue, which hopefully we can solve on the servers side.

myleshorton:
Oh I just realized I think the issues I'm seeing here are the result of #3136 not yet being merged, so a result of this branch not yet having those changes...so I wouldn't worry too much about that at this point.

fffw:
LGTM! Tested on mac, windows 7 and ubuntu 14. Merging

PR #3136, opened 2015-09-12
Lots of fixes for authentication tokens in HTTP requests closes #3131

### LINKED ISSUE #3131: Set authentication token for non-CONNECT HTTP requests

myleshorton:


src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -30,8 +30,8 @@ func (client *Client) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 		client.intercept(resp, req)
 	} else {
 		// Direct proxying can only be used for plain HTTP connections.
-		log.Tracef("Reverse proxying %s %v", req.Method, req.URL)
-		client.getReverseProxy().ServeHTTP(resp, req)
+		log.Debugf("Reverse proxying %s %v", req.Method, req.URL)
+		client.newReverseProxy().ServeHTTP(resp, req)
```
myleshorton: Now we create a new reverse proxy every time, which is unfortunate for connection caching, but we can tackle that later.

src/github.com/getlantern/flashlight/client/reverseproxy.go
```diff
@@ -40,23 +28,25 @@ func (client *Client) initReverseProxy(bal *balancer.Balancer, dumpHeaders bool)
 		DisableKeepAlives: true,
 	}
 
+	// Just choose a random dialer that also takes care of things like the
+	// authentication token.
+	dialer := client.getBalancer().RandomDialer()
```
myleshorton: This is the key change here. Just choose a random dialer from the balancer and use that instead of using the balancers's dial method. This allows us to know what dialer we're using ahead of time in order to set the auth token correctly.

myleshorton:
Hey @fffw I know you're out of commission for the weekend, but if you're around @xiam, @aranhoide, @atavism, or @uaalto could you take a quick look at this? It was a bizarrely hard change to make, and the solution isn't ideal because it bypasses some of balancer, but it does the trick.

src/github.com/getlantern/flashlight/client/reverseproxy.go
xiam: Hey @myleshorton, 

I noticed that `RandomDialer()` can return a `nil` function from `randomDialer` and that would trigger a random panic. That could not happen with `getReverseProxy()` because it waited for a valid dialer to arrive before returning.

xiam: There are other things that `balancer.Dial` did, like sending plaintext traffic only to trusted servers and retrying with another dialer in case the first random pick did not work, with `RandomDialer()` we end up with a dialer that may or may not work at all, I imagine this could case trouble when doing lots of requests, like querying small files to display a web page; as I understand each web page resource will require a new random connection.

myleshorton: Thanks @xiam! I just tweaked things to reuse `balancer.Dial` a bit more. There's a chicken and egg issue where we need to know which connection we're using before knowing which auth token to put in the request while the reverseproxy requires them in the opposite order. This gets around that by storing the connection and then just returning that existing connection in the dial method.

What do you think?

fffw:
Another tricky thing remaining. `httputil.ReverseProxy` will add itself to x-forwarded-for, so request header config-server sees eventually will be `x-forwarded-for 127.0.0.1, real client ip, chained server`. I tried to remove it in `balancer.Dialer.Director`, but seems this header is added afterwards (yes, at http://golang.org/src/net/http/httputil/reverseproxy.go#L191).

So unless we stop using `httputil.ReverseProxy`, what we can do is to on the config-server side, check the second IP in `x-forwarded-for`, or just look up the `client-ip` header. How do you think @aranhoide ?

aranhoide:
I can filter out 127.0.0.1 and otherwise keep the current logic.  Or we can
forget about `x-forwarded-for` entirely and pass the IP in a
lantern-specific header (is that `client-ip`)?

fffw:
`127.0.0.1` might not be a safe assumption as it can be changed by setting `addr` (can be useful if user want to proxy traffic from other devices through Lantern). `client-ip` is another header added by ATS to indicate its direct client IP.  It's also [somewhat de-facto](http://stackoverflow.com/questions/7445592/what-is-the-difference-between-http-client-ip-and-http-x-forwarded-for) though other implementations might using `x-client-ip` instead.

myleshorton:
Great catch on the IP issue @fffw. We could also just ignore any non-public address, as I don't think anyone would ever run Lantern on a public address.

myleshorton:
Oh, and I've got a WIP of the DDF stuff with CloudFront working -- should be ready soon.

fffw:
`fetchCloudConfig` in `client/config/config.go#516` doesn't provide auto token, so the initial config update via baked in chained servers will fail.

I created a really ugly PR https://github.com/getlantern/lantern/pull/3138 but it would be better to have cleaner way.

Another possible way is to bake complete lantern.yaml instead of chained.yaml into installer, hence all requests go through balancer and no need to bootstrap manually. Through this way, Lantern can be more robust by fetching initial config through any dialer baked in, regardless it's chained or fronted. Trying.

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -57,16 +57,16 @@ func New(dialers ...*Dialer) *Balancer {
...
-// remaining Dialers until it either manages to connect, or runs out of dialers
-// in which case it returns an error.
-func (b *Balancer) DialQOS(network, addr string, targetQOS int) (net.Conn, error) {
+// TrustedDialerAndConn creaetes a balancer.Dialer and a network connection for an HTTP connection
+// (as opposed to HTTPS).
+func (b *Balancer) TrustedDialerAndConn() (*Dialer, net.Conn, error) {
+	// At this point we don't actually have the destination URL. For our purposes, however,
+	// the destination URL is not actually necessary and is not even used in the dial function
+	// because dial is really just dialing the proxy server.
+	return b.dialerAndConn("tcp", "http://www.doesnotexist.com:80", 0)
```
xiam: Was this URL introduced to force using a trusted dialer? The `addr` will be printed in the logs a few times in case of errors, which could be confusing.

myleshorton: Yeah it's an http link in order to make sure we get a trusted dialer @xiam.

myleshorton:
OK @xiam those last changes get around the issue @fffw mentioned about reverseproxy adding X-Forwarded-For

xiam:
Thanks @myleshorton. Code looks OK to me now. I'd like to test it, do I need any special setup for that?

xiam:
Looks good to me after testing it for a while.

myleshorton:
Awesome thanks!

PR #3139, opened 2015-09-14
bake in lantern.yaml instead of chained.yaml

fffw:
This is another fix for https://github.com/getlantern/lantern/pull/3136#issuecomment-140018568, which supersedes https://github.com/getlantern/lantern/pull/3138 as it makes code simpler.

@myleshorton wanna take a look?


Makefile
```diff
@@ -49,8 +49,8 @@ LOGGLY_TOKEN_MOBILE := d730c074-1f0a-415d-8d71-1ebf1d8bd736
 
 FIRETWEET_MAIN_DIR ?= ../firetweet/firetweet/src/main/
 
-CHAINED_YAML := chained.yaml
-CHAINED_PATH := installer-resources/chained.yaml
+LANTERN_YAML := lantern.yaml
```
fffw: Embed lantern.yaml in installer, and copy it to config directory if there's no existing good config.

fffw:
Tested on mac and windows, it can fetch new config through the baked-in chained server in 1-2 seconds.

myleshorton:
Yeah overall I really like it! The main thing I think it's ultimately missing is the change in #3128 to also use fronted servers as necessary. See https://github.com/getlantern/lantern/pull/3128/files#diff-51c0f46cacd391b79c1c92419364f220R204

At the same time we *don't* want to use fronted servers ever for general traffic the way we used to, so we don't want to integrate them into balancer. Any clean way to do that I wonder?

myleshorton:
I think I really like this change basically as long as two things also happen:

1. lantern.yaml includes CFR fronted dialers (for the initial config download)
1. we automatically retry a few times if we can't get the custom config at first
1. we *don't include CFR fronted dialers* in the downloaded config so that we don't use them for anything but fetching the initial config.

Making that work would also require re-enabling fronted dialers in flashlight/balancer.go as well as disabling the masquerade checks in their current form or fixing them to do direct domain fronting in their checks. Oh that reminds me, though, that would also somehow require doing direct domain fronting when we pass the config download through our normal code path -- right now that wouldn't happen. So we need to think more about that.

What do you think @aranhoide? I'm wondering if we should maybe skip this for now given those complications actually.

myleshorton:
Reopening this for #3177

PR #3146, opened 2015-09-15
iterate through bootstrap chained servers

fffw:
Do you think it's ok to merge to `release-2.0.1`, or `valencia` instead @myleshorton? This fixes the issue of only picking first chained server if more than one available during bootstrap.


uaalto:
LGTM. It is a typical issue with closures, so I'm merging right away.

fffw:
Thanks @uaalto !

myleshorton:
Ha! Darn I had fixed that in an earlier pass too -- good catch @fffw!

PR #3163, opened 2015-09-17
Optional verbose output for checkfallbacks (useful for testing ATS)

uaalto:
I did just to be able to use checkfallbacks as a tool for testing ATS. Could you take a look when you have time @fffw or @aranhoide ? (This is definitely not urgent/important).
It's just optionally showing the req/resp in the output, should be harmless.

Also, I've tried several tools (wget, curl, httpie). They all fail because the either don't support TLS connections to the proxy, nor setting headers for the proxy request.

### DIFF
--- a/src/github.com/getlantern/checkfallbacks/checkfallbacks.go
+++ b/src/github.com/getlantern/checkfallbacks/checkfallbacks.go
@@ -12,6 +12,7 @@ import (
 	"github.com/getlantern/golog"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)


fffw:
Yeah, it's great to have this tool and improve it. I didn't even know the existence of it.

fffw:
LGTM! Inspired by you @uaalto, I'll create another PR to dump more details!

uaalto:
That's awesome! Thank you!

PR #3169, opened 2015-09-18
Fix issue in checkfallbacks

uaalto:
If there is an error in either req or resp, they are nil.


fffw:
LGTM

aranhoide:
@uaalto this is only a problem if we run with the `verbose` flag, right?  If not, we would need to deploy this to the cloudmaster.

uaalto:
No need to push this, it only affects code run by the verbose flag.

Issue #3147, opened 2015-09-15
latest lantern cannot visit website: http://www.speedtest.net/, it show "ERR_TOO_MANY_REDIRECTS"


yuan78:
latest lantern cannot visit website: http://www.speedtest.net/, it show  errors : "ERR_TOO_MANY_REDIRECTS"

What's the reason?
BTW: I built the binary based your latest code ,the last git log is 30e03f0b6a911cdcadb075441e3b59142101c89f

fffw:
Yeah good finding @yuan78 . Visiting http://www.bbc.co.uk/ has the same result.

Seems ATS doesn't handle 301 or 302 correctly.

yuan78:
@fffw  , yes, same result to http://www.bbc.co.uk/
But 2B11, 2.0.2 found no such issue, FYI
Meanwhile the lates lantern will modify the yaml file, it cause we only connect NL server while I originally can connec JP server. I need edit the yaml to specify the JP fallback chainserver IP, then I can connect JP server, hope you can fix this issue also.

fffw:
It happens when the site redirects https to http. Use www.speedtest.net as an example. When requesting https, it will return `302 Found` with `Location:http://www.speedtest.net`, so browser can fetch the http version instead.

Unfortunately, ATS seems always fetching the https version, so the site returns `302 Found` again and again, causes `ERR_TOO_MANY_REDIRECTS` eventually.

fffw:
@yuan78 Weird, your build should get JP server! Could you try turn on the "Proxy All" option, and download `http://d2wi0vwulmtn99.cloudfront.net/cloud.yaml.gz` directly to see if it gives you JP server correctly?

yuan78:
@fffw  Thanks a lot! It works now, I can connect JP server.
Pls make your efforts to fix ATS issue (ATS doesn't handle 301 or 302 correctly. It show "ERR_TOO_MANY_REDIRECTS" when visit http website)

fffw:
Oh gotcha! The request ATS got is https rather than http. We should tamper with the header somewhere.

Seen from ATS debug log.
```
+++++++++ Incoming Request +++++++++
-- State Machine Id: 29
GET https:///cloud.yaml.gz HTTP/1.1
Host: test-config-server.herokuapp.com
User-Agent: curl/7.43.0
Accept: */*
Proxy-Connection: Keep-Alive
X-Lantern-Auth-Token: 123
Accept-Encoding: gzip
```

[...]

Issue #3133, opened 2015-09-12
Close connection if authentication header doesn't validate


myleshorton:
If the auth header doesn't validate, ATS should simply close the connection without any response.

myleshorton:
I believe this was the prior behavior, @fffw.

fffw:
Is it for block resistance @myleshorton ? It's a bit unnatural for Lantern client to handle silently closed connection.

myleshorton:
Yeah exactly @fffw. It's to prevent fingerprinting.

fffw:
Due to the mechanism of ATS, if plugin doesn't respond an HTTP error, ATS will respond HTTP 500 eventually; or if the plugin stops processing immediately, ATS will RST the connection rather than close it gracefully. Do you think RST acceptable to prevent fingerprinting @myleshorton?

fffw:
From Wireshark capture, ATS did send FIN and client replied FIN, but ATS send RST afterwards. This behaviour is strange. Will compare with lantern-java.

fffw:
Lantern java doesn't close silently but mimic Aache. https://github.com/getlantern/lantern-java/blob/master/src/main/java/org/lantern/proxy/GiveModeHttpFilters.java#L167

Copy that to ATS.

fffw:
So below is the response of ATS 5.3.1 with basic configuration (reverse proxy with ssl termination). As ATS itself is widely deployed, it's probably ok to mimic this behaviour. @myleshorton How do you think?

```
 ~ () $ curl -vLk https://188.166.241.143/
*   Trying 188.166.241.143...
* Connected to 188.166.241.143 (188.166.241.143) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
* Server certificate: Moira
> GET / HTTP/1.1
> Host: 188.166.241.143
> User-Agent: curl/7.43.0
> Accept: */*
>
< HTTP/1.1 404 Not Found on Accelerator
< Date: Sun, 20 Sep 2015 06:45:38 GMT
< Connection: keep-alive
< Server: ATS/5.3.1
< Cache-Control: no-store
< Content-Type: text/html
< Content-Language: en
< Content-Length: 297
<
<HTML>
<HEAD>
<TITLE>Not Found on Accelerator</TITLE>
</HEAD>

<BODY BGCOLOR="white" FGCOLOR="black">
<H1>Not Found on Accelerator</H1>
<HR>

[...]

PR #3181, opened 2015-09-21
add -verify option to checkfallbacks fix #3179

### LINKED ISSUE #3179: Regression test of ATS configuration
To avoid those edge cases like ERR_TOO_MANY_REDIRECTS from recurring. Should be able to test through single command, or automated. Maybe add it as a functionality of `chkfallbacks`.

fffw:
Fixes https://github.com/getlantern/lantern/issues/3179

This new option can be used to verify the functionality of ATS based chained server. Currently it only checks https://github.com/getlantern/lantern/issues/3133 and https://github.com/getlantern/lantern/issues/3147, but we can add more checks later.

### DIFF
--- a/src/github.com/getlantern/checkfallbacks/checkfallbacks.go
+++ b/src/github.com/getlantern/checkfallbacks/checkfallbacks.go
@@ -24,6 +24,7 @@ var (
 	verbose       = flag.Bool("verbose", false, "Be verbose (useful for manual testing)")
 	fallbacksFile = flag.String("fallbacks", "fallbacks.json", "File containing json array of fallback information")
 	numConns      = flag.Int("connections", 1, "Number of simultaneous connections")
+	verify        = flag.Bool("verify", false, "Verify the functionality of the fallback")
 
 	expectedBody = "Google is built by a large team of engineers, designers, researchers, robots, and others in many different sites across the globe. It is updated continuously, and built with more tools and technologies than we can shake a stick at. If you'd like to help us out, see google.com/careers.\n"
 )
@@ -149,6 +150,7 @@ func testFallbackServer(fb *client.ChainedServerInfo, workerId int) (output full
 		output.info = []string{"\n" + string(reqStr)}
 	}
 
+	req.Header.Set("X-LANTERN-AUTH-TOKEN", fb.AuthToken)
 	resp, err := c.Do(req)
 	if err != nil {
 		output.err = fmt.Errorf("%v: requesting humans.txt failed: %v", fb.Addr, err)
@@ -180,5 +182,9 @@ func testFallbackServer(fb *client.ChainedServerInfo, workerId int) (output full
 	}
 
 	log.Debugf("Worker %d: Fallback %v OK.\n", workerId, fb.Addr)
+
+	if *verify {
+		verifyFallback(fb, c)
+	}
 	return
 }
--- a/src/github.com/getlantern/checkfallbacks/checkfallbacks.go
+++ b/src/github.com/getlantern/checkfallbacks/checkfallbacks.go
@@ -131,7 +131,7 @@ func testAllFallbacks(fallbacks []client.ChainedServerInfo) (output *chan fullOu
 }
 
 // Perform the test of an individual server
-func testFallbackServer(fb *client.ChainedServerInfo, workerId int) (output fullOutput) {
+func testFallbackServer(fb *client.ChainedServerInfo, workerID int) (output fullOutput) {
 	// Test connectivity
 	fb.Pipelined = true
 	dialer, err := fb.Dialer()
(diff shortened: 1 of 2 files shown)


fffw:
@uaalto My PR comes out finally, mind take a look?

src/github.com/getlantern/checkfallbacks/checkfallbacks.go
```diff
@@ -149,6 +150,7 @@ func testFallbackServer(fb *client.ChainedServerInfo, workerId int) (output full
 		output.info = []string{"\n" + string(reqStr)}
 	}
 
+	req.Header.Set("X-LANTERN-AUTH-TOKEN", fb.AuthToken)
```
uaalto: Isn't this header added already in balancer? Otherwise checkfallbacks wouldn't be working ATM.

src/github.com/getlantern/checkfallbacks/verify.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	req.Header.Set("X-LANTERN-AUTH-TOKEN", "invalid")
+	doVerifyMimic(fb.Addr, req, c)
+}
+
+func doVerifyMimic(addr string, req *http.Request, c *http.Client) {
+	resp, err := c.Do(req)
+	if err != nil {
+		log.Errorf("%v: requesting humans.txt failed: %v", addr, err)
+		if *verbose {
```
uaalto: So, with this we would also be able to print the request to the proxy, right? (Not just the request passed to it to relay, once it is connected). So, effectively we are bypassing the flashlight code currently being used to test fallbacks?

src/github.com/getlantern/checkfallbacks/verify.go
```diff
@@ -0,0 +1,137 @@
...
+		log.Debugf("Body expected: %v", string(atsDefaultBody))
+		log.Debugf("Body got: %v", string(b))
+		return
+	}
+	log.Debugf("%v: OK.", addr)
+}
+
+func verifyRedirectSites(fb *client.ChainedServerInfo, c *http.Client, url string) {
+	req, err := http.NewRequest("GET", url, nil)
+	req.Header.Set("X-LANTERN-AUTH-TOKEN", fb.AuthToken)
```
uaalto: I think that if the previous line fails, this would segfault.

src/github.com/getlantern/checkfallbacks/verify.go
```diff
@@ -0,0 +1,137 @@
...
+			log.Debug(string(reqStr))
+		}
+		return
+	}
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			log.Debugf("Unable to close response body: %v", err)
+		}
+	}()
+	if resp.StatusCode != 200 {
```
uaalto: Is this the only code we might expect? Is it possible that a server returns 20x or 30x and still be acceptable?

src/github.com/getlantern/checkfallbacks/checkfallbacks.go
fffw: Interesting. During my test, `checkfallbacks` never succeeded without this line.

src/github.com/getlantern/checkfallbacks/verify.go
fffw: Oh I didn't notice that I'm printing the original request. In fact it's somewhat meaningless to print request here if http connection failed.

src/github.com/getlantern/checkfallbacks/verify.go
fffw: Good catch!

src/github.com/getlantern/checkfallbacks/checkfallbacks.go
myleshorton: balancer never touches anything with HTTP -- it just deals with net.Dial and net.Conn @fffw and @uaalto. That was actually part of the difficulty with adding the auth header more generally outside of using CONNECT for everything in chained/dialer.go the way we used to.

src/github.com/getlantern/checkfallbacks/verify.go
fffw: For those listed url to test, 200 is the only expected code. `http.Client` will fetch the redirected page automatically.

src/github.com/getlantern/checkfallbacks/checkfallbacks.go
fffw: Yes that's really annoying. I guess the currently running `checkfallbacks` on cloudmaster is a previous version.

fffw:
@uaalto Made a few updates as per your advice.

src/github.com/getlantern/checkfallbacks/checkfallbacks.go
uaalto: I saw it [here](https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/client/chained.go#L98), but that's right, is application code. It's also [here](https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/config/config.go#L578) and [here](https://github.com/getlantern/lantern/blob/valencia/src/github.com/getlantern/flashlight/client/reverseproxy.go#L33), so from the library design point of view, it is expected to be defined always in the user code.

uaalto:
Ok, @fffw I finished with the review. Just that nil check I mentioned in the last comment and I'm ready to merge!

fffw:
Fixed that nil checking, thanks @uaalto

uaalto:
Thank you @fffw! This is going to be very useful for testing, indeed.

PR #3191, opened 2015-09-23
Use direct domain fronting if chained servers fail

myleshorton:
Only for services requesting it. This is close but probably still needs a little work if you want to take a look @fffw.


myleshorton:
Some of this is pulling in your earlier simplifications @fffw for using lantern.yaml instead of chained.yaml

fffw:
Oh it's a lot of work @myleshorton ! Reviewing.

src/github.com/getlantern/flashlight/config/config.go.orig
```diff
@@ -0,0 +1,674 @@
+package config
```
fffw: This file also seems superfluous

myleshorton:
Just added one fix in handler.go and a couple of test "fixes" -- basically deleting old tests.

src/github.com/getlantern/flashlight/client/fronted.go
```diff
@@ -76,7 +75,7 @@ func (s *FrontedServerInfo) dialer(masqueradeSets map[string][]*fronted.Masquera
 		OnDialStats:        s.onDialStats,
 		Masquerades:        masqueradeSets[s.MasqueradeSet],
 		MaxMasquerades:     s.MaxMasquerades,
-		RootCAs:            globals.TrustedCAs,
+		RootCAs:            getCertPool(),
```
fffw: Guess we can eliminate `RootCAs` field here, and let `fronted/dialer.go#L336` to use the one passed to `fronted.Configure`?

fffw:
I created https://github.com/getlantern/lantern/pull/3193 with just a few suggestions.

1. Let `fronted/dialer.go` use the `CertPool` passed to `fronted.Configure()`, so we don't need to keep it at `client/client.go`.

2. Adjust the order of statements in `serveHTTP` and `serveHTTPWithFronting`, in the hope of more clear logic to readers.

During my test on the China box, Lantern could hardly update config or do geolookup through cloudfront, basically all masquerades timed out. Will test on other boxes.

myleshorton:
OK I just added a few more cleanups @fffw. What do you think?

fffw:
New changes LGTM. I found that sometimes the chained server Lantern using is not the one in `lantern.yaml`. I'm checking to make sure it's not related to this PR.

fffw:
Seems new config fetched is saved to disk but didn't have a chance to reconfigure all components, so it will not take effect until Lantern restarts.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -362,13 +299,26 @@ func InConfigDir(filename string) (string, string, error) {
...
 	for _, ca := range cfg.TrustedCAs {
 		certs = append(certs, ca.Cert)
 	}
-	return certs
+	pool, err := keyman.PoolContainingCerts(certs...)
+	if err != nil {
+		log.Errorf("Could not create pool %v", err)
+		return err
+	}
+	poolCh <- pool
```
fffw: Deadlock between here and L302, that's why new config never get propagated.

fffw:
All good except what in https://github.com/getlantern/lantern/pull/3196 !

myleshorton:
Cool - I'll look at that right away. I also want to add in fronting support
for auto-updates so will also add that in - give me 30 mins!

myleshorton:
> Seems new config fetched is saved to disk but didn't have a chance to reconfigure all components, so it will not take effect until Lantern restarts.

Your deadlock fix also fixed this issue right @fffw?

fffw:
@myleshorton Yes, this issue is caused by the dead lock.

myleshorton:
OK guys I'm merging this in! Thanks for everyone's help on it.

PR #3196, opened 2015-09-24
remove deadlock

fffw:
This is to remove deadlock during config update

### DIFF
--- a/src/github.com/getlantern/flashlight/config/config.go
+++ b/src/github.com/getlantern/flashlight/config/config.go
@@ -51,7 +51,6 @@ var (
 	lastCloudConfigETag = map[string]string{}
 	httpClient          atomic.Value
 	r                   = regexp.MustCompile("\\d+\\.\\d+")
-	poolCh              = make(chan *x509.CertPool, 1)
 )
 
 type Config struct {
@@ -264,10 +263,6 @@ func Run(updateHandler func(updated *Config)) error {
 
 func updateGlobals(cfg *Config) error {
 	globals.InstanceId = cfg.InstanceId
-	err := cfg.createCertPool()
-	if err != nil {
-		return fmt.Errorf("Unable to configure trusted CAs: %s", err)
-	}
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -299,26 +294,16 @@ func InConfigDir(filename string) (string, string, error) {
...
-func (cfg *Config) GetTrustedCACerts() *x509.CertPool {
-	pool := <-poolCh
-	if len(poolCh) == 0 {
-		poolCh <- pool
-	}
-	return pool
-}
-
-func (cfg *Config) createCertPool() error {
+func (cfg *Config) GetTrustedCACerts() (pool *x509.CertPool, err error) {
```
fffw: `GetTrustedCACerts()` is called only during initialization or when new config available, no need to cache the certs. So move `createCertPool()` here and remove the channel which causes block.

src/github.com/getlantern/fronted/direct.go
```diff
@@ -22,8 +22,10 @@ var (
 )
 
 func Configure(pool *x509.CertPool, masquerades []*Masquerade) {
-	poolCh <- pool
-	masqueradesCh <- masquerades
+	go func() {
```
fffw: This will block until the channel is consumed, so wrap it in goroutine.

src/github.com/getlantern/fronted/direct.go
```diff
@@ -34,6 +36,17 @@ func getCertPool() *x509.CertPool {
 	return pool
 }
 
+type Direct struct {
+	tlsConfigs      map[string]*tls.Config
+	tlsConfigsMutex sync.Mutex
+}
+
+func NewDirect() *Direct {
```
fffw: Hide the details in constructor.

fffw:
Finally got the bug found and fixed. @myleshorton This is the only remaining thing I found regarding your PR.

src/github.com/getlantern/fronted/direct.go
myleshorton: Yeah good catch

myleshorton:
LGTM!

PR #3202, opened 2015-09-27
Client-side parallel fronted and chained server access

myleshorton:
Hey @fffw this is again a pretty big change to the way we handle the whole parallel fronted and chained server access thing. It basically moves things out of handler.go and into the new chainedAndFronted struct in util/http.go. The fundamental reason is that we need to do both in parallel, and that was difficult to achieve on the "server" side (handler.go).

This also cleans up issues with config.go and loading of masquerades, adding masquerades to lantern.yaml. In theory we could now not generate masquerades.go and could instead basically ship with a modified cloud.yaml, but I haven't had a chance to tackle that yet.

Let me know what you think!


src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -11,18 +11,10 @@ import (
```
myleshorton: All the changes to handler.go basically just revert the previous change to using direct fronting here.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -171,13 +170,17 @@ func useGoodOldConfig(configDir, configPath string) bool {
...
 	file := "lantern-" + version + ".yaml"
-	configDir, configPath, err := InConfigDir(file)
+	_, configPath, err := InConfigDir(file)
 	if err != nil {
 		log.Errorf("Could not get config path? %v", err)
 		return nil, err
 	}
-	ok := useGoodOldConfig(configDir, configPath)
-	if !ok {
+	run := isGoodConfig(configPath)
```
myleshorton: For now we go back to just using configs for this version, as the old configs will typically have bad CloudFlare masquerades we can't use and that will cause problems.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -469,21 +471,19 @@ func fetchCloudConfig(client *http.Client, url, authToken string) ([]byte, error
...
+	req.Header.Set("Lantern-Fronted-URL", frontedCloudConfigUrl)
+
 	// make sure to close the connection after reading the Body
 	// this prevents the occasional EOFs errors we're seeing with
 	// successive requests
 	req.Close = true
 
-	resp, err := client.Do(req)
+	// Request the config via either chained servers or direct fronted servers.
+	cf := util.NewChainedAndFronted()
```
myleshorton: Use our new chainedAndFronted thing here.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -293,6 +293,13 @@ func runClientProxy(cfg *config.Config) {
 	}
 
 	applyClientConfig(client, cfg)
+
+	// Only run analytics once on startup. It subscribes to IP discovery
+	// events from geolookup, so it needs to be subscribed here before
+	// the geolookup code executes.
+	analytics.Configure(cfg, version)
+	geolookup.Start()
```
myleshorton: Just pulling these out -- an unrelated change but they really shouldn't be updated with every config update...

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -362,7 +376,7 @@ func applyClientConfig(client *client.Client, cfg *config.Config) {
 	if err != nil {
 		log.Errorf("Unable to get trusted ca certs, not configure fronted: %s", err)
 	} else {
-		fronted.Configure(certs, cfg.Client.MasqueradeSets["cloudfront"])
+		fronted.Configure(certs, cfg.Client.MasqueradeSets)
```
myleshorton: We really don't want to hardcode cloudfront here -- use whatever we have.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -381,12 +394,11 @@ func applyClientConfig(client *client.Client, cfg *config.Config) {
 
 	// We offload this onto a go routine because creating the http clients
 	// blocks on waiting for the local server, and the local server starts
-	// on the thread, otherwise creating a deadlock.
+	// later on this same thread, so it would otherwise creating a deadlock.
 	go func() {
-		withHttpClient(cfg.Addr, config.Configure)
```
myleshorton: This is essentially replaced with the StartPolling call. We also don't want to call this on every config change.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -455,15 +464,6 @@ func runServerProxy(cfg *config.Config) {
...
 
-func updateServerSideConfigClient(cfg *config.Config) {
-	client, err := util.HTTPClient(cfg.CloudConfigCA, "")
-	if err != nil {
-		log.Errorf("Couldn't create http.Client for fetching the config")
-		return
-	}
-	config.Configure(client)
-}
-
```
myleshorton: Just ditching this for now...

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -2,21 +2,138 @@ package util
...
+				c.chainedSucceeded.Store(true)
+			}
+			responses <- resp
+		}
+	}
+
+	go func() {
+		// If chained servers have been succeeding generally, just keep using
+		// them instead of using fronting.
+		if c.chainedSucceeded.Load().(bool) {
```
myleshorton: Don't use fronting if chained is working...

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -2,21 +2,138 @@ package util
...
+				return nil, errors.New("All requests errored")
+			}
+		case <-timeout:
+			log.Errorf("Timed out!")
+			return nil, errors.New("Timed out!")
+		}
+	}
+	return nil, errors.New("Reached end")
+}
+
```
myleshorton: This could probably be done better -- just still not that experience with all this =)

src/github.com/getlantern/flashlight/util/out.txt
```diff
@@ -0,0 +1,16 @@
+=== RUN TestChainedAndFronted
```
fffw: Intentionally put this file in repo or not?

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -2,21 +2,138 @@ package util
...
 	"time"
 
+	"github.com/getlantern/fronted"
 	"github.com/getlantern/golog"
 	"github.com/getlantern/keyman"
 	"github.com/getlantern/waitforserver"
 )
 
+const (
+	defaultAddr = "127.0.0.1:8787"
```
fffw: We may not want to hardcode proxy address here, though it's also the behaviour of 2.0.6.

src/github.com/getlantern/flashlight/config/config.go
fffw: We'd better to respect existing config file as it still contains user settings. We can add the check for cloudfront masquerades in `isGoodConfig` so 2.0.7 will  drop config of previous versions, but later versions can keep user settings.

If we do decide to drop previous config, function `useGoodOldConfig` can be removed.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -231,7 +234,7 @@ func pollWithHttpClient(currentCfg yamlconf.Config, client *http.Client) (mutate
 	// We don't pass an auth token here, as the http client is actually hitting the localhost
```
fffw: This line of comment seems odd as auth token is not something `config` package should care.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -469,21 +471,19 @@ func fetchCloudConfig(client *http.Client, url, authToken string) ([]byte, error
 		req.Header.Set(ifNoneMatch, lastCloudConfigETag[url])
 	}
 
-	if authToken != "" {
-		req.Header.Set("X-LANTERN-AUTH-TOKEN", authToken)
-	}
-
-	req.Header.Set("Lantern-Fronted-URL", frontedCloudConfigUrl)
-
 	// Prevents intermediate nodes (CloudFlare) from caching the content
```
fffw: This stale comment can be changed slightly.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -469,21 +471,19 @@ func fetchCloudConfig(client *http.Client, url, authToken string) ([]byte, error
...
 	req.Header.Set("Cache-Control", "no-cache")
 
+	req.Header.Set("Lantern-Fronted-URL", frontedCloudConfigUrl)
```
fffw: A comment  on the purpose of this header would make reader happier.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -381,12 +394,11 @@ func applyClientConfig(client *client.Client, cfg *config.Config) {
 
 	// We offload this onto a go routine because creating the http clients
 	// blocks on waiting for the local server, and the local server starts
-	// on the thread, otherwise creating a deadlock.
+	// later on this same thread, so it would otherwise creating a deadlock.
 	go func() {
-		withHttpClient(cfg.Addr, config.Configure)
-		withHttpClient(cfg.Addr, geolookup.Configure)
 		withHttpClient(cfg.Addr, statserver.Configure)
```
fffw: `statserver` uses the `HttpClient` to geolocate, we can also remove `Configure` here and reply on `pubsub` to keep consistency.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -381,12 +394,11 @@ func applyClientConfig(client *client.Client, cfg *config.Config) {
...
 	}()
+
 }
 
 func withHttpClient(addr string, withClient func(client *http.Client)) {
```
fffw: This can also be removed if `statserver` changed to using `pubsub`.

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -2,21 +2,138 @@ package util
...
+			log.Debugf("Got an error: %v", err)
+			if i == 1 {
+				return nil, errors.New("All requests errored")
+			}
+		case <-timeout:
+			log.Errorf("Timed out!")
+			return nil, errors.New("Timed out!")
+		}
+	}
+	return nil, errors.New("Reached end")
```
fffw: Maybe panic here?

src/github.com/getlantern/flashlight/util/http.go
fffw: One slight improvement would be to store the two goroutine funcs in a slice as a member of `chainedAndFronted`, and once chained goroutine succeeded, remove fronted func (fixed index in slice) from the slice. Through this way, we don't need to maintain the flag and no longer need atomic operation, as access to the slice has no race condition.

Just an extra option. Current implementation is good enough.

src/github.com/getlantern/flashlight/util/util_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"crypto/x509"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/getlantern/fronted"
+	"github.com/getlantern/keyman"
+	"github.com/mailgun/oxy/forward"
```
fffw: `oxy/forward` package is not in repo, though it will definitely be added later by the chained server implementation.

`flashlight/util` is not added in `testpackages.txt`.

src/github.com/getlantern/fronted/direct.go
```diff
@@ -36,31 +53,23 @@ func getCertPool() *x509.CertPool {
...
-func (d *Direct) getMasquerade() *Masquerade {
-	if len(masqCh) > 0 {
-		return <-masqCh
-	}
-	log.Debugf("Refreshing live masquerades")
-	ms := <-masqueradesCh
-	size := len(ms)
-	for i := 0; i < 10; i++ {
+func (d *direct) fillMasquerades() {
+	for i := 0; i < 40; i++ {
```
fffw: We only want to dial first 40 masquerades from shuffled sets in parallel and ignore the rest?

I guess we can just spawn a few goroutines, say, 10, and loop in the goroutine when error encountered to dial one masquerade, so we always have and only have a few valid masquerades in stock. Testing against one masquerade will fail soon unless times out, so it will not slow down the caller of `getMasquerade` I think.

fffw:
Ok I've done the first pass of review @myleshorton . You've made lots of critical improvements!

fffw:
I saw 100+ `CLOSE_WAIT` connections to masquerades after running for 10 hours, which caused `too many open files` error finally.

myleshorton:
> I saw 100+ CLOSE_WAIT connections to masquerades after running for 10 hours, which caused too many open files error finally.

I think I have an idea on why those are happening -- basically the chainedAndFronted Do won't properly close the second response if both succeed, and I'm sure other leaky cases as well. I'll work on that more in the morning.

fffw:
> I saw 100+ CLOSE_WAIT connections to masquerades after running for 10 hours, which caused too many open files error finally.

@myleshorton The `defer` call in `fronted/direct.go#L79` is also responsible. The connection should be closed immediately instead.

src/github.com/getlantern/flashlight/util/http.go
myleshorton: Yeah that's an interesting idea @fffw. I was also considering using states to clean it up a little bit. I might fiddle a little bit.

src/github.com/getlantern/flashlight/util/http.go
myleshorton: It's just hard for me to imagine a scenario where we'd actually ever change that using the cloud config, and supporting that option complicates the code quite a bit because then anything using an HTTP client to the localhost has to reconfigure with new configs.

src/github.com/getlantern/flashlight/config/config.go
myleshorton: removed

src/github.com/getlantern/flashlight/flashlight.go
myleshorton: `statserver` uses the http client to lookup the IP address of other clients, though, not us. So `pubsub` won't work I don't think.

myleshorton:
OK @fffw I made a few changes. I tweaked the masquerade checking in direct.go quite a bit. Basically, I now don't automatically requeue masquerades as we use them because I'm seeing a lot of masquerades where the IP we have no longer matches the domain, so re-verifying them is probably a good idea. I also now re-add masquerades to the candidate channel if they connected or had a reasonable error, and because we don't auto-requeu them we'll make it well past the initial 40.

I'm not totally against doing this directly in getMasquerade however, but I think this is more or less what we want as well.

myleshorton:
We should add the actual IPs here before merging.

fffw:
I guess the whole `src/github.com/getlantern/Resources/` directory is added to repo unintentionally right?

fffw:
We can `gost add github.com/mailgun/oxy/forward` to avoid adding the whole `github.com/mailgun/oxy` code and its dependency (`gopkg.in/check.v1`).

src/github.com/getlantern/flashlight/util/http.go
fffw: I came across this when running different versions of Lantern instance on same VPS to measure load time. I supplied different `-addr=` options to each instance, and lots of `connection refused` error reported during config update and geolookup. I redirected output to `/dev/null` to prevent those errors from blowing up the whole disk.

I agree that regular Lantern client will hardly change the port, so it's probably ok.

src/github.com/getlantern/flashlight/config/config.go
fffw: What do you think about this one @myleshorton ?

myleshorton: Hey @fffw yeah I really think we should just have separate settings honestly. Maybe I'll just go ahead and do that.

myleshorton:
Unfortunately `gost get github.com/mailgun/oxy/forward master` seems to pull those other dependencies in as well @fffw.

uaalto:
Are you using Oxy in Lantern client too? I mention it because I'm adding these dependencies in the `chained-server` repo so they are isolated.

myleshorton:
Just in a test @uaalto

uaalto:
Ok! Just making sure I'm not duplicating stuff unnecessarily.

myleshorton:
Makes sense. Yeah I think for now we want the chained server to be separate regardless.

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -2,21 +2,162 @@ package util
...
+
+	cf.fetcher = dual
+	cf.dual = dual
+	cf.chained = &chainedFetcher{}
+	return cf
+}
+
+// ChainedAndFronted fetches HTTP data in parallel using both chained and fronted
+// servers.
+type chainedAndFronted struct {
```
fffw: I would create `chainedFetcher` and `dualFetcher` on the fly and assign to `fetcher` at that moment, as no extra cost involved. Three members with same type make me a bit confusion at first glance.

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -2,21 +2,162 @@ package util
...
+	fetcher HTTPFetcher
+	chained HTTPFetcher
+	dual    HTTPFetcher
+}
+
+// Do will attempt to execute the specified HTTP request using only a chained fetcher
```
fffw: Revise the comment slightly?

myleshorton: Will do thanks @fffw! I also changed direct.go to just use the first
connection that succeeds to keep things simple (there was a bug in the
previous where it wouldn't keep filling the connection queue).

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -2,21 +2,162 @@ package util
...
+
+	for i := 0; i < 2; i++ {
+		select {
+		case resp := <-responses:
+			if i == 1 {
+				log.Debugf("Got second response -- sending")
+				return resp, nil
+			} else if success(resp) {
+				log.Debugf("Got good response")
+				return resp, nil
```
fffw: This will leave 2nd response not closed, if any. Maybe leave a comment here if we don't have a viable way to solve it for now.

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -2,21 +2,162 @@ package util
...
+				_ = resp.Body.Close()
+			}
+		case err := <-errs:
+			log.Debugf("Got an error: %v", err)
+			if i == 1 {
+				return nil, errors.New("All requests errored")
+			}
+		case <-timeout:
+			log.Errorf("Timed out!")
+			return nil, errors.New("Timed out!")
```
fffw: Save as above. In extreme cases, all 2 response will be left unclosed.

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -2,21 +2,162 @@ package util
...
+	defaultAddr = "127.0.0.1:8787"
+)
+
 var (
 	log = golog.LoggerFor("flashlight.util")
+
+	// This is for doing direct domain fronting if necessary. We store this as
+	// an instance variable because it caches TLS session configs.
+	direct  = fronted.NewDirect()
+	timeout = time.After(60 * time.Second)
```
fffw: `timeout` should be a variable instantiated at L98, or it will fire once and never again. Or, remove it at all and rely on underlie clients?

fffw:
@myleshorton LGTM except those I commented above. Code becomes much cleaner!

src/github.com/getlantern/flashlight/util/http.go
myleshorton: Yup good catch!!

src/github.com/getlantern/flashlight/util/http.go
myleshorton: Yeah good point @fffw. We should probably shoot resp to a channel here instead and keep looping, something like that, but for now I'll just leave a comment.

src/github.com/getlantern/flashlight/util/http.go
myleshorton: Stupidly made `timeout` external for testing...whoops! :+1:

myleshorton:
Awesome thanks for all those @fffw. What do you think now? I started trying to fix lantern-mobile/client_test.go but wasn't quite able to. Might just leave it for now!

fffw:
LGTM!

myleshorton:
Awesome thanks @fffw!

PR #3213, opened 2015-10-01
Changes to return correct response code from the local HTTP proxy and…

### LINKED ISSUE #3210: Lantern doesn't include default list of blocked sites on initial startup

### LINKED ISSUE #3211: Consolidate lantern.yaml and cloud.yaml
This would unify scripts for releasing new versions.

### LINKED ISSUE #3212: Lantern incorrectly thinking chained servers are succeeding
...and using them exclusively in http.go as a result. The problem is that the local Lantern proxy will return 200 OK even on failures.

myleshorton:
… to generate configs the client side can properly use closes #3212, closes #3211, closes #3210


myleshorton:
Hey @fffw I know you're supposed to be on vacation, but if you have a second would you mind taking a look at this one?

src/github.com/getlantern/flashlight/genconfig/genfallbacks.go
```diff
@@ -0,0 +1,256 @@
...
+	"sync"
+	"text/template"
+
+	"github.com/getlantern/flashlight/client"
+	"github.com/getlantern/golog"
+	"github.com/getlantern/yaml"
+)
+
+const (
+	numberOfWorkers = 50
```
fffw: These two constants are no longer used.

src/github.com/getlantern/flashlight/genconfig/genfallbacks.go
```diff
@@ -0,0 +1,256 @@
...
+	ftVersionFile   = `https://raw.githubusercontent.com/firetweet/downloads/master/version.txt`
+)
+
+var (
+	help            = flag.Bool("help", false, "Get usage help")
+	masqueradesFile = flag.String("masquerades", "", "Path to file containing list of pasquerades to use, with one space-separated 'ip domain' pair per line (e.g. masquerades.txt)")
```
fffw: Also these variables.

src/github.com/getlantern/flashlight/genconfig/genfallbacks.go
myleshorton: I think they are still used right @fffw?

src/github.com/getlantern/flashlight/genconfig/genfallbacks.go
myleshorton: Wait I think those are still also used right? The masqueradesFile and help?

src/github.com/getlantern/flashlight/genconfig/genfallbacks.go
fffw: I think they are used by `genconfig.go` but not `genfallbacks.go`? Nothing important anyway, it's just a command line tool.

src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
```diff
@@ -7,7 +7,7 @@ client:
   frontedservers: []
   chainedservers: {{range .fallbacks}}
     fallback-{{.ip}}:
-      addr: {{.ip}}:443
+      addr: {{.ip}}
```
fffw: What's the intention of this change?

myleshorton: Basically the ChainedServerInfo Addr field includes the port, which wasn't the case with the previous IP setting @fffw.

fffw: So `fallback-{{.ip}}:` will expand to `fallback-x.x.x.x:443:`? The extra colon seems invalid.

src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
```diff
@@ -20,6 +20,9 @@ client:
     - domain: {{.Domain}}
       ipaddress: {{.IpAddress}}{{end}}
 proxiedsites:
+  delta:
+    additions: []
+    deletions: []
```
myleshorton: This change just makes the generated yaml file actually work on the client side -- without this the client side won't load.

src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
myleshorton: Oh good point @fffw. It seems to work for some reason, but let me clarify.

myleshorton: Or sorry let me make sure rather =).

myleshorton: Yeah it works for whatever reason @fffw -- maybe it looks for colon followed by some whitespace?

fffw: Maybe wrap above two line with quotes? That will make it more clear.

fffw: Ok http://www.yaml.org/spec/1.2/spec.html#id2759963 says that `Mappings use a colon and space (“: ”) to mark each key: value pair`, so it's legal yaml.

myleshorton: Awesome @fffw thanks!

fffw:
in `util/http.go#L69`, `chainedFetcher.Do` should check the response status code

fffw:
Ok @myleshorton once the above bug fixed, everything should be ok to me.

myleshorton:
OK thanks @fffw I'm on it!

myleshorton:
OK got that in @fffw!

PR #3216, opened 2015-10-02
Moved to use embedded bootstrap config to fix auto-updates closes #3215

### LINKED ISSUE #3215: Embedded bootstrap configuration breaks auto-updates

myleshorton:


src/github.com/getlantern/flashlight/config/bootstrap.go
```diff
@@ -81,26 +81,39 @@ func readSettingsFromFile(yamlPath string) (*BootstrapSettings, error) {
...
+	fs, err := tarfs.New(Resources, dir)
 	if err != nil {
+		log.Errorf("Could not read resources? %v", err)
 		return err
 	}
-	bytes, err := ioutil.ReadFile(src)
+
+	bytes, err := fs.Get("lantern.yaml")
 	if err != nil {
+		log.Errorf("Could not read bootstrap file %v", err)
```
uaalto: Could this be logged twice? Both here and where the returned error is collected.

myleshorton: Certainly possible but looks like it's not in this case (https://github.com/getlantern/lantern/blob/release-2.0.1/src/github.com/getlantern/flashlight/config/config.go#L186)

myleshorton:
The addition of the resources.go file is really the big change @uaalto -- basically that includes lantern.yaml so that the lantern binary will have a lantern.yaml file when it auto-updates.

uaalto:
The code seems OK. The `resources.go` file is a go file with a bunch of `byte[]` definitions, right?. I understand it tries to read from an embedded yaml file instead of from a regular file.

src/github.com/getlantern/flashlight/config/bootstrap.go
```diff
@@ -121,7 +134,7 @@ func bootstrapPath(fileName string) (string, error) {
 	}
 	fullPath := filepath.Join(yamldir, fileName)
 	log.Debugf("Opening bootstrap file from: %v", fullPath)
-	return fullPath, nil
+	return yamldir, fullPath, nil
```
uaalto: maybe we don't need to return the fullPath anymore? it can be derived from the parameter and the return value.

src/github.com/getlantern/flashlight/config/bootstrap_test.go
```diff
@@ -55,7 +55,7 @@ func TestBootstrapSettings(t *testing.T) {
 	assert.Equal(t, local, path, "Wrote to unexpected path")
 	assert.True(t, err == nil, "Should not be an error")
 
-	path, err = bootstrapPath(name)
+	_, path, err = bootstrapPath(name)
```
uaalto: well, you are using it alternatively here, just the fullPath

src/github.com/getlantern/flashlight/config/bootstrap.go
myleshorton: Yeah in theory the caller should really use filepath.Join, though, which seems like it's kind of asking a bit much

src/github.com/getlantern/flashlight/config/bootstrap_test.go
myleshorton: Yeah exactly..

src/github.com/getlantern/flashlight/config/bootstrap.go
```diff
@@ -81,26 +81,39 @@ func readSettingsFromFile(yamlPath string) (*BootstrapSettings, error) {
...
 		return err
 	}
-	bytes, err := ioutil.ReadFile(src)
+
+	bytes, err := fs.Get("lantern.yaml")
 	if err != nil {
+		log.Errorf("Could not read bootstrap file %v", err)
 		return err
 	}
 	err = ioutil.WriteFile(configPath, bytes, 0644)
```
uaalto: This will overwrite any previous file, because we didn't check for existence. Is that what we are looking for?

uaalto: This line explains it I guess: https://github.com/getlantern/lantern/blob/release-2.0.1/src/github.com/getlantern/flashlight/config/config.go#L180

myleshorton: Yeah exactly -- in this case we've already checked that the initial file doesn't exist. Good question though.

myleshorton:
One good test @uaalto is actually just to run it with no existing yaml in Application Support/Lantern and with no yaml in ../Resource/en.lproj (if you had created that manually by chance).

src/github.com/getlantern/flashlight/config/bootstrap.go
uaalto: I guess this is the gist: we first check if we have a good config. Otherwise we extract one from the tarfs system and save it as a regular file. Next time we run Lantern, it will read that one.
What happens with autoupdates then?

myleshorton:
You should basically just see Lantern create a valid yaml file in Application Support/Lantern in that case, and it should run fine.

myleshorton:
That geolookup_test.go failure is a little weird, but I don't think it's related.

src/github.com/getlantern/flashlight/config/bootstrap.go
myleshorton: Yeah that's right @uaalto, although it will actually read the extracted one *this time we run Lantern* as well as next and subsequent times. Then, when we autoupdate, the new updated binary will have its own embedded lantern.yaml, and we'll use that one.

uaalto: OK. It's clear now.

uaalto:
I'm going to do that test.

uaalto:
Works fine.
Haven't seen that _../Resource/en.lproj_ before. Where do you mean exactly?

myleshorton:
> Works fine.

Awesome!

> Haven't seen that ../Resource/en.lproj before. Where do you mean exactly?

Basically on OSX that's where Lantern looks for any "embedded" files that are packaged with the installer. Since in production we run from /Applications/Lantern.app/Contents/MacOS/lantern, the ../Resources/en.lproj folder is where we need to look for those files (that's one of the only directories that won't break code signatures). When running from the command line, Lantern will look for that directory also. Actually prior to this change I think you even needed that directory to exist (or to have a previous yaml file in Application Support/Lantern), but so testing without that directory at all should successfully verify that the embedded resource approach is working.

myleshorton:
Actually that extracted yaml config doesn't look quite right -- I'm investigating now!

myleshorton:
Oh actually I think it's just the one that's fetched remotely...probably fine..

uaalto:
What did you see wrong? It looks fine in my case, and works

uaalto:
The tests are stuck at the IP geolookup, it's probably going to timeout. That doesn't seem related, though. I can merge, or do you want to see what happens with the geolookup before?

myleshorton:
well actually it's a little odd that it includes that `frontedservers` field at all because I updated the cloud config just this morning (or late last night can't remember now!). I'm thinking maybe that part's not actually just taken from cloud.yaml. I.e. if you look at the cloud.yaml itself it doesn't have any frontedservers:

https://github.com/getlantern/lantern/blob/release-2.0.1/src/github.com/getlantern/flashlight/genconfig/cloud.yaml

myleshorton:
Oh does the geolookup test hang for you too @uaalto? It's working for me, so I was trying to figure that out.

uaalto:
It hangs in Travis.

uaalto:
Travis: `Test killed with quit: ran too long (10m0s).`

myleshorton:
Right yeah I know about the travis hang. Does it work for you locally?

uaalto:
Yes, it does. I passes and returns immediately. Both getlantern/geolookup and getlantern/flashlight/geolookup.

myleshorton:
> Yes, it does. I passes and returns immediately. Both getlantern/geolookup and getlantern/flashlight/geolookup.

OK cool. Yeah I think it's fine, although it's it looks like it's just not getting any masquerade candidates in Travis's case, which is very strange. I might just quickly test on Linux to make sure there's not some deadlock we're missing. Actually I'll run with race detection too!

myleshorton:
Yup there's a race condition in there -- fiddling with it now!

myleshorton:
OK just fixed it @uaalto -- man I've got to remember to run more often with -race! What an awesome tool.

uaalto:
Great!! Yes, it's awesome. Countless hours saved. So, let's see Travis results!
Puzzles me that the bug wasn't triggered on OSX though..

myleshorton:
> Puzzles me that the bug wasn't triggered on OSX though..

Just a thread scheduling thing I suspect -- definitely won't happen every time.

myleshorton:
I'm sure something else will fail in Travis (lantern-mobile test in particular is still broken on this branch).

myleshorton:
Hey @uaalto just fixed a couple of race conditions from running with -race! I think this should be good to go now -- could fix the lantern-mobile test, but it's a little annoying and was breaking with 2.0.7 anyway

uaalto:
The last commit LGTM! Ready to merge?

myleshorton:
Awesome! Yeah merge away @uaalto!

PR #3239, opened 2015-10-06
retry requests to get passed occasional 403 errors closes #3238

### LINKED ISSUE #3238: Some domains return 403 to fronted requests

myleshorton:


src/github.com/getlantern/fronted/direct.go
```diff
@@ -117,6 +117,24 @@ func (d *direct) NewDirectHttpClient() *http.Client {
 	}
 }
 
+// Do continually retries a given request until it succeeds because some fronting providers
+// will return a 403 for some domains.
+func (d *direct) Do(req *http.Request) (*http.Response, error) {
+	for i := 0; i < 6; i++ {
+		client := d.NewDirectHttpClient()
```
atavism: Do we have to re-create the http.Client during each iteration or could we put this outside of the loop?

src/github.com/getlantern/fronted/direct.go
```diff
@@ -117,6 +117,24 @@ func (d *direct) NewDirectHttpClient() *http.Client {
...
+		if resp, err := client.Do(req); err != nil {
+			log.Errorf("Could not complete request %v", err)
+			continue
+		} else if resp.StatusCode > 199 && resp.StatusCode < 400 {
+			return resp, err
+		} else {
+			_ = resp.Body.Close()
+			continue
+		}
```
atavism: We should be able to get rid of these ``continue`` statements and we can just change ``_ = resp.Body.Close()`` to ``res.Body.Close()`` since we don't care about the return value (or is that to just be explicit that we're ignoring it?)

myleshorton: Yeah we totally could get rid of the continues. The _ = is really just there so it passes errcheck @atavism

src/github.com/getlantern/fronted/direct.go
myleshorton: We probably could create a single one outside of the loop, but it would just introduce shared state between the requests with maybe saving a few bytes of memory as far as I can tell. What's the advantage of it being outside of the loop?

src/github.com/getlantern/fronted/direct.go
atavism: Cool, do we want to just remove those continue statements then? Everything else looks good to me so we can go ahead and merge.

myleshorton: Yeah that sounds good man -- I'll make that change and will then put it
into QA. I am tempted to add a test for that new Do method - might do that
in the morning. Thanks for reviewing!

myleshorton:
merging -- test errors are with CloudFront distributions again.

Issue #3246, opened 2015-10-09
`systray` prevents Lantern from compiling on darwin


fffw:
Same as https://github.com/getlantern/systray/issues/6

> warning: 'menuZone' is deprecated: first deprecated in OS X 10.11 [-Wdeprecated-declarations]

PR #3247, opened 2015-10-09
remove deprecated menuZone from systray

fffw:
It's a rather quick fix to https://github.com/getlantern/lantern/issues/3246

As per [doc](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Miscellaneous/Foundation_Functions/index.html#//apple_ref/c/func/NSDefaultMallocZone)

> Zones are ignored on iOS and 64-bit runtime on OS X. You should not use zones in current development.

So it's safe to run on any supported OS X versions.

@myleshorton mind take a look?

### DIFF
--- a/src/github.com/getlantern/systray/systray_darwin.m
+++ b/src/github.com/getlantern/systray/systray_darwin.m
@@ -52,8 +52,7 @@
 - (void)applicationDidFinishLaunching:(NSNotification *)aNotification
 {
   self->statusItem = [[NSStatusBar systemStatusBar] statusItemWithLength:NSVariableStatusItemLength];
-  NSZone *menuZone = [NSMenu menuZone];
-  self->menu = [[NSMenu allocWithZone:menuZone] init];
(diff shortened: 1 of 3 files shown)


src/github.com/getlantern/systray/systray_darwin.m
```diff
@@ -52,8 +52,7 @@ - (IBAction)menuHandler:(id)sender;
 - (void)applicationDidFinishLaunching:(NSNotification *)aNotification
 {
   self->statusItem = [[NSStatusBar systemStatusBar] statusItemWithLength:NSVariableStatusItemLength];
-  NSZone *menuZone = [NSMenu menuZone];
-  self->menu = [[NSMenu allocWithZone:menuZone] init];
+  self->menu = [[NSMenu alloc] init];
```
fffw: Actual change comes here. Others are just to make `golint` happier.

myleshorton:
LGTM @fffw!

PR #3249, opened 2015-10-09
Release 2.0.1

myleshorton:
Merging the latest from 2.0.10


myleshorton:
Merging -- test failures are again with CloudFront distributions. @aranhoide do we even need that peerscanner test at all at this point?

PR #3269, opened 2015-10-21
Separate user/instance-specific settings from global config closes #2896

### LINKED ISSUE #2896: User settings not saved across versions
As a result of changing the config to embed the version name in the yaml config file name breaks saving user settings across versions. This is related to #2390, but ideally we would move user settings into a separate file, say settings.yaml.

myleshorton:


myleshorton:
Hey @fffw you mind taking a look at this one?

src/github.com/getlantern/flashlight/settings/settings.go
```diff
@@ -32,51 +38,108 @@ type Settings struct {
...
+	sync.RWMutex
 }
 
-func Configure(cfg *config.Config, version, revisionDate string, buildDate string) {
+// Load loads the initial settings at startup, either from disk or using defaults.
+func Load(version, revisionDate, buildDate string) {
+	// Create default settings that may or may not be overridden from an existing file
+	// on disk.
+	settings = &Settings{
+		Version:      version,
```
fffw: `Version` and below two structure members are not used anywhere else, what's the purpose of writing them to setting file?

myleshorton: Good question. I guess they're only used in the UI correct? I don't see much harm in writing them to the file, but I could also probably add an easy comment/tag that would remove them. Is that what you're suggesting?

fffw: Oh I see, they are shown in Lantern UI. But by persisting them to file, they will not change across upgrades. Maybe we can set it after loading from settings file, or change lantern-ui assets directly for each version.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -94,7 +96,7 @@ func logPanic(msg string) {
 		panic("Error initializing logging")
 	}
 
-	<-logging.Configure("", "", cfg.InstanceId, version, revisionDate)
+	<-logging.Configure("", "", "", version, revisionDate)
```
fffw: We still need the instanceId for logging. `logging` package can also retrieve instance id by itself.

myleshorton: I figured it wasn't a big deal for a panic, but I just added it anyway since I changed the settings.Load call to be in init (so it now has it when it gets here). Thanks @fffw good catch.

fffw: Oh I didn't notice it's in panic wrapper.

src/github.com/getlantern/flashlight/settings/settings.go
fffw: I created a small pr https://github.com/getlantern/lantern/pull/3273 to not persist versions in settings.yaml. Do you think it's ok @myleshorton ?

Other parts are LGTM.

fffw:
All tests passed. Merging.

myleshorton:
Awesome thanks @fffw!!

PR #3275, opened 2015-10-23
adjust wait time for fdcount test

fffw:


fffw:
Trivial changes to make test pass. Will merge directly.

Issue #2875, opened 2015-08-04
Some ad blocker conflicts with Lantern


fffw:
From feedback from an user on FB.

Those ad blocker using [hosts file](http://winhelp2002.mvps.org/hosts.htm) will be affected. Basically, they point the ad sites to 0.0.0.0, which results TCP RST, and be detected by detour as being blocked.

Resolving IP and checking for this specific address can fix it, but it's not clear if it's worth to do it.

fffw:
Some of the ad blockers add entries to 127.0.0.1, instead of 0.0.0.0.

myleshorton:
Ha! Fascinating -- Lantern/detour is routing around ad blockers! I think the ideal solution here is to actually add ad blocking into Lantern, but that could take a little while. Then again, a simple version might be as easy as auto-downloading a domain list from adblocker.

fffw:
It can be fixed by not dialing detour if the host resolves to private IP addresses. Hold it until https://github.com/getlantern/lantern/issues/2470 gets resolved.

Issue #2915, opened 2015-08-10
If Lantern is unable to proxy traffic for a site, it should be added to the detour whitelist so that we don't proxy it


fffw:
We see error `Bad status code on CONNECT response: 403 on pass 0...continuing`.

It's the same error as https://github.com/getlantern/lantern/issues/2395. As we already removed the country list, the cause might not be the same.

> ERROR balancer: balancer.go:99 Unable to dial tcp://mynamedomain.koko:443: Bad status code on CONNECT response: 403 on pass 0...continuing

A quick scan give me an impression that all those domains are not resolvable.

myleshorton:
Yup they look like almost all at least very weird domains. Some are direct IP addresses, others look like they're likely DNS entries for localhost or something else odd like that. Not sure how to solve the latter.

oxtoacart:
It seems that if Lantern encounters something that it definitely can't proxy, we should add it to the detour "whitelist".  I'm renaming the ticket to reflect that.

PR #3282, opened 2015-10-26
[don't merge] add option to detour to skip loopback and LAN address

fffw:
It resolves https://github.com/getlantern/lantern/issues/2875 as well as https://github.com/getlantern/lantern/issues/2915 hopefully. Also refactored detour a bit and fixed the randomly failed test.

There's still one minor issue left: if the loopback or LAN address is unable to dial for some reason, `detour.Dialer` will not return immediately, so caller will wait until it times out. Hope I can fix it tomorrow.

### DIFF
--- a/src/github.com/getlantern/detour/detour.go
+++ b/src/github.com/getlantern/detour/detour.go
@@ -41,13 +41,19 @@ import (
 // If no any connection made after this period, stop dialing and fail
 var TimeoutToConnect = 30 * time.Second
 
-// To avoid unnecessarily proxy not-blocked url, detour will dial detour connection
-// after this small delay. Set to zero to dial in parallel to not introducing any delay.
+// To avoid unnecessarily proxy not-blocked url, detour will dial detour
+// connection after this small delay. Set to zero to dial in parallel and not
+// introduce any delay.
 var DelayBeforeDetour = 0 * time.Millisecond
 
 // If DirectAddrCh is set, when a direct connection is closed without any error,
-// the connection's remote address (in host:port format) will be send to it
+// the connection's remote address (in host:port format) will be send to it.
 var DirectAddrCh = make(chan string)
+
+// When set, try to resolve the address in a short period (50ms), and if it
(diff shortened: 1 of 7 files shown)


src/github.com/getlantern/detour/detour.go
```diff
@@ -137,11 +147,23 @@ func Dialer(detourDialer dialFunc) func(network, addr string) (net.Conn, error)
 
 		// dialing sequence
 		if whitelisted(addr) {
-			dialDetour(network, addr, detourDialer, ch)
+			go func() {
```
fffw: moved all goroutines to `detour.go` to make the intention more explicit.

src/github.com/getlantern/detour/detour.go
```diff
@@ -137,11 +147,23 @@ func Dialer(detourDialer dialFunc) func(network, addr string) (net.Conn, error)
...
+				dialDetour(network, addr, detourDialer, ch)
+			}()
 		} else {
+			tt := time.AfterFunc(TimeoutToConnect, func() {
```
fffw: extract the timer from the loop below (L187 previously)

myleshorton:
Ha very interesting on the ad blocker front @fffw -- I'll have a look at this thanks.

myleshorton:
Oh looks like one of the detour tests is failing @fffw

fffw:
This should also resolve https://github.com/getlantern/lantern/issues/3319.

Seems the test is not reliable enough...

myleshorton:
Wait so is this one now obsolete @fffw?

fffw:
Yeah indeed though lots of code can be reused. Mark it as don't merge until new detour merged.

fffw:
This PR doesn't bypass LAN addresses if user chooses "Proxy all traffic". It will also expose user's access attempts via DNS queries.

One alternative would be to have user supply a list of sites to bypass proxy. It adds complexity on UI but provides a way to user for special requirements: local development sites, slow sites but must accessed inland (https://github.com/getlantern/lantern/issues/3325). It will not resolve https://github.com/getlantern/lantern/issues/2875 though.

Issue #3297, opened 2015-11-04
Add loggly tag to show as a separate source group


fffw:
Related to https://github.com/getlantern/http-proxy/issues/47

We need to have a way to show client and servers separately.

fffw:
Closed via #3298

PR #3298, opened 2015-11-04
add logglyTag

fffw:
It's a rather simple change to show client errors in separate source group.


fffw:
It resolves #3297

fffw:
Thanks @atavism I almost forgot it!

PR #3310, opened 2015-11-13
Add permanent ID

uaalto:
hey @myleshorton @fffw can you take a look at this when you have a chance, please? In the servers, the idea is that this field won't be mandatory, but we will only get usage statistics from those who provide it.

### DIFF
--- a/src/github.com/getlantern/flashlight/settings/settings.go
+++ b/src/github.com/getlantern/flashlight/settings/settings.go
@@ -2,6 +2,7 @@
 package settings
 
 import (
+	"encoding/base64"
 	"io/ioutil"
 	"net/http"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


myleshorton:
Hey sorry for the delay @uaalto - I'll give this a look.

src/github.com/getlantern/flashlight/settings/settings.go
```diff
@@ -51,7 +52,10 @@ func Load(version, revisionDate, buildDate string) {
 		AutoReport: true,
 		AutoLaunch: true,
 		ProxyAll:   false,
-		InstanceID: uuid.New(),
+		// There is no true privacy or security in instance ID.  For that, we rely on
+		// transport security.  Hashing MAC would buy us nothing, since the space of
+		// MACs is trivially mapped, especially since the salt would be known
+		InstanceID: base64.StdEncoding.EncodeToString(uuid.NodeID()),
```
myleshorton: It actually seems like this shouldn't be stored in settings at all since it should never change and since we don't even really want to store it to disk independently from the OS. It seems to me that we should remove InstanceID from the setting struct entirely and just have GetInstanceId be `return base64.StdEncoding.EncodeToString(uuid.NodeID())`

uaalto: Sure, I'll add that change!

fffw:
Gonna merge this as `Lantern-Device-Id` is required by Pro.

Issue #3313, opened 2015-11-16
Do HTTP client proxy connectivity check for all configured proxies, not just loopback


myleshorton:

myleshorton:
This is useful whenever we want to run the proxy on a non-localhost address.

myleshorton:
This could be related to #3312 as well, as that VPS was not running on localhost.

Issue #3312, opened 2015-11-14
Leaked connections to 8787 port


fffw:
Lots of following errors in logs. Should be HTTP requests of config update, geolookup, etc.

> ERROR flashlight.client: server.go:1777 http: Accept error: accept tcp [::]:8787: too many open files; retrying in 1s

And pairs of connections from `lsof`.
```
lantern_l 10698 lantern 1018u  IPv4 492836932      0t0       TCP 127.0.0.1:38809->127.0.0.1:8787 (ESTABLISHED)
lantern_l 10698 lantern 1020u  IPv4 492837714      0t0       TCP 127.0.0.1:38816->127.0.0.1:8787 (ESTABLISHED)
lantern_l 10698 lantern 1022u  IPv6 492836934      0t0       TCP 127.0.0.1:8787->127.0.0.1:38809 (ESTABLISHED)
lantern_l 10698 lantern 1023u  IPv6 492837717      0t0       TCP 127.0.0.1:8787->127.0.0.1:38816 (ESTABLISHED)
```

fffw:
Found on Iran VPS, version 2.0.10

myleshorton:
Did you restart Lantern on that machine @fffw?

myleshorton:
Never mind @fffw I see the restart now.

PR #3314, opened 2015-11-16
Remove unnecessary localhost check when waiting from proxy server clos…

myleshorton:
…es #3313

### DIFF
--- a/src/github.com/getlantern/flashlight/util/http.go
+++ b/src/github.com/getlantern/flashlight/util/http.go
@@ -152,13 +152,11 @@ func (df *dualFetcher) Do(req *http.Request) (*http.Response, error) {
 
 	go readResponses(finalResponseCh, responses, finalErrorCh, errs)
 
-	for {
-		select {
-		case resp := <-finalResponseCh:
-			return resp, nil
-		case err := <-finalErrorCh:
-			return nil, err
-		}
+	select {
+	case resp := <-finalResponseCh:
+		return resp, nil
+	case err := <-finalErrorCh:
+		return nil, err
 	}
 }
 
--- a/src/github.com/getlantern/flashlight/util/http.go
+++ b/src/github.com/getlantern/flashlight/util/http.go
@@ -219,7 +219,6 @@ func httpClient(rootCA string, proxyAddr string, persistent bool) (*http.Client,
 	}
 
 	if proxyAddr != "" {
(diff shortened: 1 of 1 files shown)


myleshorton:
You mind looking at this @fffw? Small change really.

myleshorton:
Hey @fffw I added the fix (hopefully!) for closing all connections (#3312) here as well since it touches the same file.

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -143,30 +143,59 @@ func (df *dualFetcher) Do(req *http.Request) (*http.Response, error) {
...
+	// Create channels for the final response or error. The response channel will be filled
+	// in the case of any successful response as well as a non-error response for the second
+	// response received. The error channel will only be filled if the first response non-successful
+	// and the second is an error.
+	finalResponseCh := make(chan *http.Response, 1)
+	finalErrorCh := make(chan error, 1)
+
+	go readResponses(finalResponseCh, responses, finalErrorCh, errs)
+
+	for {
```
fffw: This for loop is unnecessary.

src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -143,30 +143,59 @@ func (df *dualFetcher) Do(req *http.Request) (*http.Response, error) {
...
+				log.Debug("Got second response -- sending")
+				finalResponse <- resp
 			} else if success(resp) {
-				log.Debugf("Got good response")
-				// Returning preemptively here means the second response
-				// will not be closed properly. We need to ultimately
-				// handle that.
-				return resp, nil
+				log.Debug("Got good response")
+				finalResponse <- resp
```
fffw: Inspired by a golang.org blog article, an alternative way would be to spawn a goroutine to receive response from channel and close the body, so we can return directly here instead of sending to channel.

Current method is good enough, don't bother change it.

fffw:
LGTM except that extra `for` loop. I'll do that minor change and merge.

src/github.com/getlantern/flashlight/util/http.go
myleshorton: Interesting. Which article @fffw? I was going to make a little change to make this code more straightforward actually -- another PR coming soon.

PR #3326, opened 2015-11-26
tunio integration

xiam:
Hey @atavism,

This PR provides support for tunio via the `USE_LANTERN_SOCKS` configuration constant in VpnBuilder.java. It would be awesome if we can find another way to change this at compile time, it feels a bit weird, any suggestion?. In order to use tunio, the `USE_LANTERN_SOCKS` value must be set to `false`.

This is up to date with the latest `lantern-mobile` commit. Could you help me checking it?


atavism:
Hey @xiam, yep, I agree it's awkward maintaining the config from there which is why I spent the weekend refactoring the config setup, and now we're finally using the same code as Lantern desktop. We can start updating that value from ``settings.yaml`` which gets copied from the Android resources directory to the app internal storage directory on start; those settings will also be available on the Java side. I'll fix some of the resulting conflicts on your branch and then review this next week.

xiam:
That's awesome @atavism! I'm going to check out your refactor and rebase accordingly.

PR #3331, opened 2015-12-01
only render error page to browsers

### LINKED ISSUE #3330: Not render ErrorAccessingPage for requests from Lantern itself
It obfuscates the actual error message in Loggly with some strange strings like:

```
5pt; } </style> </head> <body> <div id="frame"> <img src="data:image/png;b
```

or

```
image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAKwAAADTCAYAAAAcRfjMAAAABHNCSVQICAgIfAhkiAAAAAlwSFlzAAALEwA
```

fffw:
Resolves #3330 to eliminate those weird meaningless Loggly entries.

@xiam mind take a look? It's a slight change to `errorRewritingRoundTripper`. Changes to `flashlight_test.go` are simply making `go vet` happy.

By requesting an inexist port with different user agent via Lantern, we can see the effect.
`curl -vLx localhost:8787 http://twitter.com:8000`
`curl -vLx localhost:8787 http://twitter.com:8000  -H "User-Agent: Mozilla/5.0"`

### DIFF
--- a/src/github.com/getlantern/flashlight/client/reverseproxy.go
+++ b/src/github.com/getlantern/flashlight/client/reverseproxy.go
@@ -131,29 +131,31 @@ type errorRewritingRoundTripper struct {
 func (er *errorRewritingRoundTripper) RoundTrip(req *http.Request) (resp *http.Response, err error) {
 	res, err := er.orig.RoundTrip(req)
 	if err != nil {
-		agent := req.Header.Get("User-Agent")
+		var htmlerr []byte
+
 		// Virtually all browsers mark themselves as Mozilla compatible
 		// while Opera has a user selectable option.
 		// Ref http://webaim.org/blog/user-agent-string-history/
 		// We only render user friendly error page to browsers.
+		agent := req.Header.Get("User-Agent")
 		isBrowser := strings.HasPrefix(agent, "Mozilla") || strings.HasPrefix(agent, "Opera")
-		if !isBrowser {
-			return res, err
-		}
-
-		// It is likely we will have lots of different errors to handle but for now
-		// we will only return a ErrorAccessingPage error.  This prevents the user
-		// from getting just a blank screen.
-		htmlerr, err := status.ErrorAccessingPage(req.Host, err)
-
-		if err != nil {
-			log.Debugf("Got error while generating status page: %q", err)
+		if isBrowser {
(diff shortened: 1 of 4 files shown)


xiam:
Hey @fffw, would it be possible to identify Lantern initiated requests instead of trying to guess if the user is using a browser by checking the user-agent?

fffw:
By passing a customized header for HTTP requests in `util/http.go`, `errorRewritingRoundTripper ` can identify them. It is the way I implemented first which requires changes in two places and adds a bit of complexity.

Checking user agent is way simpler and makes sense to me. An error page is useful only if user can see it, while rendering such page to non-browser applications adds confusion I think.

All browsers regardless of desktop or mobile follow that implicit rule and unlikely to change in the future, and it's not that harmful if it really changes.

How do you think @xiam ?

xiam:
I think that in general is safer to test for the presence of the exception (like a `req.Header.Get("X-Lantern-Special-Header") header`) rather than for a list of possibly accepted escenarios (like relying on the browser's user-agent, which could be stripped or modified by a privacy tool or something like that).

We could say that all browsers that *follow the standard* are going to send either a "Mozilla" or a "Opera" string, but we can be *sure* that no other thing besides Lantern is sending a `X-Lantern-Special-Header` header because only we use it.

Lantern core is the affected part, so probably we should find a solution within Lantern core, instead of relying on external data out of our control, like the user's choice of browser or privacy tool.

fffw:
I'd like to treat the checking of `User-Agent` as part of the feature, rather than an exception. We want to render error page only if it makes sense to user.

You reminded me another more precise approach. We can choose to *not* render error page if the request has an `Accept` header other than `text/html` or `application/xhtml+xml`, then explicitly add appropriate `Accept` header in Lantern initiated requests. Error page is meaningless if the requested resource is an image or json. If there's no `Accept` header, we still render error page.

Through this way, each part of code conforms to HTTP standard rather than depends on each other.

xiam:
The `Accept` suggestion sounds fantastic!

fffw:
I made that change. @xiam mind take a look again?

xiam:
@fffw looks good to me :)

fffw:
Thanks @xiam !

PR #3332, opened 2015-12-01
set 500ms DelayBeforeDetour by default resolves #3325

### LINKED ISSUE #3325: Lantern never use direct connection for non-CONNECT requests
Detour uses whichever connection made first and ignore another, but [reverseproxy.go#L55](https://github.com/getlantern/lantern/blob/3a907ed76aeb50aeb948997f485727c668f1c7d6/src/github.com/getlantern/flashlight/client/reverseproxy.go#L55) returns detour connection immediately, so direct connection will always be late.

fffw:
Ok this partly resolves https://github.com/getlantern/lantern/issues/3325.

During the ~ 1h test on the China box, with a few exceptions, all non-blocked sites are dialed directly. Those exceptions are much slower sites with ~ 2s to have a connection.

But on Iran box, most Farsi sites took > 1s to connect, the only exceptions were blogfa.com and watchmojo.com. Good news is: Lantern indeed accelerated many sites in test set. www.irna.ir and www.sanjesh.org were 2x faster, www.ut.ac.ir was 7x faster, by running `time curl -vLx localhost:8888 <site>` and `time curl -vL <site>` respectively (in that order).

This workaround seems ok to me. How do you think @myleshorton?

### DIFF
--- a/src/github.com/getlantern/detour/detour.go
+++ b/src/github.com/getlantern/detour/detour.go
@@ -43,7 +43,7 @@ var TimeoutToConnect = 30 * time.Second
 
 // To avoid unnecessarily proxy not-blocked url, detour will dial detour connection
 // after this small delay. Set to zero to dial in parallel to not introducing any delay.
-var DelayBeforeDetour = 0 * time.Millisecond
+var DelayBeforeDetour = 500 * time.Millisecond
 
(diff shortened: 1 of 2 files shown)


fffw:
Close in favor of https://github.com/getlantern/lantern/pull/3360

PR #3356, opened 2015-12-17
add pprofaddr option resolves #3291

### LINKED ISSUE #3291: trace all goroutines
And trace out detailed list. Can be turned on via flag as a developer tool.

It can be useful to debug leaking goroutines (we saw thousands of them from panic dump) and connections ("too many open files" after running for hours to days).

`runtime.NumGoroutine()` is useful, but for a detailed list, we may need the `goroutine` profile in `runtime/pprof`.

fffw:
It's a very useful debug tool.

### DIFF
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -7,6 +7,7 @@ import (
 	"math/rand"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


myleshorton:
You're seeing this issue with libappindicator3-dev on CI @fffw right?

myleshorton:
Merging since the libgtk issue is fixed on valencia.

PR #3360, opened 2015-12-18
rewrite detour

### LINKED ISSUE #3296: New detour logic tends to leave more CLOSE_WAIT connections

### LINKED ISSUE #3325: Lantern never use direct connection for non-CONNECT requests

### LINKED ISSUE #3343: The website not blocked in China also use lantern.
Sometimes, when I visit some web sile like `tower.im`, it's not blocked in China.
But the lantern will proxy it, and make the web site very slow.

fffw:
Resolves #3296, resolves #3343, resolves #3325 

It's a total rewrite. `ioLoop` sets a barrier for asynchronous events and spawn goroutines to handle them immediately. pipelining and merging are used to handle dialing and read operations.

It seems more fluent than current valencia code and 2.0.10 from my own test. It doesn't leave any CLOSE_WAIT connections. goroutine number doesn't increase either.

To inspect goroutines, run `./lantern_darwin_amd64 -pprofaddr=localhost:3000` and access `http://127.0.0.1:3000/debug/pprof/` from browser.

To have a sense of memory allocation, `go tool pprof -top http://127.0.0.1:3000/debug/pprof/heap`


src/github.com/getlantern/detour/ioloop.go
```diff
@@ -0,0 +1,315 @@
...
+				}
+				reRead = true
+			}
+			chConns <- c
+			// spawn goroutine after above statement so channel
+			// operations are in determined order.
+			if reRead {
+				go func() {
+					log.Tracef("Re-read from %s connection to %s", c.Type(), dc.addr)
+					buf := make([]byte, len(firstReadReq.buf))
```
fffw: We need separate buffers to avoid race condition. `sync.Pool` can be employed to avoid frequent memory allocation, but let's leave it for future optimization.

fffw:
It works pretty stable except:

* It will not add the sites of non-idempotent requests (POST/PUT/PATCH) to whitelist. Should add to or remove from whitelist in a single place.

* A few false positives if the site is slow accessing directly but faster through proxy. It causes problem if the site only allow access from certain area.

fffw:
I traced open connections from detour, the count is not identical to `lsof -nPp <pid>`. There's no hanging goroutines either. So there should be connection leaks somewhere else. We may decrease the `idleTimeout` value in `client/chain.go` to close idle connections sooner until we find out the actual cause.

src/github.com/getlantern/detour/ioloop.go
```diff
@@ -0,0 +1,348 @@
...
+	chRemoveConn := make(chan conn)
+	// Hold the read request so we can re-read after replay.
+	var firstReadReq *innerReadRequest
+
+	// Can we replay this request
+	// TODO: How to check when using pipelining or HTTP/2?
+	var nonidempotentHTTPRequest bool
+
+	for {
+		select {
```
fffw: After Lantern running for hours, 100+ goroutines blocked here, which means the caller doesn't close `detour.Conn` at all! Let's see how many goroutines will accumulate after running for days.

oxtoacart:
@fffw Do you want me to review?

fffw:
Yeah @oxtoacart that would be great!

oxtoacart:
@fffw I'm planning to review this week.

src/github.com/getlantern/detour/ioloop.go
```diff
@@ -0,0 +1,348 @@
...
+
+var nonidempotentMethods = [][]byte{
+	[]byte("PUT "),
+	[]byte("POST "),
+	[]byte("PATCH "),
+}
+
+// Ref https://tools.ietf.org/html/rfc2616#section-9.1.2
+// We consider the https handshake phase to be idemponent.
+func isNonidempotentHTTPRequest(b []byte) bool {
```
oxtoacart: This is scary for multiple reasons.

1. What about HTTPS where the verb is encrypted?  This method will identify it as idempotent even though it may not be.

2. What if an application author is misusing an idempotent method like GET to actually alter server-side state?

oxtoacart:
In testing, this seems to sometimes hang when I try to proxy traffic through it.  I'm working on understanding the code better so I can hopefully provide more information.

src/github.com/getlantern/detour/ioloop.go
fffw: For HTTPS, the first a few bytes written to a connection are TLS Client Hello, no HTTP verb at this moment. I'm not 100% sure but an duplicated HTTPS handshake packet should not do much harm.

For 2, it will cause problem in that case. Replay is very important to detour. I just didn't find a more unobstructed way to implement it.

oxtoacart: 1. Makes sense

2. Let's not worry about it, as it looks like Chrome will actually retry idempotent requests too, so we're not doing anything fundamentally new :)

oxtoacart:
@fffw Do you know of a good way to simulate the blockages that detour should handle so that I can do some manual testing with it?

fffw:
@oxtoacart I don't have better way except experimenting in real environment. `detect.go` didn't change at all since 2.0.0, the blockage detection should be time proven.

oxtoacart:
@fffw Okay, I'll assume the detection is good.  So I think the only thing left is the hanging.  I'll see if I can reproduce reliably.

src/github.com/getlantern/detour/ioloop.go
```diff
@@ -0,0 +1,348 @@
...
+			}
+
+		case r := <-chRemoveConn:
+			conns.Remove(r)
+			atomic.AddUint32(&dc.expectedConns, ^uint32(0))
+
+		case req := <-dc.chReadRequest:
+			if atomic.LoadUint32(&dc.expectedConns) == 1 {
+				c := conns.Next()
+				go func() {
```
oxtoacart: It's kind of expensive to spawn a goroutine for every call to read, isn't it?

src/github.com/getlantern/detour/ioloop.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+type innerReadRequest struct {
+	buf      []byte
+	chResult chan innerReadResult
+}
+
+// ioLoop is the core of detour. It waits for connections and handles
+// read/write requests
+func (dc *Conn) ioLoop() {
```
oxtoacart: Because reads and writes are handled by the same goroutine, this ends up being half-duplex.  For TLS negotation and typical HTTP traffic, that's not a problem, but if someone is proxying stuff through lantern that benefits from full duplex, this might be an issue.

fffw:
Hanging happens more frequently when concurrency level increases, and when latency increases. Hanged requests are usually less than 5. Requests with concurrency level of 1 never experience hanging.

There are two places goroutines were blocked on: [ioloop.go#L42](https://github.com/getlantern/lantern/blob/issue3296/src/github.com/getlantern/detour/ioloop.go#L42) and [detour.go#L105](https://github.com/getlantern/lantern/blob/issue3296/src/github.com/getlantern/detour/detour.go#L105). There should be two different issues, as it's impossible to block on both the send side and receive side of an channel (`chReadRequest`).

src/github.com/getlantern/detour/dialer.go
```diff
@@ -0,0 +1,104 @@
...
+					atomic.AddUint32(&dc.expectedConns, ^uint32(0))
+					continue
+				}
+				// prevent goroutine from blocking if
+				// I/O loop already exited, or lots of
+				// CLOSE_WAIT connections will be left.
+				select {
+				case dc.chConnToIOLoop <- res.c:
+					chLastError <- nil
+				default:
```
fffw: Something like this is definitely a bad idea, see https://golang.org/ref/mem#tmp_7. It's a shame that I didn't read doc carefully until now.

I made a simple demo at http://play.golang.org/p/8VJqfmzZnO. By changing the sleep time, sometimes it shows correct result, sometimes it reports deadlock.

fffw:
Close for now. Will create new PR to `devel` branch.

PR #3364, opened 2015-12-24
By install deltarpm , boost docker yum update

netroby:
If installed deltarpm, yum update will faster

Delta RPMs reduced 50 M of updates to 16 M (67% saved)

Signed-off-by: ZhiFeng Hu <hufeng1987@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,8 @@ ENV WORKDIR /lantern
 ENV GO_PACKAGE_URL https://storage.googleapis.com/golang/go1.4.3.linux-amd64.tar.gz
 
 # Updating system.
+RUN yum makecache
+RUN yum install -y deltarpm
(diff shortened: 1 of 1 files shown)


fffw:
Tens of megabytes are not a huge save considering the total download size to `make docker`, but it will invalidate all existing docker images. We may keep it open and merge when we have other significant changes to Dockerfile.

netroby:
It's about the time, download packages need more times to wait. decrease waiting time can save several minutes to finish the build progress.

fffw:
The Dockerfile needs to be built only once. Subsequent `make` will be in seconds. That's why it's important to not invalidate existing docker images.

xiam:
Thanks @netroby, I added your suggestion as part of this other PR here: https://github.com/getlantern/lantern/pull/3395.

PR #3389, opened 2016-01-04
return DIRECT for non-HTTP url in PAC

### LINKED ISSUE #3388: Check for url scheme in PAC script
Or browser will direct non-HTTP requests (FTP as an example) to Lantern which is not able to handle them.

fffw:
Fixes #3388


myleshorton:
LGTM. A lantern-mobile test is failing, but it's clearly unrelated.

Issue #3371, opened 2015-12-28
Consider upgrading to Go 1.5


oxtoacart:
Go 1.5 has improved garbage collection (does some collection concurrently now) and also better support for cross-compilation.

xiam:
True @oxtoacart, I recall we had a problem running a binary built with go1.5 when it was still beta but I guess we can check again.

oxtoacart:
Definitely not something I would consider high priority, just a thought that I didn't want to lose :)

xiam:
@oxtoacart It looks like the latest go1.5 is working well on all platforms. See https://github.com/getlantern/lantern/pull/3395

xiam:
https://github.com/getlantern/lantern/pull/3395

oxtoacart:
@xiam That was quick!

PR #3378, opened 2015-12-31
Build/README updates

oxtoacart:
This has a few updates:

1. Support use of newest Docker Tools on Mac.  The script should still work with boot2docker, but it will use the newer docker-machine command if available.

2. Support building mobile locally.

3. Made pre-push hook work on pre-push (as originally written, it didn't seem to work for pre-push because it was using git status which only looks at the working tree, not comparing remote to origin)

4. Added some more details to README

PR #3395, opened 2016-01-08
Updating Dockerfile to use go1.5

xiam:
See https://github.com/getlantern/lantern/issues/3371

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ RUN yum install -y glibc-devel.i686 glib2-static.i686 glibc-2.20-8.fc21.i686 lib
 # ARM EABI toolchain must be grabbed from an contributor repository, such as:
 # https://copr.fedoraproject.org/coprs/lantw44/arm-linux-gnueabi-toolchain/
 RUN yum install -y yum-utils && \
+  rpm --import https://copr-be.cloud.fedoraproject.org/results/lantw44/arm-linux-gnueabi-toolchain/pubkey.gpg && \
   yum-config-manager --add-repo=https://copr.fedoraproject.org/coprs/lantw44/arm-linux-gnueabi-toolchain/repo/fedora-21/lantw44-arm-linux-gnueabi-toolchain-fedora-21.repo && \
   yum install -y arm-linux-gnueabi-gcc arm-linux-gnueabi-binutils arm-linux-gnueabi-glibc && \
   yum clean packages
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,8 +11,7 @@ RUN mkdir -p $WORKDIR
 RUN mkdir -p $SECRETS
 
 # Updating system.
-RUN yum update -y && yum makecache && yum clean packages
-RUN yum install -y deltarpm && yum clean packages
+RUN yum install -y deltarpm && yum update -y && yum clean packages
 
 # Requisites for building Go.
 RUN yum install -y git tar gzip curl hostname && yum clean packages
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,20 +17,21 @@ ENV SECRETS /secrets
 ENV GO_PACKAGE_URL https://storage.googleapis.com/golang/go1.5.2.linux-amd64.tar.gz
 
 # Updating system.
-RUN yum -y update && yum clean all
+RUN yum update -y && yum makecache && yum clean packages
+RUN yum install -y deltarpm && yum clean packages
 
 # Requisites for building Go.
-RUN yum install -y git tar gzip curl hostname && yum clean all
+RUN yum install -y git tar gzip curl hostname && yum clean packages
 
 # Compilers and tools for CGO.
-RUN yum install -y gcc gcc-c++ libgcc.i686 gcc-c++.i686 pkg-config && yum clean all
+RUN yum install -y gcc gcc-c++ libgcc.i686 gcc-c++.i686 pkg-config && yum clean packages
 
 # Getting Go.
 RUN curl -sSL $GO_PACKAGE_URL | tar -xvzf - -C /usr/local
 
 # Requisites for bootstrapping.
-RUN yum install -y glibc-devel glibc-static && yum clean all
-RUN yum install -y glibc-devel.i686 glib2-static.i686 glibc-2.20-8.fc21.i686 libgcc.i686 && yum clean all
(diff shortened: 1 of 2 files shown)


xiam:
Looks like the go1.5 problem we had with the beta versions is gone. Would you like to check this out @fffw, @oxtoacart?

fffw:
As we are making significant changes to Dockerfile, it's an opportunity to consolidate `yum` packages, so future changes of go version will not invalidate the docker image. It's also helpful to developers with slow Internet connection by adding `deltarpm` (https://github.com/getlantern/lantern/pull/3364).

xiam:
Sounds good @fffw, we could try that as well.

myleshorton:
Hey this generally looks good to me, but it seems like we might as well do this along with @oxtoacart's https://github.com/getlantern/lantern/pull/3378 at the same time. That also upgrades the go version I believe.

xiam:
Hey @myleshorton, Not sure if we should do that, #3378 and this one attack two different problems, this only affects the main `Dockerfile` and focuses on using a newer Go version, while #3378 bumps the Go version from go1.5.1 to go1.5.2 for gomobile only (it's a different `Dockerfile`).

The thing that I can do is create another PR that targeting `valencia_ox` instead of `valencia`, sounds good?

oxtoacart:
Either way - if they don't touch each other might as well keep them
separate!

xiam:
@oxtoacart Cool, let's keep them separated then.

xiam:
Hey @fffw, I added that deltarpm suggestion and also changed `yum clean all` to `yum clean packages` and noticed a difference, how does it look now?

Dockerfile
```diff
@@ -1,93 +1,74 @@
...
+RUN yum install -y osslsigncode mingw32-nsis && yum clean packages
 
 # Required for compressing update files
-RUN yum install -y bzip2 && yum clean all
+RUN yum install -y bzip2 && yum clean packages
 
 # Requisites for genassets.
-RUN yum install -y nodejs npm && yum clean all
+RUN yum install -y nodejs npm && yum clean packages
 RUN npm install -g gulp
```
fffw: @xiam Do you think we can move all `yum install`, `gem install` and `npm install` commands to the beginning of Dockerfile so they will not get invalidated when we bump go version again?

xiam: @fffw Good idea, I'm gonna move things that do not change that often to the beginning of the file.

xiam:
Hey @fffw, I just pushed that last one.

Dockerfile
```diff
@@ -1,93 +1,72 @@
...
 
-# Go binary for bootstrapping.
-ENV GO_PACKAGE_URL https://storage.googleapis.com/golang/go1.4.3.linux-amd64.tar.gz
+RUN mkdir -p $WORKDIR
+RUN mkdir -p $SECRETS
 
 # Updating system.
-RUN yum -y update && yum clean all
+RUN yum update -y && yum makecache && yum clean packages
+RUN yum install -y deltarpm && yum clean packages
```
fffw: I found that `deltarpm` only works for `yum update`, so we need to change these two lines as below to have it take effect.

```
RUN yum install -y deltarpm && yum update -y && yum clean packages
```

It reduced 50 M of updates to 16 M (67% saved), not a significant save considering lots of `yum install` hereafter, but we can also just keep it.

I had no knowledge of `makecache` but seems we benefit litle from it as we didn't supply `-C` option to `yum install` (ref [Red Hat doc](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Deployment_Guide/sec-Working_with_Yum_Cache.html)).

xiam: I though that `deltarpm` was like a compressed bindiff, but I guess you're right, it seems like in this particular case it will only work for `yum update`. I've just added that line, thanks!

The `-C` flag would case yum to look into the cached data only but that's not what we want. `makecache` should reduce the times yum tries to update its database, it should read packages and dependencies from disk automatically instead of trying yo update the yum database each time it's called.

Dockerfile
```diff
@@ -1,93 +1,71 @@
...
-RUN yum install -y glibc-devel.i686 glib2-static.i686 glibc-2.20-8.fc21.i686 libgcc.i686 && yum clean all
+RUN yum install -y glibc-devel glibc-static && yum clean packages
+RUN yum install -y glibc-devel.i686 glib2-static.i686 glibc-2.20-8.fc21.i686 libgcc.i686 && yum clean packages
 
 # Requisites for ARM
 # ARM EABI toolchain must be grabbed from an contributor repository, such as:
 # https://copr.fedoraproject.org/coprs/lantw44/arm-linux-gnueabi-toolchain/
-RUN yum install -y yum-utils && yum-config-manager --add-repo=https://copr.fedoraproject.org/coprs/lantw44/arm-linux-gnueabi-toolchain/repo/fedora-21/lantw44-arm-linux-gnueabi-toolchain-fedora-21.repo && yum install -y arm-linux-gnueabi-gcc arm-linux-gnueabi-binutils arm-linux-gnueabi-glibc && yum clean all
+RUN yum install -y yum-utils && \
+  yum-config-manager --add-repo=https://copr.fedoraproject.org/coprs/lantw44/arm-linux-gnueabi-toolchain/repo/fedora-21/lantw44-arm-linux-gnueabi-toolchain-fedora-21.repo && \
```
fffw: I encountered following warning:

> warning: /var/cache/yum/x86_64/21/lantw44-arm-linux-gnueabi-toolchain/packages/arm-linux-gnueabi-filesystem-2-1.fc21.noarch.rpm: Header V3 RSA/SHA1 Signature, key ID 1bf83516: NOKEY
> 
> Public key for arm-linux-gnueabi-filesystem-2-1.fc21.noarch.rpm is not installed

It is misleading when I check for failed build caused by other errors. I suggest we add below line before L30.
```
  rpm --import https://copr-be.cloud.fedoraproject.org/results/lantw44/arm-linux-gnueabi-toolchain/pubkey.gpg && \
```

fffw:
@xiam All good except the warning I mentioned above.

Dockerfile
xiam: Thanks @fffw, I added this other line as well.

fffw:
LGTM! The failed test is peerscanner, which is unrelated.

Issue #3409, opened 2016-01-19
Add support for tracking statistics on most frequently visited sites to http-proxy-lantern


oxtoacart:
If we can use Google Analytics, that might be a nice way to track this.

Per request of @myleshorton

oxtoacart:
I've added a new property in Google Analytics called "Proxied Sites" where we track access to sites as if they were page views.

oxtoacart:
@aranhoide I'd like to try this on a limited basis by pushing to one of our cloud servers.  What would be the best way to do this?  Just manually upload?

aranhoide:
Do you want to try it in a live server, or would a test server do?  In either case, just uploading and substituting `/home/lantern/http-proxy` with your version should work.  But if it's a live server let me know which one so I won't clobber your version in a salt update.

oxtoacart:
Okay, I've deployed this to one server (fp-doams3-20151201-011) and it seems to be working okay.  We're getting way more ip addresses than host names, which is unexpected, but could just be due to what that particular client is doing.

@aranhoide Once this is pulled into master, how can I go about getting it pushed to all servers?  Assuming I do this quickly enough, I won't worry about my manual update getting clobbered :)

aranhoide:
Since this involves a http-proxy update, the process would be:

- make a new release of the `http-proxy-lantern` project,

- change the http-proxy salt configuration to use that release:

https://github.com/getlantern/lantern_aws/blob/master/salt/http_proxy/install_http_proxy.py#L8

- deploy that Salt change into each production cloudmaster (in lantern_aws/bin):

    ./update.py  # Amsterdam
    DC=dosgp1 ./update.py  # Singapore
    DC=vltok1 ./update.py  # Tokyo

- Log into each of these cloudmasters:

    ./ssh_cloudmaster.py
    DC=dosgp1 ./ssh_cloudmaster.py
    DC=vltok1 ./ssh_cloudmaster.py

- In theory, running `salt -b 10 "fp-*" state.highstate` should update all proxies, and that often works in Singapore.  In practice, this is very unreliable with the number of proxies we have in Amsterdam and Tokyo, so I do a cycle of updates and checks that takes forever (but is mostly a background job so I can do other things in the meanwhile).

[...]

PR #3418, opened 2016-01-21
getlantern/lantern#3409 Added ability to force use of a specific prox…

oxtoacart:
…y and token to support testing


myleshorton:
There are still test issues with the direct domain fronting stuff, but they're clearly unrelated to this PR.

PR #3428, opened 2016-01-23
Turn on mobile analytics

atavism:
This records a new session on the back-end for mobile


myleshorton:
I think the frontend analytics are fine and not including these on the backend is better to avoid double counting. Closing!

PR #3429, opened 2016-01-24
support minimum OSX 10.7

fffw:
Hey @xiam this change should work.
```
> otool -l ./lantern_darwin_amd64  | grep -A 3 LC_VERSION_MIN_MACOSX
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.7
      sdk 10.11
```

`man clang` gives me the clue. golang doesn't recognize `-mmacosx-version-min` as CGO hint. Fortunately environment variable works.
```
MACOSX_DEPLOYMENT_TARGET
           If -mmacosx-version-min is unspecified, the default deployment target is read from this environment variable.  This option only affects darwin
           targets.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 SHELL := /bin/bash
 
+MIN_OS_VERSION_MAC := 10.7
+
 DOCKER := $(shell which docker 2> /dev/null)
 GO := $(shell which go 2> /dev/null)
 NODE := $(shell which node 2> /dev/null)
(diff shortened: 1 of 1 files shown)


myleshorton:
Great work @fffw!

xiam:
This is awesome @fffw! I think we should make it a global and put it at the beginning on the file:

```
MACOSX_DEPLOYMENT_TARGET = 10.7
```

fffw:
I thought about that but it only applies to that particular target, plus there's no other relevant environment variables to be grouped together. Any strong reason to do so @xiam ?

xiam:
@fffw I just think that something important like this should be evident and in plain sight. It works fine if we leave it where it is now but I am just concerned about it not being too visible.

fffw:
Defining a variable far away from the place to use it seems unnatural to me. But to make it explicit, I added a separate variable at the top of the file. Does it look good to you @xiam ?

xiam:
Thanks!

PR #3433, opened 2016-01-25
Refreshed masquerades

oxtoacart:
I ran genconfig to get more recent default masquerades.  I also changed direct.go to try all masquerades.


src/github.com/getlantern/fronted/direct.go
```diff
@@ -135,42 +135,49 @@ func (d *direct) Do(req *http.Request) (*http.Response, error) {
...
+					log.Debugf("Connection to %s via %s idle for %v, closing", addr, conn.RemoteAddr(), idleTimeout)
+					if err := conn.Close(); err != nil {
+						log.Debugf("Unable to close connection: %v", err)
+					}
+				})
+				log.Debug("Returning connection")
+				return conn, nil
+			}
+		default:
+			if gotFirst {
```
myleshorton: It looks like it's possible to never get here based on https://github.com/getlantern/lantern/pull/3433/files#diff-e149e4a7348208a557074a227404c192R157 -- i.e. with the candidatesCh repopulated with candidates that will always be re-added

oxtoacart: Correct.  Is that a problem?

myleshorton: well just mean we'd have a dangling go routine. not a huge problem i suppose.

oxtoacart: We could add a timeout too if we want, but yeah, I wouldn't worry about it too much, as in this case we'll have bigger problems than the dangling goroutine (namely that we can't domain front anything!).

myleshorton: True! Ok I just realized only cloudfront was disabled from the peerscanner tests. I'm just going to disable all of it directly on this branch to get this PR pipeline moving..

myleshorton:
OK seems there are periodic issues with some of the mobile tests still, but merging.

PR #3435, opened 2016-01-25
Disabled cfr test since we're not using peerscanner at the moment

oxtoacart:


myleshorton:
LGTM. It strikes me it already isn't used in the Makefile, but should be commented out here too.

PR #3448, opened 2016-01-27
Valencia embeddable

oxtoacart:
This is a fairly large refactor that's intended to prepare lantern for easier embedding as a library (in preparation for lantern-mobile).  It includes a bunch of changes:

1. Provides a Run method in flashlight that encapsulates the standard lifecycle for starting a client proxy
1. Provides the ability to add application-specific logic through lifecycle callbacks like beforeStart, etc.
1. Starts client proxy on a random port
1. Removes dependency from core proxying functionality to settings (which is desktop-app specific)
1. Removes server functionality (which is unused)
1. Removes statistics reporting and statistics ui functionality (which is unused)
1. Removes geolookup UI support (which is unused). In the processed, eliminated repetitive geolookup calls since app doesn't actually need constant geo updates.
1. Removed old form of lantern-mobile (which has long since dramatically changed on the lantern-mobile branch anyway)


oxtoacart:
cc: @atavism

myleshorton:
The one issue I'm seeing is that it looks like some timing issue on startup has changed and that it's possible for the UI to open before the backend is ready -- just got an open UI with a completely blank screen

myleshorton:
Oh, actually scratch that -- Lantern was already previously running. That used to work, however, so it's possible something broke with the logic for opening up an existing lantern @oxtoacart

oxtoacart:
@myleshorton Nice catch!  What steps can I use to reproduce?

myleshorton:
Could be an 8787 thing but not sure. Basically start with Lantern already running and try running it again..."works" on valencia but not here

oxtoacart:
Must be a timing thing - I haven't been able to reproduce yet.

myleshorton:
Yeah to be clear @oxtoacart I think it's actually an issue with the second lantern instance not properly telling the first lantern instance to pop up the UI. We mostly do that so that if a user double clicks on a desktop shortcut on Windows, for example, it will open up the UI.

myleshorton:
Ha! Yeah that's it I think -- if the first was started on valencia and the second on valencia-embeddable, this can happen @oxtoacart, so probably not something to worry about.

oxtoacart:
Nice edge case @myleshorton !

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -50,18 +50,31 @@ type Client struct {
...
+// on localhost. onListeningFn is a callback that gets invoked as soon as the
+// server is accepting TCP connections.
 func (client *Client) ListenAndServe(onListeningFn func()) error {
+	log.Debug("About to listen")
 	var err error
 	var l net.Listener
 
-	if l, err = net.Listen("tcp", client.Addr); err != nil {
-		return fmt.Errorf("Client proxy was unable to listen at %s: %q", client.Addr, err)
+	if l, err = net.Listen("tcp", "localhost:0"); err != nil {
```
myleshorton: Yeah so the big problem with this change @oxtoacart is people using things like SwitchySharp and SwitchyOmega who know to set it to use Lantern on 8787. That's a really common use case. This also disables people setting local network devices to run through 8787 on Lantern running on another machine.

myleshorton: We could try 8787 and then use a random port if we can't bind... We should probably at least support running on 0.0.0.0 if the user specifies it from the command line though (even though I believe that's broken on the current valencia).

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -259,8 +259,12 @@ func Update(mutate func(cfg *Config) error) error {
 }
 
 // InConfigDir returns the path to the given filename inside of the configdir.
-func InConfigDir(filename string) (string, string, error) {
-	cdir := *configdir
+func (cfg *Config) InConfigDir(filename string) (string, string, error) {
+	return inConfigDir(cfg.configDir, filename)
+}
```
myleshorton: Looks like this isn't used publicly at all @oxtoacart -- just kill it?

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -1,147 +1,103 @@
...
-			return nil
-		}
-		log.Trace("Sending last known location to new client")
-		return write(country)
+// Refresh refreshes the geolookup information by calling the remote geolookup
+// service. It will keep calling the service until it's able to determine an IP
+// and country.
+func Refresh(proxyAddrFN eventual.Getter) {
+	select {
+	case refreshRequest <- proxyAddrFN:
```
myleshorton: So I believe this will re-queue a request as long as there's not a current one in-flight correct? In practice we only call Refresh once on startup AFAICT, and I'm not sure the existing behavior is necessarily wrong but maybe just not super intuitive (i.e. what are we really trying to do here? call this once? call it if an existing call isn't in flight?)

oxtoacart: Refresh just means "get up-to-date geolocation information".  In practice we only care about that at startup, but as far as the geolookup module is concerned, it could be called anytime.  Failing to enqueue when a refresh is already in progress is just a way of avoiding unnecessary work.

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -1,147 +1,103 @@
...
+func init() {
+	go run()
+}
 
-	if err != nil {
-		log.Errorf("Could not lookup IP %v", err)
-		return "", "", err
+func run() {
+	for proxyAddr := range refreshRequest {
+		gi := lookup(util.NewChainedAndFronted(proxyAddr))
```
myleshorton: So chainedAndFronted contains logic that will make it only use chained servers if they've succeeded on a previous call. If we create new one each time, though, that won't work.

src/github.com/getlantern/flashlight/config/config.go
oxtoacart: Killed.

src/github.com/getlantern/flashlight/client/client.go
oxtoacart: Fixed - caller can now specify an address explicitly, or just let Lantern randomly assign one.  In desktop app, I explicitly request localhost:8787, in mobile we'll let it randomly assign.

src/github.com/getlantern/flashlight/geolookup/geolookup.go
oxtoacart: Fixed.

oxtoacart:
@myleshorton Thanks for the thorough review!  I believe I've addressed all your comments.

myleshorton:
Cool thanks @oxtoacart! I unfortunately didn't quite finish and still just
want to go through things a bit more - should be able to do it this weekend!

oxtoacart:
Thanks @myleshorton.  This can definitely wait until next week - enjoy your weekend!

src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -55,26 +44,15 @@ func (client *Client) initBalancer(cfg *ClientConfig) {
 	}
 
 	bal := balancer.New(dialers...)
-
-	if client.balInitialized {
-		log.Trace("Draining balancer channel")
-		old := <-client.balCh
+	oldBal, ok := client.bal.Get(25 * time.Millisecond)
```
myleshorton: Any reason not to just make the wait time here 0 @oxtoacart? Seems like waiting at all just introduces an unnecessary delay.

oxtoacart: 25 ms is a timeout, not a wait time.  If we set it to 0, we would likely fail to get anything back within that time.

oxtoacart: Oh, I see what you mean though.  In this particular case, if we don't get anything immediately, there's nothing to get.  Yes, I'll change it :)

oxtoacart: Fixed.

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -71,19 +89,20 @@ func (client *Client) ListenAndServe(onListeningFn func()) error {
...
-// returns the highest QOS fronted.Dialer available, or nil if none available.
-func (client *Client) Configure(cfg *ClientConfig) {
+// before or after ListenAndServe, and can be called multiple times.
+func (client *Client) Configure(cfg *ClientConfig, proxyAll func() bool) {
 	client.cfgMutex.Lock()
 	defer client.cfgMutex.Unlock()
 
 	log.Debug("Configure() called")
+	if client.bal == nil {
+		client.bal = eventual.NewValue()
```
myleshorton: It seem like it would be more straightforward to keep all the balancer state local to balancer.go?

oxtoacart: I actually like that balancer doesn't have global state of its own - I kinda wish all our modules were like that :)

myleshorton: What about just moving these lines to balancer but keeping client.bal where it is @oxtoacart? The state is the same in both cases, right, just a matter of what file it's in? I think the two separate places where some sort of balancer init happens feels odd to me more than where the variable itself lives, but not a huge deal.

oxtoacart: The point is that as written, there can be an arbitrary number of balancers used by whomever, right?  To put this initialization logic into the balancer package, we'd have to enshrine the fact that there's only a single balancer at a time as part of the balancer package, yes?

oxtoacart: Oh, I'm a dufus!  You're talking about the balancer.go file in the client package.  Yes, I'm cool with that move :)

myleshorton: Nice yeah just seems a little clearer to me. I also wouldn't be at all opposed to just moving all the code from client/balancer.go into client/client.go, which to me might the least confusing of all, but whatever you think.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -46,21 +47,16 @@ var (
 	m                   *yamlconf.Manager
 	lastCloudConfigETag = map[string]string{}
 	r                   = regexp.MustCompile("\\d+\\.\\d+")
-	// Request the config via either chained servers or direct fronted servers.
-	cf = util.NewChainedAndFronted()
```
myleshorton: Oh here too re-creating chained and fronted each time will lose its internal state that will only use direct chained proxies if they're working.

oxtoacart: Fixed.

src/github.com/getlantern/flashlight/client/client.go
oxtoacart: As noted in the commit, I wasn't able to do this.  In fact, I had to move the initialization of the variable even earlier since it comes into use as soon as a user agent connects to the client proxy, which can happen before Configure gets called.

oxtoacart: Note - this is intentional.  The idea is to bring up the listener ASAP so that an embedding application doesn't block on our whole initialization sequence.  Of course any network access will still have to wait for the initialization sequence to finish.

oxtoacart: Another note - the desktop app doesn't actually turn on the PAC file until after Configure() has been called, so in practice the above comment really only applies to Manoto.

myleshorton:
Nice catch on the closing sequence @oxtoacart! This is all looking good to me.

myleshorton:
OK LGTM @oxtoacart! Merging!

oxtoacart:
Thanks for the review @myleshorton!

PR #3464, opened 2016-01-29
Issue 3460

xiam:


myleshorton:
Cool - yeah same thing -- we should be merging to the release-2.0.1 branch.

myleshorton:
I'm going to close these since we don't want to go straight into valencia (and since we need them for the release).

PR #3468, opened 2016-01-29
Stop Lantern when user disconnects from WiFi network

atavism:
This registers a new broadcast receiver that detects changes in network connectivity. If Lantern is running, and we disconnect from Wi-Fi, we automatically turn off the VPN service.


oxtoacart:
In my testing, the VPN service does indeed stop when I disable the network.

oxtoacart:
If I disable wifi while I'm in the app, the slider doesn't change to show that Lantern is actually off.

oxtoacart:
With the exception of that one small usability niggle, this works for me.  Merging.

myleshorton:
Nice! So does this mean we can push out a new mobile build @oxtoacart and @atavism?

oxtoacart:
@myleshorton Yes!

PR #3480, opened 2016-02-03
Reusing ReverseProxy

oxtoacart:
This requires a PR for servers to get deployed first so that they can handle multiple auth tokens - https://github.com/getlantern/http-proxy-lantern/pull/28

### DIFF
--- a/src/github.com/getlantern/flashlight/client/reverseproxy.go
+++ b/src/github.com/getlantern/flashlight/client/reverseproxy.go
@@ -3,75 +3,23 @@ package client
 import (
 	"bytes"
 	"io/ioutil"
-	"net"
 	"net/http"
 	"net/http/httputil"
 	"time"
 
-	"github.com/getlantern/balancer"
 	"github.com/getlantern/detour"
 	"github.com/getlantern/flashlight/proxy"
 	"github.com/getlantern/flashlight/status"
 )
 
-// authTransport allows us to override request headers for authentication and for
-// stripping X-Forwarded-For
-type authTransport struct {
-	http.Transport
-	balancedDialer *balancer.Dialer
-	deviceID       string
-}
+// newReverseProxy creates a reverse proxy that uses the client's balancer to
+// dial out.
+func (client *Client) newReverseProxy() *httputil.ReverseProxy {
+	bal := client.getBalancer()
 
-// We need to set the authentication token for the server we're connecting to,
-// and we also need to strip out X-Forwarded-For that reverseproxy adds because
-// it confuses the upstream servers with the additional 127.0.0.1 field when
-// upstream servers are trying to determin the client IP.
-// We need to add also the X-Lantern-Device-Id field.
-func (at *authTransport) RoundTrip(req *http.Request) (resp *http.Response, err error) {
-	norm := new(http.Request)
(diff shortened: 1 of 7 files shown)


myleshorton:
Looking good to me! Can you also update the Dockerfile to use 1.6 or is that not strictly necessary?

oxtoacart:
@myleshorton Oh yeah, I forgot about the docker build.  Let me fix that.

oxtoacart:
@myleshorton I've upgraded the go version in the Dockerfile to 1.6rc1.  It's able to build Windows and Linux binaries.

src/github.com/getlantern/flashlight/client/reverseproxy.go
```diff
@@ -3,90 +3,49 @@ package client
...
 	}
 
-	rp := &httputil.ReverseProxy{
+	allAuthTokens := bal.AllAuthTokens()
+	return &httputil.ReverseProxy{
+		// We need to set the authentication tokens for all servers that we might
+		// connect to because we don't know which one the dialer will actually
+		// pick. We also need to strip out X-Forwarded-For that reverseproxy adds
+		// because it confuses the upstream servers with the additional 127.0.0.1
+		// field when upstream servers are trying to determin the client IP.
```
myleshorton: nit: typo determin -> determine

myleshorton:
This otherwise looks great to me @oxtoacart! Great stuff.

src/github.com/getlantern/flashlight/client/reverseproxy.go
oxtoacart: Fixed.

myleshorton:
Cool not sure what's up with Travis, but tests are passing for me locally with the exception of that detour test that's fails inconsistently but is unrelated. Merging!

oxtoacart:
Thanks @myleshorton !

myleshorton:
Thank you @oxtoacart! Great change.

fffw:
@oxtoacart I saw strange behaviour with latest valencia code. With only one chained server in my config, I frequently get mimicked page, especially if the site is inaccessible. Known sites such as Twitter and FB never show that page. From the Lantern server, I can see lots of `No token provided from x.x.x.x:41601` message.

Is it expected with current http-proxy-lantern? It looks like in some cases, Lantern client never set the auth token in header.

oxtoacart:
@fffw Can you send me your lantern config?

myleshorton:
Is the multiple token checking code already deployed on the server?

fffw:
Can't reproduce it now. Will keep experimenting.

oxtoacart:
@myleshorton No, the multiple token checking code is not yet deployed.

oxtoacart:
I wonder if it could be a timing in the initialization sequence ...

oxtoacart:
@fffw If you run into the problem again, I wonder if this branch improves things for you:

https://github.com/getlantern/lantern/tree/initfix

fffw:
Thanks @oxtoacart I'll try

oxtoacart:
@fffw Something that may help you reproduce the problem is to remove the config files and start lantern again - the problem may only happen during the period when your first configuration is getting fetched.

PR #3485, opened 2016-02-03
Valencia mobile

oxtoacart:
These changes include what's necessary to build a library for use by the Manoto TV app.  The next phase is to integrate the library into the Lantern Mobile app, which will be a separate PR.


oxtoacart:
cc: @myleshorton

myleshorton:
Do you want the LanternMobileTestbed as a more permanent part of the code base @oxtoacart? If so maybe just a quick README describing what it is and how to build it would be worthwhile?

Makefile
```diff
@@ -25,7 +25,7 @@ BUILD_DATE := $(shell date -u +%Y%m%d.%H%M%S)
 
 LOGGLY_TOKEN := 2b68163b-89b6-4196-b878-c1aca4bbdf84
 
-LDFLAGS := -w -X=main.version=$(GIT_REVISION) -X=main.revisionDate=$(REVISION_DATE) -X=main.buildDate=$(BUILD_DATE) -X=github.com/getlantern/flashlight/logging.logglyToken=$(LOGGLY_TOKEN)
+LDFLAGS := -X github.com/getlantern/flashlight/flashlight.Version=$(GIT_REVISION) -X github.com/getlantern/flashlight/flashlight.RevisionDate=$(REVISION_DATE) -X github.com/getlantern/flashlight/flashlight.BuildDate=$(BUILD_DATE) -X github.com/getlantern/flashlight/logging.logglyToken=$(LOGGLY_TOKEN) -X github.com/getlantern/flashlight/logging.logglyTag=$(LOGGLY_TAG)
```
myleshorton: Nice way to do it!

Makefile
```diff
@@ -498,36 +504,15 @@ genconfig:
...
-		echo "cp -v $(LANTERN_MOBILE_DIR)/$(LANTERN_MOBILE_LIBRARY) \$$FIRETWEET_MAIN_DIR"; \
-	fi
+	gomobile bind -target=android -tags='headless' -o=$(LANTERN_MOBILE_LIBRARY) -ldflags='$(LDFLAGS) $$EXTRA_LDFLAGS' $(LANTERN_MOBILE_PKG) && \
+	mkdir -p $(MOBILE_TESTBED_LIBS) && cp -f $(LANTERN_MOBILE_LIBRARY) $(MOBILE_TESTBED_LIBS) && \
+	mkdir -p $(MOBILE_SDK_LIBS) && cp -f $(LANTERN_MOBILE_LIBRARY) $(MOBILE_SDK_LIBS)
 
-android-lib-dist: genconfig android-lib
+android-sdk: android-lib
+	(cd $(MOBILE_SDK) && gradle assembleDebug) && \
+	cp $(MOBILE_SDK)/sdk/build/outputs/aar/sdk-debug.aar $(MOBILE_TESTBED_LIBS)
```
myleshorton: Is there another release build in addition to building with the debug SDK? Or do we always want debug for some reason @oxtoacart?

Makefile
oxtoacart: Yeah, the README in the final pull request from valencia-mobile-full addresses this sort of thing.  That Makefile is also cleaned up, so look there for that kind of polish stuff.

Makefile
oxtoacart: Great point.  Let me loop back with @atavism on Monday and make sure that we have a proper release build in the final Makefile.

src/github.com/getlantern/lantern/lantern.go
```diff
@@ -0,0 +1,68 @@
...
+//
+// If a Lantern proxy is already running within this process, that proxy is
+// reused.
+//
+// Note - this does not wait for the entire initialization sequence to finish,
+// just for the proxy to be listening. Once the proxy is listening, one can
+// start to use it, even as it finishes its initialization sequence. However,
+// initial activity may be slow, so clients with low read timeouts may
+// time out.
+func Start(configDir string, timeoutMillis int) (string, error) {
```
myleshorton: @oxtoacart what calls Start in desktop Lantern? Or is that part in another PR?

oxtoacart: Desktop lantern just calls `flashlight.Run' directly from main.go.

LanternMobileTestbed/app/src/main/java/org/lantern/lanternmobiletestbed/Browse.java
```diff
@@ -0,0 +1,141 @@
...
+                        Scanner s = new Scanner(in).useDelimiter("\\A");
+                        return s.hasNext() ? s.next() : "";
+                    } finally {
+                        urlConnection.disconnect();
+                        Log.i(TAG, "Finished doing geolookup");
+                    }
+                } catch (Exception e) {
+                    return "Unable to refresh IP: " + e.getMessage();
+                }
+            }
```
atavism: This refreshIP test would be really useful to add to the connectivity checker, as a quick test to verify the status of the VPN connection. What do you think of factoring this out (and reusing it here of course) for that purpose? Maybe we could add it to ``Utils`` that comes with the SDK?

oxtoacart: How is the connectivity checker used?

atavism: It's just a part of the main activity, but changes in network connectivity are monitored using a [BroadcastReceiver](http://developer.android.com/reference/android/content/BroadcastReceiver.html) using several intent filters. When WiFi is switched off, for instance, we automatically receive a ``ConnectivityManager.CONNECTIVITY_ACTION`` broadcast, and know to switch Lantern off.

We've seen plenty of instances where a device goes to sleep, with Lantern left running, only for the VPN connection to be unresponsive on return for whatever reason, necessitating a similar refresh. That's handled with ``ACTION_USER_PRESENT`` broadcast. Anyway, maybe this refresh test would be a useful addition.

oxtoacart: I've opened issue #3524 for this.  If it's okay with you, let's just leave it as is for the purpose of this PR.

atavism:
Previously, we were displaying the version number at the bottom of the side-menu and passing it to the ``afterStart`` method. It's set using ``UI.setVersionNum(str)``. Maybe we could include a version field with the new ``StartResult`` type? Then add a ``getVersion()`` method to ``org.lantern.mobilesdk``?

oxtoacart:
@atavism What version number do we want to display?  The version of the library or the version of Lantern Mobile?

atavism:
@oxtoacart I think we want to display both. Maybe we could combine them? Presumably the library version will be the same as Lantern, and we could use the build number to represent the Lantern Mobile version? i.e., 2.0.11.x -- or what do you think is the best approach here?

PR #3486, opened 2016-02-03
Lantern mobile release 2-3

atavism:


myleshorton:
Nice @atavism! This has one last QA cycle to go before being pushed out, is that right @sunshine19090?

sunshine19090:
Todd sent me with the fix for blank space and I did some tests with that
too. So if this is same, then I don't have to check again. It is good to go!

On Wed, Feb 3, 2016 at 2:43 PM, myleshorton <notifications@github.com>
wrote:

> Nice @atavism <https://github.com/atavism>! This has one last QA cycle to
> go before being pushed out, is that right @sunshine19090
> <https://github.com/sunshine19090>?
>
> —
> <https://github.com/getlantern/lantern/pull/3486#issuecomment-179508450>.
>



-- 
Regards,
Sunshine

PR #3496, opened 2016-02-05
Add SOCKS capability to Lantern

oxtoacart:


MobileSDK/sdk/src/main/java/org/lantern/mobilesdk/Lantern.java
```diff
@@ -0,0 +1,127 @@
...
+        if (analyticsTrackingId != null && !enabled) {
+            trackStartSession(context, analyticsTrackingId);
+        }
+        return result;
+    }
+
+    private static go.lantern.Lantern.StartResult enable(String configDir, int timeoutMillis) {
+        try {
+            go.lantern.Lantern.StartResult result = go.lantern.Lantern.Start(configDir, timeoutMillis);
+            String addr = result.getHTTPAddr();
```
atavism: Does ``getHTTPAddr()`` always return a string that can be split on ':' now, assuming we get a valid result?

MobileSDK/sdk/src/main/java/org/lantern/mobilesdk/Lantern.java
```diff
@@ -0,0 +1,127 @@
...
+            String host = addr.split(":")[0];
+            String port = addr.split(":")[1];
+            System.setProperty("http.proxyHost", host);
+            System.setProperty("http.proxyPort", port);
+            System.setProperty("https.proxyHost", host);
+            System.setProperty("https.proxyPort", port);
+            return result;
+        } catch (Exception e) {
+            throw new RuntimeException("Unable to start Lantern: " + e.getMessage(), e);
```
atavism: I wonder if throwing a ``RuntimeException`` here is too severe. It means Lantern could not be initialized, and there is nothing the client can do to recover. But applications that integrate the SDK may want the option to continue running regardless.

MobileSDK/sdk/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+        versionCode 1
+        versionName "1.0"
+    }
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+}
```
atavism: Are we going to add back all the publishing options/commands for uploading artifacts to Maven Central in a separate PR, or are we getting rid of all that for some reason? The whole point of a separate SDK project was to turn Lantern into an Android library our own apps ("Lantern Android", FireTweet) and third party apps could integrate easily http://lantern.readthedocs.org/en/latest/ Just clarifying.

atavism:
We probably want to remove all the ``.class`` files from this PR and add a .gitignore *.class entry? Actually, we shouldn't include the ``app/build`` directory at all

MobileSDK/sdk/src/main/java/org/lantern/mobilesdk/Lantern.java
oxtoacart: For IPv4 it does, but for IPv6 we actually need to use the portion after the last colon as the port.  I'll fix that.

MobileSDK/sdk/src/main/java/org/lantern/mobilesdk/Lantern.java
oxtoacart: They can do that by catching the exception.  I'll make it a checked exception to be clear.

MobileSDK/sdk/build.gradle
oxtoacart: Still need to add that back in.  I could use your help with that, as I don't understand it very well.

oxtoacart:
@atavism Sorry about those build outputs, that's fixed.  I've addressed your other two comments as well.  I also noticed that I committed some .gradle directories, that's fixed in the final PR (valencia-mobile-full).

MobileSDK/sdk/build.gradle
atavism: Oh sure =P Let's just do that in a separate PR as this one is already massive!

PR #3497, opened 2016-02-05
Integrate embeddable Lantern with Lantern Mobile

oxtoacart:
This includes a slightly altered version of `protected`.  Unlike the previous version, `Dial` doesn't do its own DNS lookups anymore, rather the package provides an API for `tlsdialer` to use in doing its own DNS lookup.  That avoids doing duplicate DNS lookups.  Also, rather than making an explicit dependency from `tlsdialer` to `protected`, `tlsdialer` provides a generic API to allow overriding DNS lookup and dialing behavior, which is then used from within the Lantern Mobile app to register `protected`.


LanternMobileTestbed/settings.gradle
```diff
@@ -0,0 +1 @@
+include ':app', ':libflashlight'
```
atavism: This is called ``liblantern`` now, right?

Makefile
```diff
@@ -499,39 +514,70 @@ genconfig:
...
-		echo ""; \
-		echo "Either no FIRETWEET_MAIN_DIR variable was passed or the given value is not a";\
-		echo "directory. You'll have to copy the $(LANTERN_MOBILE_LIBRARY) file manually:"; \
-		echo ""; \
-		echo "cp -v $(LANTERN_MOBILE_DIR)/$(LANTERN_MOBILE_LIBRARY) \$$FIRETWEET_MAIN_DIR"; \
-	fi
+$(ANDROID_LIB):
+	source setenv.bash && \
+	go install golang.org/x/mobile/cmd/gomobile && \
+	gomobile init && \
```
atavism: Not sure if you've encountered the same issue, but ``gomobile init`` seems to take an inordinate amount of time to complete. At least running off the latest version, It appears to download and re-install the existing toolchain, regardless of the ``-u`` flag, every time. Anyway, it's bearable, just time-consuming.

atavism:
@oxtoacart Still testing, but your changes really look fantastic! Totally love how you've restructured and simplified things while preserving key aspects. I also think the new mobile test bed app will be useful for automating testing (either through AWS Device Farm and/or Travis? see https://docs.travis-ci.com/user/languages/android) as a next step potentially. A huge improvement, and I'm excited to get it over to @sunshine19090! I've tested over the weekend, and the new version has been working great for me.

LanternMobileTestbed/settings.gradle
oxtoacart: Thanks!  I'd been wondering why Intellij kept showing an empty libflashlight project.  I've removed this.

Makefile
oxtoacart: You're right, it does take a while.  I've updated the makefile so that it only installs and initializes gomobile when necessary.

oxtoacart:
@atavism Thanks for reviewing!

atavism:
I'm just going to merge this! It's working great and has been for me; most of the issues we've discussed we can just address in separate, smaller PRs. Thanks, @oxtoacart!

Issue #3358, opened 2015-12-17
When we have multiple servers available, use the fastest one


myleshorton:
It's not 100% clear what the best approach is here, but when Lantern is assigned multiple servers for whatever reason, it would be good to have logic that always chooses the fastest server. One approach would be to have some process that periodically checks all servers and ranks them, possibly only as long as Lantern is active. 

The idea would be to bypass network congestion, overloaded servers, and servers with high latency.

myleshorton:
@fffw can you take a look at this? I think we need it for pro.

fffw:
I found it related to https://github.com/getlantern/lantern/issues/3127

My plan is to rank by real traffic data, or even maintain a routing table for frequently visited sites (overskill maybe). Balancer is where I want to make that change.

PR #3507, opened 2016-02-06
Balancer pick dialers per different strategy resolves #3127

### LINKED ISSUE #3127: More robust and adaptive balancer
Balancer is the key component to speed and stability of Lantern client. It should work systematically regardless of 1 or N dialers available, and be adaptive to dynamic network condition, especially when we bring back P2P.

A few points:
1. Use real measurement to weight different dialers, including error rate, delay and throughput, rather than predefined QoS.
2. Differentiate errors returned by proxy server from errors between client and proxy server. Only take latter into account when calculating weight.
3. Maintain only a few persistent connections to proxy server. Currently we predial in `connpool` and close connection immediately, which is a waste of resource in both client and proxy server.

fffw:
Weighted strategy is not implemented correctly yet. More detailed experiment is required to measure of actual performance. But the structure is kinda complete.

@oxtoacart want to have a look?

### DIFF
--- a/src/github.com/getlantern/balancer/README.md
+++ b/src/github.com/getlantern/balancer/README.md
@@ -12,24 +12,16 @@ For docs:
 
 ===Benchmark
 
-To evaluate performance of different strategy to pick dialer, run below
-command. The 20s `benchtime` is to simulate network latency in real world.
-`Sticky` and `QualityFirst` strategy seems has better result at this moment.
-
-```
-go test -bench . -benchtime 20s
-```
+`go test -bench .` to evaluate performance of different strategy to pick
+dialer. `Sticky` and `QualityFirst` strategy seems has better result at this
+moment.
 
 Example output:
 ```
-BenchmarkSticky-4           1000      23855693 ns/op
---- BENCH: BenchmarkSticky-4
-    benchmark_test.go:89: 'Fail 1% 10ms ± 0 dialer': 11
-    benchmark_test.go:89: 'Fail 10% 10ms ± 0 dialer': 12
-    benchmark_test.go:89: 'Fail 50% 10ms ± 0 dialer': 10
-    benchmark_test.go:92: - Failed dial attempts: 33 out of 1000
-    benchmark_test.go:89: 'Fail 1% 10ms ± 8ms dialer': 12
-    benchmark_test.go:89: 'Fail 10% 10ms ± 8ms dialer': 12
-    benchmark_test.go:89: 'Fail 50% 10ms ± 8ms dialer': 8
-    benchmark_test.go:92: - Failed dial attempts: 32 out of 1000
+BenchmarkQualityFirst-4   100000         23222 ns/op
+--- BENCH: BenchmarkQualityFirst-4
+    benchmark_test.go:98: '1%': 83/16850, '10%': 199/3596, '99%': 228/524,  fail/total = 510/20970 (2.4%) in 10000 runs
+    benchmark_test.go:98: '1% 10ns±8ns': 85/18319, '10% 10ns±8ns': 86/1982, '50% 10ns±8ns': 85/457,  fail/total = 256/20758 (1.2%) in 10000 runs
+    benchmark_test.go:98: '1%': 843/169828, '10%': 1784/35104, '99%': 3074/6879,  fail/total = 5701/211811 (2.7%) in 100000 runs
+    benchmark_test.go:98: '1% 10ns±8ns': 955/184247, '10% 10ns±8ns': 954/19488, '50% 10ns±8ns': 954/4734,  fail/total = 2863/208469 (1.4%) in 100000 runs
 ```
--- a/src/github.com/getlantern/balancer/README.md
+++ b/src/github.com/getlantern/balancer/README.md
@@ -9,3 +9,27 @@ To install:
 For docs:
 
 `godoc github.com/getlantern/balancer`
+
+===Benchmark
+
+To evaluate performance of different strategy to pick dialer, run below
+command. The 20s `benchtime` is to simulate network latency in real world.
+`Sticky` and `QualityFirst` strategy seems has better result at this moment.
+
+```
+go test -bench . -benchtime 20s
+```
+
+Example output:
+```
+BenchmarkSticky-4           1000      23855693 ns/op
+--- BENCH: BenchmarkSticky-4
+    benchmark_test.go:89: 'Fail 1% 10ms ± 0 dialer': 11
+    benchmark_test.go:89: 'Fail 10% 10ms ± 0 dialer': 12
+    benchmark_test.go:89: 'Fail 50% 10ms ± 0 dialer': 10
+    benchmark_test.go:92: - Failed dial attempts: 33 out of 1000
+    benchmark_test.go:89: 'Fail 1% 10ms ± 8ms dialer': 12
+    benchmark_test.go:89: 'Fail 10% 10ms ± 8ms dialer': 12
+    benchmark_test.go:89: 'Fail 50% 10ms ± 8ms dialer': 8
+    benchmark_test.go:92: - Failed dial attempts: 32 out of 1000
+```
--- a/src/github.com/getlantern/balancer/balancer.go
+++ b/src/github.com/getlantern/balancer/balancer.go
@@ -22,7 +22,7 @@ type Balancer struct {
 }
 
 // New creates a new Balancer using the supplied Dialers.
-func New(cr HeapCreater, dialers ...*Dialer) *Balancer {
+func New(st Strategy, dialers ...*Dialer) *Balancer {
 	trustedDialersCount := 0
 
 	var dls []*dialer
@@ -39,7 +39,7 @@ func New(cr HeapCreater, dialers ...*Dialer) *Balancer {
 		}
 	}
 
-	bal := &Balancer{cr(dls), cr(tdls)}
+	bal := &Balancer{st(dls), st(tdls)}
 	heap.Init(&bal.dialers)
 	heap.Init(&bal.trusted)
 	return bal
--- a/src/github.com/getlantern/balancer/balancer.go
+++ b/src/github.com/getlantern/balancer/balancer.go
@@ -3,10 +3,10 @@
 package balancer
 
 import (
+	"container/heap"
 	"fmt"
 	"math/rand"
 	"net"
-	"sort"
 
 	"github.com/getlantern/golog"
 )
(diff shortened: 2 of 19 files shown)


src/github.com/getlantern/balancer/balancer.go
```diff
@@ -15,69 +15,49 @@ var (
...
+	var result []string
+	for i := 0; i < b.dialers.Len(); i++ {
+		result = append(result, b.dialers.dialers[i].AuthToken)
 	}
 	return result
 }
 
-// TrustedDialerAndConn creaetes a balancer.Dialer and a network connection for an HTTP connection
-// (as opposed to HTTPS).
-func (b *Balancer) TrustedDialerAndConn() (*Dialer, net.Conn, error) {
```
fffw: Not used anymore

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -95,20 +75,19 @@ func (b *Balancer) dialerAndConn(network, addr string, targetQOS int) (*Dialer,
 	// To prevent dialing infinitely
 	attempts := 3
 	for i := 0; i < attempts; i++ {
-		if len(dialers) == 0 {
+		if dialers.Len() == 0 {
 			return nil, nil, fmt.Errorf("No dialers left to try on pass %v", i)
 		}
 		var d *dialer
-		d, dialers = randomDialer(dialers, targetQOS)
+		d = heap.Pop(&dialers).(*dialer)
```
fffw: Replace weight based `randomDialer` with strategy based dialer heap.

src/github.com/getlantern/balancer/benchmark_test.go
```diff
@@ -0,0 +1,94 @@
...
+	Skip:
+		// Use goto because defer has significant performance impact
+		_ = c.Close()
+	}
+	if b.N < 1000 {
+		return
+	}
+	var total int
+	for _, d := range bal.dialers.dialers {
+		re, err := regexp.Compile(d.Label)
```
fffw: We can track the fail count of each dialer instead of regexping the error output. As it's benchmarking code, it's probably ok to keep it as is.

src/github.com/getlantern/balancer/dialer.go
```diff
@@ -48,45 +48,49 @@ type Dialer struct {
...
-				log.Tracef("Mark dialer %s as inactive, scheduling check", d.Label)
-				timeout := time.Duration(consecCheckFailures*consecCheckFailures) * 100 * time.Millisecond
-				if timeout > maxCheckTimeout {
-					timeout = maxCheckTimeout
-				}
-				timer.Reset(timeout)
-			} else {
-				atomic.StoreInt32(&d.active, 1)
-				log.Tracef("Mark dialer %s as active", d.Label)
+			cf := atomic.LoadInt32(&d.consecFailures)
```
fffw: Removed all timestamp checking and rely on counters for dialer checking logic.

src/github.com/getlantern/balancer/dialer.go
```diff
@@ -48,45 +48,49 @@ type Dialer struct {
...
-	closeCh chan interface{}
-	errCh   chan time.Time
+	consecSuccesses int32
+	consecFailures  int32
+	// It's actually the average of last connect time and previous average. so
+	// if the connect time for i iteration is t[i], after n iteration, its value
+	// will be 1/2(t[n] + 1/2(t[n-1] + 1/2(t[n-2) + ... + t[1]))...), most
+	// recent connect time contributes most to the value, seems a good indicator.
+	avgConnTime int64
+	closeCh     chan struct{}
```
fffw: I changed it to chan of [empty struct](http://dave.cheney.net/2014/03/25/the-empty-struct) mostly because `chan interface{}` accepts any type while `chan struct{}` can only accept `chan struct{}{}`.

oxtoacart:
I'll gladly take a look.  Do we have a specification or requirements for
what this is supposed to accomplish?



-- 
Cheers,
Ox

- Marge Piercy

fffw:
Sorry @oxtoacart I forgot it's weekend. This PR should have lower priority than other outstanding ones.

The basic idea is at https://github.com/getlantern/lantern/issues/3358, to choose the most reliable and fastest servers and possibly distributing traffic among them. It's not clear which way is the best at this moment, so this PR is to provide a simple way to implement different strategies, benchmark them and adjust parameters.

oxtoacart:
Cool, thanks!
>
>

-- 
Cheers,
Ox

- Marge Piercy

src/github.com/getlantern/balancer/README.md
```diff
@@ -9,3 +9,19 @@ To install:
 For docs:
 
 `godoc github.com/getlantern/balancer`
+
+===Benchmark
+
+`go test -bench .` to evaluate performance of different strategy to pick
```
oxtoacart: Very cool!

src/github.com/getlantern/balancer/balancer.go
oxtoacart: Good catch.

src/github.com/getlantern/balancer/balancer.go
oxtoacart: What happens if the heap is empty?  Will Pop() panic?  Or will it return nil?  If it returns nil, I think the .(*dialer) type assertion will fail.

src/github.com/getlantern/balancer/dialer.go
```diff
@@ -94,29 +98,52 @@ func (d *dialer) start() {
...
-					consecCheckFailures += 1
+					d.errCh <- struct{}{}
 				}
 			}
 		}
 	}()
 }
 
+func (d *dialer) metrics() metrics {
+	return metrics{atomic.LoadInt64(&d.avgConnTime), atomic.LoadInt32(&d.consecSuccesses), atomic.LoadInt32(&d.consecFailures)}
```
oxtoacart: You could actually get rid of the avgConnTime, consecSuccesses and consecFailures variables and have just a single metrics instance, and that way you don't have to construct a new one every time metrics() is called.

src/github.com/getlantern/balancer/strategy.go
```diff
@@ -0,0 +1,106 @@
+package balancer
```
oxtoacart: I like how you used a heap+sort order to structure this.

src/github.com/getlantern/balancer/balancer.go
oxtoacart: Also, what if there are multiple dialers with effectively the same priority (or similar priorities)?

src/github.com/getlantern/balancer/strategy.go
```diff
@@ -0,0 +1,106 @@
...
+	}}
+}
+
+// Sticky strategy always pick the dialer with largest consecutive success
+// count or the smallest consecutive failure count
+func Sticky(dialers []*dialer) dialerHeap {
+	return dialerHeap{dialers, func(i, j int) bool {
+		mi := dialers[i].metrics()
+		mj := dialers[j].metrics()
+		return (mi.consecSuccesses - mi.consecFailures) >
```
oxtoacart: Based on the logic in dialer.go, either consecSuccesses or consecFailures will be 0.  So when doing this comparison, you're really just saying:

1. If only one of the dialers succeeded on its last attempt, use that otherwise
2. Use the dialer that's had the most consecutive successes.

Is that what you intended?  If so, it might be clearer to express the logic that way, e.g.:

```golang
return mi.consecSuccesses > mj.consecSuccesses
```

src/github.com/getlantern/balancer/strategy.go
```diff
@@ -0,0 +1,106 @@
...
+			(mj.consecSuccesses - mj.consecFailures)
+	}}
+}
+
+// Fastest strategy always pick the dialer with lowest recent average connect time
+func Fastest(dialers []*dialer) dialerHeap {
+	return dialerHeap{dialers, func(i, j int) bool {
+		mi := dialers[i].metrics()
+		mj := dialers[j].metrics()
+		return mi.avgConnTime < mj.avgConnTime
```
oxtoacart: I don't know that connect time is a great measure of the proxy's performance.  It's a decent measure of latency to first byte, which is important for really small requests, but less important for larger requests.  Is there a way that we can measure throughput as well?  I know that throughput is sensitive to the performance of the destination sites, so that's tricky, but if we can get it right it would be a useful measure.

src/github.com/getlantern/balancer/balancer.go
oxtoacart: Specifically, my concern here is that once we find a "best" proxy, as long as that doesn't fail, we'll never try any of the other proxies, so we don't have an opportunity to find out if any of them has gotten better than the one that we thought was "best".

fffw: We have a check in L78 to avoid `Popping` an empty heap.

It's the responsibility of different "strategies" to handle multiple good proxies. `Sticky` will keep using current proxy given it doesn't fail, but other strategies may introduce randomness or consider other metrics.

src/github.com/getlantern/balancer/dialer.go
fffw: Sounds a good idea.

src/github.com/getlantern/balancer/strategy.go
fffw: One special case is when local network just recovers, both have positive `consecFailures` and zero `consecSuccesses`, the one with smaller `consecFailures` is more likely reachable before network was down.

oxtoacart: Got it, thanks!

src/github.com/getlantern/balancer/balancer.go
oxtoacart: Gotcha, thanks.

oxtoacart:
@fffw This looks pretty good to me, except for the hanging thing.  Unfortunately I haven't had time to dig into the code to figure out why it hangs sometimes.  Have you ever observed this?

fffw:
@oxtoacart I guess you mean the PR of new detour instead right?

oxtoacart:
@fffw Oops, yeah, I posted on the wrong PR.

src/github.com/getlantern/balancer/strategy.go
fffw: Thinking twice about throughput, there are 3 main physical factors: bandwidth, latency and packet drop rate. In normal condition, TCP flow control will gradually reach the limit of lowest bandwidth link, which is usually user's local network.

Apart from latency, the main factor to differentiate throughput among proxies is network congestion at international trunk or the network proxies reside, which is roughly random to TCP endpoints, so TCP flow control can't handle it gracefully, hence results more dropped packets. Packet dropping has two consequences:

1. Increase the time required to `Read()` or connect, because of TCP retransmission.

2. Fail to connect or close prematurely, if the congestion is serious enough.

Consider the randomness of network congestion, the packet drop rate should be the same at each stage of a TCP connection, so statistically, the delay and fail rate to establish connections can represent the overall delay and fail rate, unless there's routing policy somewhere to drop more larger packets than smaller ones.

That is, we don't need to measure throughput separately. Or, if we confirmed that drop rate for large packets is higher, just periodically check with large enough payload for single packet.

Is it right @oxtoacart ?

Inspired by:
1. https://blog.thousandeyes.com/a-very-simple-model-for-tcp-throughput/
2. http://bradhedlund.com/2008/12/19/how-to-calculate-tcp-throughput-for-long-distance-links/

oxtoacart: @fffw Aside from the network, we need to consider the behavior of the server itself.  If a server starts being overloaded, does that have a similar effect on latency and throughput or is one affected more than the other?

We had a recent email thread entitled "Proxying super slow at peak times".  I did some measurements from the affected server by downloading a small, medium and large file.  As server performance degraded, the effect was most noticeable on the large file, suggesting that bandwidth may be more affected than latency.

I think the best way to understand this is to run some tests.  To simulate a server being overloaded CPU-wise, we could just run a program on the server that consumes a lot of CPU without doing any real work.

fffw: Good point @oxtoacart will do more tests, maybe go further to simulate different pattern using [Comcast](https://github.com/tylertreat/Comcast).

oxtoacart: Comcast looks like a good idea!

oxtoacart: FWIW, we now have the ping capability in http-proxy-lantern, so if you wanted to integrate some of that data into your selection strategy, we can collect data from the field.  If we can compare the selections based on latency with the selections based on throughput, we can see whether or not they're similar.  If they are, we can turn off the throughput-based measurement for efficiency.

oxtoacart: See [pinger.go](https://github.com/getlantern/http-proxy-lantern/blob/master/pinger/pinger.go) for code that pings the proxy server.

oxtoacart:
@fffw I ran some tests using pinger from my home network against Amsterdam, Frankfurt and Singapore.  As you suggested, TCP connect time was very closely related with transfer times for small, medium and large files, so I agree that's a good enough metric to start with.

fffw:
So I tested this `QualityFirst` strategy with real config on the China box. It will pick faster JP server at first and stick with that one. After throttled `eth0` to 10kbps, 1000ms latency or 8% packet loss rate respectively, it will dial both JP and AMS servers. It will pick JP server again soon after stopped throttling.

`comcast` has a bug to emulate network for specific address on Mac, so we need to find a better way to emulate network dynamically. Tests also need to be updated. But in general, `QualityFirst` is good enough to pick up the fastest server.

oxtoacart:
@fffw Nice!

fffw:
Close in favor of https://github.com/getlantern/lantern/pull/3755

Issue #3437, opened 2016-01-25
Run process that automatically updates the cloud masquerade configs daily


oxtoacart:
Basically, do what we do manually right now with `genconfig.bash`.  We need this because CloudFront certificates go bad relatively quickly, so we need to update the configs pretty frequently.

oxtoacart:
cc: @myleshorton

myleshorton:
@aranhoide can you add this one to your list?

aranhoide:
This is already working on the lantern_aws project.  I think this new setup calls for some changes in genconfig itself.  I've added them in the following PR:

https://github.com/getlantern/lantern/issues/3511

Feel free to close this issue as soon as that PR is merged or rejected.

oxtoacart:
#3511 has been merged.

PR #3511, opened 2016-02-07
move uploading of global configuration to the cloudmaster

aranhoide:
This is part of issue #3437 .  There's a [complementary PR] (https://github.com/getlantern/lantern_aws/pull/77) to the lantern_aws project.

These changes involve removing support for uploading the config from here, because any manual uploads would be overwritten by the daily automatic uploads from the cloudmaster.  It also makes the output golang-formatted files configurable so they can be skipped in the cloudmaster, which will use a genconfig binary.

I chose not to require a whole lantern checkout (and golang installation) on the cloudmaster because we would need to maintain the versions for both anyway.  This approach is explicit about what's in lantern_aws and how to update it.  It introduces some duplication but we can factor out the changing parts into some common source as we learn what those actually are in practice.


oxtoacart:
LGTM.

PR #3514, opened 2016-02-08
Issue 3513

xiam:
Should take care of the Ad not showing when updating a previous installation of Lantern.


myleshorton:
Why not just use `{{ model.settings.version }}` from the backend @xiam?

xiam:
@myleshorton I understand correctly that value is not available when the UI starts up, that one is sent by the backend in one of the async ajax calls; it should be something superfast, though, should we try that?

myleshorton:
Well it's tempting to not worry too much about it, especially with valencia using React. Let's just stick with what you've got!

Thanks @xiam.

PR #3530, opened 2016-02-09
Added chan of chans to fix race loading candidates closes #3528

### LINKED ISSUE #3528: Data race using DDF dialers
```
WARNING: DATA RACE
Write by goroutine 21:
  github.com/getlantern/fronted.Configure()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/fronted/direct.go:46 +0x41a
  github.com/getlantern/flashlight.applyClientConfig()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/flashlight.go:141 +0x3b8
  github.com/getlantern/flashlight.Run()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/flashlight.go:89 +0x459
  main.doMain.func2()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/main/main.go:181 +0x152

Previous read by goroutine 44:
  github.com/getlantern/fronted.(*direct).Dial()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/fronted/direct.go:141 +0x74
  github.com/getlantern/fronted.(*direct).Dial-fm()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/fronted/direct.go:112 +0x7b
  net/http.(*Transport).dial()
      /private/var/folders/bn/49l0ffts381_wqsh02663nr40000gn/T/workdir/go/src/net/http/transport.go:499 +0xd1
  net/http.(*Transport).dialConn()
      /private/var/folders/bn/49l0ffts381_wqsh02663nr40000gn/T/workdir/go/src/net/http/transport.go:596 +0x21f9
  net/http.(*Transport).getConn.func4()
      /private/var/folders/bn/49l0ffts381_wqsh02663nr40000gn/T/workdir/go/src/net/http/transport.go:549 +0x73

Goroutine 21 (running) created at:
  main.doMain()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/main/main.go:186 +0x195
  main._main()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/main/main.go:130 +0x34
  github.com/getlantern/systray.Run.func1()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/systray/systray.go:53 +0x5f

Goroutine 44 (running) created at:
  net/http.(*Transport).getConn()
      /private/var/folders/bn/49l0ffts381_wqsh02663nr40000gn/T/workdir/go/src/net/http/transport.go:551 +0x363
  net/http.(*Transport).RoundTrip()
      /private/var/folders/bn/49l0ffts381_wqsh02663nr40000gn/T/workdir/go/src/net/http/transport.go:228 +0x62a
  github.com/getlantern/fronted.(*directTransport).RoundTrip()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/fronted/direct.go:106 +0x1f6
  net/http.send()
      /private/var/folders/bn/49l0ffts381_wqsh02663nr40000gn/T/workdir/go/src/net/http/client.go:220 +0x73d
  net/http.(*Client).send()
      /private/var/folders/bn/49l0ffts381_wqsh02663nr40000gn/T/workdir/go/src/net/http/client.go:143 +0x1f7
  net/http.(*Client).doFollowingRedirects()
      /private/var/folders/bn/49l0ffts381_wqsh02663nr40000gn/T/workdir/go/src/net/http/client.go:380 +0x1059
  net/http.(*Client).Do()
      /private/var/folders/bn/49l0ffts381_wqsh02663nr40000gn/T/workdir/go/src/net/http/client.go:175 +0xc1
  github.com/getlantern/fronted.(*direct).Do()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/fronted/direct.go:125 +0x7e
  github.com/getlantern/flashlight/util.(*dualFetcher).Do.func1()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/util/http.go:97 +0x8f
  github.com/getlantern/flashligFeb 09 18:38:31.500 - 0m0s ht/util.(*dualFetcher).Do.func2()
      /Users/afisk/gopath/src/github.com/getlantern/lantern/src/github.com/getlantern/flashlight/util/http.go:122 +0x3b4
```

myleshorton:


myleshorton:
You want to take a look at this @oxtoacart? Found another issue while working on this but will open it in a separate ticket...

src/github.com/getlantern/fronted/direct.go
```diff
@@ -136,6 +137,8 @@ func (d *direct) Do(req *http.Request) (*http.Response, error) {
 // Dial persistently dials masquerades until one succeeds.
 func (d *direct) Dial(network, addr string) (net.Conn, error) {
 	gotFirst := false
+	candidateCh := <-candidateChCh
+	candidateChCh <- candidateCh
```
oxtoacart: It's unclear to me what happens if Configure() is called concurrently to this.  It might be safer and easier to read to use an [atomic.Value](https://golang.org/pkg/sync/atomic/#Value).

myleshorton: Ahh you're saying because the `candidateChCh <- candidateCh` call will block if `Configure` sneaks in there and adds it's own channel, is that right? I guess we could also just throw it on a go routine and/or make candidateChCh a little bigger so it would never really happen in practice.

TBC, what value are you suggesting we make an atomic.Value @oxtoacart?

oxtoacart: Make candidateCh an atomic.Value.  Dial would get the value each time, and Configure would set it.

oxtoacart: BTW, the problem with putting candidateCh back on the candidateChCh is that you'll end up with two versions of candidateCh on candidateChCh after Configure() is called a second time (because candidateChCh is sized to 1).  After a 3rd call to Configure(), either Configure() or Dial() will get deadlocked because the channel is full at that point.

You could do this in Dial to resolve that:

```golang
select {
    case candidateChCh <- candidateCh:
        // okay
    default:
        // discard
}
```

If you make candidateChCh a blocking (size 0) channel, that also addresses the multiple versions of candidateCh issue.  Still, I think using an atomic.Value is cleaner.

myleshorton: > BTW, the problem with putting candidateCh back on the candidateChCh is that you'll end up with two versions of candidateCh on candidateChCh after Configure() is called a second time (because candidateChCh is sized to 1). After a 3rd call to Configure(), either Configure() or Dial() will get deadlocked because the channel is full at that point.

Yeah that's what I was saying about that `candidateChCh <- candidateCh` line blocking.

The big problem with the atomic.Value, though, goes back to the original race issue, i.e. if `Dial` reads it before `Configure` is called, then `candidateCh` will be empty and never populated (it will be replaced with a new one when `Configure` is called, but `Dial` will be none the wiser).

I just committed your select change. I like that a little better than the straight blocking channel because AFAICT in the case of multiple Configure calls we'd be building up a bunch of chans in the queue that we will never need (since we only care about the latest from `Configure`) as well as building up a bunch of go routines with a goroutine call that executes `candidateChCh <- candidateCh`

Anyway, what do you think of the code now?

src/github.com/getlantern/fronted/direct.go
```diff
@@ -136,6 +137,13 @@ func (d *direct) Do(req *http.Request) (*http.Response, error) {
 // Dial persistently dials masquerades until one succeeds.
 func (d *direct) Dial(network, addr string) (net.Conn, error) {
 	gotFirst := false
+	candidateCh := <-candidateChCh
+	select {
+	case candidateChCh <- candidateCh:
```
oxtoacart: @myleshorton 

The following sequence will break this:

1. Configure() pushes candidateCh A
2. Dial() pops candidateCh A
3. Configure() pushes candidateCh B (this will succeed because there's room in the channel)
4. Another Dial() pops candidateCh B
5. The first Dial() pushes candidateCh A
6. The second Dial() attempts to push candidateCh B, but the channel is full so it discards it

Now, subsequence dials using candidateCh A even though we want them to use candidateCh B.

You mentioned that Dial() can happen before Configure(), which keeps atomic.Value from working.  You can use eventual.Value for this.

PR #3534, opened 2016-02-09
Made MobileSDK compatible with Xamarin again

oxtoacart:


MobileSDK/sdk/src/main/java/org/lantern/mobilesdk/Lantern.java
```diff
@@ -66,7 +66,7 @@
             System.setProperty("http.proxyPort", port);
             System.setProperty("https.proxyHost", host);
             System.setProperty("https.proxyPort", port);
-            return result;
+            return new StartResult(result.getHTTPAddr(), result.getSOCKS5Addr());
```
myleshorton: Why do you need to create a new result here @oxtoacart?

oxtoacart: Because the other one is defined in liblantern.aar and a binding project in Xamarin can only reference one aar (i.e. sdk.aar).  I had to do this to get Xamarin to actually generate the C# wrapper for the Lantern class.

oxtoacart: Oh hey, I can get around this by making the one Xamarin binding project for sdk.aar depend on the other for liblantern.aar.  Closing PR (sorry for the spam).

myleshorton:
Nice!

Issue #3532, opened 2016-02-09
Add debug target to makefile that includes the -race flag for go


oxtoacart:
This will make sure that when we're running during development, we use race detection by default.

PR #3539, opened 2016-02-10
Added lantern target that includes race detection

oxtoacart:
Closes getlantern/lantern#3532

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ flashlight.yaml
 envvars.bash
 Lantern.app
 lantern.dmg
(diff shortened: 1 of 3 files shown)


oxtoacart:
Note - this doesn't work when using go1.6rc1 because the race detection library was compiled with a newer OS X SDK than what we're using.

oxtoacart:
@myleshorton Thanks for the review!  I'm going to be a rebel and just merge this now.

myleshorton:
=) Meant to merge it so sounds good

PR #3543, opened 2016-02-10
Issue3528

oxtoacart:
@myleshorton - I ended up just using a mutex instead of eventual, because the original logic already handles the case where no candidates are available when dialing, and I didn't want to change the behavior in case something else relies on Dial failing in this situation (rather than pausing while it waits for the candidateCh to become available).

### DIFF
--- a/src/github.com/getlantern/eventual/eventual.go
+++ b/src/github.com/getlantern/eventual/eventual.go
@@ -7,6 +7,11 @@ import (
 	"time"
 )
 
+const (
+	FALSE = 0
+	TRUE  = 1
+)
+
 // Value is an eventual value, meaning that callers wishing to access the value
 // block until the value is available.
 type Value interface {
@@ -25,7 +30,7 @@ type value struct {
 	val      atomic.Value
 	wg       sync.WaitGroup
 	updates  chan interface{}
-	gotFirst bool
+	gotFirst int32
 }
 
 // NewValue creates a new Value.
@@ -52,15 +57,20 @@ func (v *value) Set(val interface{}) {
 func (v *value) processUpdates() {
 	for val := range v.updates {
 		v.val.Store(val)
-		if !v.gotFirst {
+		if v.gotFirst == FALSE {
 			// Signal to blocking callers that we have the first value
 			v.wg.Done()
-			v.gotFirst = true
+			v.gotFirst = TRUE
 		}
 	}
 }
 
 func (v *value) Get(timeout time.Duration) (interface{}, bool) {
+	if atomic.LoadInt32(&v.gotFirst) == TRUE {
+		// Short-cut used once value has been set, to avoid extra goroutine
+		return v.val.Load(), true
+	}
+
 	valCh := make(chan interface{})
 	go func() {
 		v.wg.Wait()
--- a/src/github.com/getlantern/eventual/eventual_test.go
+++ b/src/github.com/getlantern/eventual/eventual_test.go
@@ -28,6 +28,18 @@ func TestSingle(t *testing.T) {
 	assert.Equal(t, "hi", r, "Wrong result")
 }
 
+func BenchmarkGet(b *testing.B) {
+	v := NewValue()
+	go func() {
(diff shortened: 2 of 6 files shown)


myleshorton:
Hmmn I think if anything something will fail because it tries to dial but can't @oxtoacart - anything that would be calling this before Configure is likely to be some super critical startup service that we need to work, such as looking up your config

oxtoacart:
@myleshorton That's existing behavior.  Should I log a new bug for it?

myleshorton:
Well, I just think it's existing behavior that we almost certainly don't want @oxtoacart. We could theoretically merge this, log a bug, and then go back and change this. I'm certainly not opposed to that, but it's also pretty hard to catch, and we could just go with the assumption that whenever something calls `Dial` here it actually wants to be able to make a connection, so needs real masquerades. If you want to do it as separate tickets though that's fine, but basically the race condition revealed existing behavior that was unintended and only occurs when the race happens, which seems to be fairly rare. I also don't think this happens in our released versions, as I run with -race all the time and have never seen it.

myleshorton:
It's also only existing behavior because of the race condition!

oxtoacart:
@myleshorton This uses eventual now.

src/github.com/getlantern/eventual/eventual.go
```diff
@@ -52,15 +57,20 @@ func (v *value) Set(val interface{}) {
...
+			v.gotFirst = TRUE
 		}
 	}
 }
 
 func (v *value) Get(timeout time.Duration) (interface{}, bool) {
+	if atomic.LoadInt32(&v.gotFirst) == TRUE {
+		// Short-cut used once value has been set, to avoid extra goroutine
+		return v.val.Load(), true
+	}
```
oxtoacart: @myleshorton This optimization speeds up Get() by 2 orders of magnitude.  I guess spawning goroutines does cost something.  Time for op went from about 2 microseconds to 20 nanoseconds.

myleshorton: Nice catch!! You caught that while profiling @oxtoacart?

src/github.com/getlantern/eventual/eventual_test.go
```diff
@@ -28,6 +28,18 @@ func TestSingle(t *testing.T) {
...
+	v := NewValue()
+	go func() {
+		time.Sleep(20 * time.Millisecond)
+		v.Set("hi")
+	}()
+
+	for i := 0; i < b.N; i++ {
+		v.Get(20 * time.Millisecond)
+	}
+}
```
myleshorton: Oh you caught it here...I see @oxtoacart!

src/github.com/getlantern/fronted/direct.go
```diff
@@ -180,6 +183,18 @@ func (d *direct) Dial(network, addr string) (net.Conn, error) {
...
+	result, ok := _candidateCh.Get(5 * time.Minute)
+	if !ok {
+		panic("Unable to get candidateCh within 5 minutes")
+	}
+	return result.(chan *Masquerade)
+}
+
+func setCandidateCh(newCh chan *Masquerade) {
+	_candidateCh.Set(newCh)
+}
```
myleshorton: A bit of a nit, but do we really need a separate method here @oxtoacart? I'm just thinking we probably only ever want to set it in `Configure`, but creating the separate method implicitly sort of encourages other code to call it.

myleshorton:
Looks awesome. I just added that one comment on the setter, but it doesn't matter much. Otherwise ready to merge!

src/github.com/getlantern/eventual/eventual_test.go
oxtoacart: I had a hunch it was a little expensive, so I created the benchmark.  Until now, we weren't really using it on anything critical path.  Even in Dial() it was cheap compared with all the other work, but now we can confidently use it liberally without worrying much about performance.

oxtoacart:
@myleshorton I removed that setCandidateCh method.

myleshorton:
Lgtm!!

Issue #3551, opened 2016-02-11
Data race in Eventual


oxtoacart:
==================
WARNING: DATA RACE
Read by goroutine 72:
  sync/atomic.LoadInt32()
      /private/var/folders/vd/7l9ys5k57l91x63sh28wl_kc0000gn/T/workdir/go/src/runtime/race_amd64.s:192 +0xb
  github.com/getlantern/eventual.(*value).Get()
      /Users/ox.to.a.cart/lantern2/src/github.com/getlantern/eventual/eventual.go:69 +0x62
  github.com/getlantern/flashlight/client.(*Client).ServeHTTP()
      /Users/ox.to.a.cart/lantern2/src/github.com/getlantern/flashlight/client/handler.go:31 +0x2a8
  net/http.serverHandler.ServeHTTP()
      /private/var/folders/vd/7l9ys5k57l91x63sh28wl_kc0000gn/T/workdir/go/src/net/http/server.go:1862 +0x206
  net/http.(*conn).serve()
      /private/var/folders/vd/7l9ys5k57l91x63sh28wl_kc0000gn/T/workdir/go/src/net/http/server.go:1361 +0x117c

Previous write by goroutine 29:
  github.com/getlantern/eventual.(*value).processUpdates()
      /Users/ox.to.a.cart/lantern2/src/github.com/getlantern/eventual/eventual.go:63 +0x104

Goroutine 72 (running) created at:
  net/http.(*Server).Serve()
      /private/var/folders/vd/7l9ys5k57l91x63sh28wl_kc0000gn/T/workdir/go/src/net/http/server.go:1910 +0x464
  github.com/getlantern/flashlight/client.(*Client).ListenAndServeHTTP()
      /Users/ox.to.a.cart/lantern2/src/github.com/getlantern/flashlight/client/client.go:110 +0x6d1
  github.com/getlantern/flashlight.Run()
      /Users/ox.to.a.cart/lantern2/src/github.com/getlantern/flashlight/flashlight.go:126 +0x60c
  main.doMain.func2()
      /Users/ox.to.a.cart/lantern2/src/github.com/getlantern/flashlight/main/main.go:181 +0x152

Goroutine 29 (running) created at:
  github.com/getlantern/eventual.NewValue()
      /Users/ox.to.a.cart/lantern2/src/github.com/getlantern/eventual/eventual.go:42 +0x11d
  github.com/getlantern/flashlight/client.NewClient()
[...]

PR #3552, opened 2016-02-11
Fixed data race in eventual

oxtoacart:
closes getlantern/lantern#3551

One can verify this by running `go test -race` in the eventual package.


myleshorton:
LGTM!

PR #3557, opened 2016-02-11
Added full head check closes #3531

### LINKED ISSUE #3531: Successful DDF connections can sometimes return 403 forbidden
This is the result of not testing connections end-to-end, as sometimes Cloudfront connections that succeed in connecting will respond with a 403 when we actually try to use direct domain fronting through them.

myleshorton:


myleshorton:
OK this should be all set @oxtoacart

myleshorton:
I feel like we might be leaking connections somewhere here too @oxtoacart...

oxtoacart:
@myleshorton This looks good to me, but the build fails.  I don't think that's because of this PR though.

PR #3559, opened 2016-02-11
Allow overriding update server URL.

xiam:


src/github.com/getlantern/flashlight/autoupdate/autoupdate.go
```diff
@@ -90,7 +100,7 @@ func applyNext() {
 	if httpClient != nil {
 		err := autoupdate.ApplyNext(&autoupdate.Config{
 			CurrentVersion: Version,
-			URL:            serviceURL,
+			URL:            updateServerURL + "/update",
```
myleshorton: Why not just let the update service URL specify the full path always @xiam?

xiam: Because this endpoint is not expected to change regardless of the server URL, I think that keeping this as a constant value is better and future safe (like if we ever want to add another endpoint for whatever reason, we could use `updateServerURL + "/newendpoint" ` or even versioning over the same endpoint `updateServerURL + "/v2.0/update"`).

myleshorton: It still seems like it would be better to control all of that on the server side, though, since then we could make it whatever we want always. It's a really minor point, though -- merging!

myleshorton:
OK merging thanks @xiam!

PR #3585, opened 2016-02-17
Valencia service

oxtoacart:
@myleshorton @atavism

This adds support for running the embedded Lantern as a service via `Lantern.enableAsService()`. The testbed app is updated to support testing both regular embedded as well as service-mode embedded.


MobileSDK/sdk/src/main/java/org/lantern/mobilesdk/Lantern.java
```diff
@@ -44,44 +35,80 @@
...
     }
 
-    private static go.lantern.Lantern.StartResult enable(String configDir, int timeoutMillis)
-            throws LanternNotRunningException {
+    /**
+     * Note - we use dynamic class loading to avoid loading unused classes into the caller's
+     * classloader (i.e. to avoid loading native dependencies when running as service). This is
+     * important because in some situations, it appears that including the Lantern native library
+     * inside the same process as an application can cause instability on some phones (e.g. Samsung
+     * Galaxy S4).
```
oxtoacart: This comment is important - it turns out that simply loading the native library into the Manoto app introduced instability.  I structured Lantern so that simply importing it does not cause the native library to get loaded.  That now happens only when the proxy is started, which when running as a service happens in a separate process.

LanternMobileTestbed/app/src/main/AndroidManifest.xml
```diff
@@ -21,6 +21,12 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+
+        <service
+            android:name="org.lantern.mobilesdk.service.LanternService"
+            android:exported="false"
+            android:process=":LanternService"
+            />
```
oxtoacart: To be able to run the LanternService, it needs to be declared in the app's manifest.  And, it needs to be given a process name to force it to be forked.

atavism: Suppose I'm running multiple apps that integrate Lantern--do we want to have this service run in a global process, and have every app share the same process? It sounds like that can be achieved by making the first letter of the process name lowercase. Whenever you go to start the service, I think if there's already a running process associated, it'll just use that automatically, but I'm not sure. This brings up a good question too: what's the behavior like if I'm running the Lantern VPN app and an app that integrates Lantern as a service simultaneously?

oxtoacart: Good questions.

1. Making the service global seems like a good idea

2. If you're running the Lantern VPN app and an app the embeds Lantern simultaneously, I imagine that the embedded Lantern will proxy via the VPN, so we'll be doing double-duty on both the client and server-side.

oxtoacart:
@atavism @myleshorton What do you think?  Merge worthy?

myleshorton:
I'm still knocking this around a little bit @oxtoacart

MobileSDK/sdk/src/main/java/org/lantern/mobilesdk/embedded/EmbeddedLantern.java
```diff
@@ -0,0 +1,30 @@
...
+import org.lantern.mobilesdk.Lantern;
+import org.lantern.mobilesdk.LanternNotRunningException;
+import org.lantern.mobilesdk.StartResult;
+
+public class EmbeddedLantern extends Lantern {
+    static {
+        // Track extra info about Android for logging to Loggly.
+        EmbeddedLantern.addLoggingMetadata("androidDevice", android.os.Build.DEVICE);
+        org.lantern.mobilesdk.Lantern.addLoggingMetadata("androidModel", android.os.Build.MODEL);
+        org.lantern.mobilesdk.Lantern.addLoggingMetadata("androidSdkVersion", "" + android.os.Build.VERSION.SDK_INT + " (" + android.os.Build.VERSION.RELEASE + ")");
```
myleshorton: Do we not want to log to Loggly in the service case @oxtoacart? I can definitely see not doing it, but want to make sure that's intentional.

oxtoacart: I think it makes sense to log to Loggly so that we get feedback on any issues that might be encountered by users of the embedded version.

myleshorton: Yeah I'm guess I'm asking the opposite: do we also want to log to Loggly for the service version?

oxtoacart: Oh, now I understand your question.  The service uses EmbeddedLantern, so they both already log to Loggly.

myleshorton: Oh I see - I missed that! Cool thanks @oxtoacart.

MobileSDK/sdk/src/main/java/org/lantern/mobilesdk/Lantern.java
```diff
@@ -6,26 +6,17 @@
...
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
- * Created by ox.to.a.cart on 1/28/16.
+ * API for embedding the Lantern proxy
  */
-public class Lantern {
+public abstract class Lantern {
```
atavism: I went to build a new version of FireTweet that uses the new SDK. I was successful up to the point of trying to sign into Twitter. This page is loaded in a WebView, and we had a special method in the old SDK for configuring WebViews to inherit an app's proxy settings. See https://github.com/getlantern/lantern/blob/lantern-mobile/src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/sdk/Lantern.java#L139 and https://github.com/getlantern/lantern/blob/lantern-mobile/src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/sdk/ProxySettings.java Any possibility we could restore this? I just want to verify running Lantern as a service works in FireTweet.

oxtoacart: The `enable()` method returns the address of the proxy.  Can you just use that from within Firetweet to configure the webview?

atavism: Yeah, that works. I wanted to see how Lantern as a service works with FireTweet since there's been instability issues in the past too. I also think the functionality to configure a webview to use our proxy is really useful and should be part of the SDK but we can add that back in a separate PR. 

One other thing I noticed is that FireTweet is using an old version of Google's analytics library, and integrating the SDK actually caused the app to crash until I upgraded for whatever reason. It's working fine now though and I've uploaded the new version here: https://raw.githubusercontent.com/firetweet/downloads/master/firetweet.apk

myleshorton:
All good on this PR from my end @atavism and @oxtoacart.

atavism:
Same here. Let's go head and merge this! Thanks, @oxtoacart!

oxtoacart:
@atavism @myleshorton Thanks for the review!  I've opened #3633 for the WebView thing.

Issue #3587, opened 2016-02-17
Keep masquerades in only one place


oxtoacart:
Right now, we have various tests (like util_test.go) that duplicate the masquerades configured in masquerades.go.  These tests have to be manually update when we updates masquerades.go, which is burdensome and error prone.  It would be better to have everything use a single configuration.

myleshorton:
I believe you knocked this out right @oxtoacart?

oxtoacart:
Yes, I think so.

PR #3588, opened 2016-02-17
No longer continually vetting masquerades

oxtoacart:
Sorry the diff on util_test is so big, I just added all the masquerades because it was failing with its initial masquerades.  Eventually, we should just have one place where we keep the list of masquerades (see #3587).


src/github.com/getlantern/fronted/direct.go
```diff
@@ -76,26 +77,28 @@ func (d *direct) vet(initial map[string][]*Masquerade) {
...
-		go d.vetSome()
+func (d *direct) vetInitial() {
+	log.Debugf("Vetting %d initial candidates in parallel", NumberToVetInitially)
+	for i := 0; i < NumberToVetInitially; i++ {
+		go d.vetOne()
 	}
 }
 
-func (d *direct) vetSome() {
+func (d *direct) vetOne() {
```
oxtoacart: This goroutine now terminates once it has successfully verified one masquerade.  So, when we reconfigure, we don't end up piling up a bunch of goroutines.  Also, we don't do a bunch of work vetting masquerades that we may never use.

fffw:
LGTM. Travis failed because of an unresolved dependency, all relevant test were passed locally.

fffw:
Plus: The unnecessary dependency of Travis is removed in valencia. Manually running Lantern for 1+ hours on Mac, memory keeps at 30MB.

oxtoacart:
@fffw Thanks for reviewing!

myleshorton:
Awesome Ox and thanks for testing Merlin! I'll get this into a new build
ASAP.

PR #3596, opened 2016-02-17
Lantern mobile: remove obj folder

atavism:
This removes the obj folder which holds compiled binary files generated during a build we don't want in version control


atavism:
Just merging this since it's super straightforward!

PR #3620, opened 2016-02-22
Added example proxy for detour

oxtoacart:


src/github.com/getlantern/detour/proxy/proxy.go
```diff
@@ -0,0 +1,31 @@
+// package main provides a simple proxy program that uses detour, useful for
```
oxtoacart: @fffw This is an example proxy for testing detour interactively.  If you test with boom, you'll find that it hangs on about 20% of requests.

```
boom -disable-keepalive -n 100 -x http://localhost:8080 http://www.google.com/humans.txt
```

fffw:
It's a neat tool to find performance issue @oxtoacart ! Will merge it

PR #3623, opened 2016-02-22
Keep configuration files on uninstall closes #3607

### LINKED ISSUE #3607: Don't empty the %APPDATA%\Lantern directory when uninstalling
It is an unexpected empty my files place on %APPDATA%\Lantern dir (some try and hacks for yaml).
While emptying it makes sense, but I hope to have a checklist for files be deleted and confirm for unexpected files when uninstalling. Automatic update within the software is correct, you don't empty it, but rename the exe file to ".old" and other files is retained.

As well as, the installer should not be starts uninstall and install automatically after launch it, it is an unexpected behavior, although very intelligent and automatic.

myleshorton:


myleshorton:
Hey @fffw you want to take a look at this one?

fffw:
We may also need to remove below 3 files if they exist. They are not changed after installation.
```
    File lantern.ico
    File .packaged-lantern.yaml
    File lantern.yaml
```
It's also a plus if we can confirm before uninstall, or double click on Control Panel will uninstall it directly. But that's beyond the purpose of this PR.

fffw:
Ok adding below two lines before sections can do the trick. It doesn't affect silent uninstall.

```
UninstPage uninstConfirm
UninstPage instfiles
```

myleshorton:
Thanks @fffw! I just for now added deleting those few extra files, as it seems like ideally using UninstPage would mean using Page/UninstPage for everything, which is maybe more of a destabilizing change than would be ideal at this stage on Valencia.

myleshorton:
OK I'm going to merge this in with those fixes you noted @fffw.

PR #3628, opened 2016-02-23
Allow users to configure whether Lantern should set the system proxy

myleshorton:


myleshorton:
Hey @oxtoacart you mind taking a look at this one?

myleshorton:
Or @atavism you want to give this one a look?

oxtoacart:
Sure, I'm kind of in this area for the Windows stuff anyway.

src/github.com/getlantern/flashlight/main/settings.go
```diff
@@ -73,91 +72,114 @@ func Load(version, revisionDate, buildDate string) {
...
+// SetSystemProxy sets whether or not to set system proxy when lantern starts
+func (s *Settings) SetSystemProxy(enable bool) {
+	s.Lock()
+	defer s.Unlock()
+	s.SystemProxy = enable
+	if enable {
+		go pacOn()
+	} else {
+		go pacOff()
 	}
```
oxtoacart: @myleshorton It's very unlikely to happen in practice, but if this function is called in rapid succession, there's no guarantee that these goroutines will run in the right order.  Better would be to set up a chan bool and a single goroutine that toggles based on what it gets from that channel.

src/github.com/getlantern/flashlight/main/settings.go
```diff
@@ -73,91 +72,114 @@ func Load(version, revisionDate, buildDate string) {
...
+
+// SetSystemProxy sets whether or not to set system proxy when lantern starts
+func (s *Settings) SetSystemProxy(enable bool) {
+	s.Lock()
+	defer s.Unlock()
+	s.SystemProxy = enable
+	if enable {
+		pacOn()
+	} else {
+		pacOff()
```
oxtoacart: The handler for updating this from the UI already uses a buffered channel, so the simplest thing to do to ensure consistent ordering is to just not spawn goroutines for pacOn and pacOff.

oxtoacart:
LGTM!  I just made one small tweak to not launch goroutines for the pac settings change.

oxtoacart:
Once travis check passes, I'll merge.

oxtoacart:
@myleshorton Sorry, I just remembered that I need to integrate this with the ui.lantern.io change.  Working on that now, will submit a PR to the PR.

myleshorton:
Awesome thanks @oxtoacart! The other thing we need is to add logic to not
set the UI to ui.lantern.io if the user does not have Lantern set as the
system proxy.

PR #3636, opened 2016-02-24
Use fake domain for lantern-ui, closes #3363

### LINKED ISSUE #3363: Edge on Windows 10 can't open Lantern page
image

[http://localhost:16823]() can be opened instead. `curl 127.0.0.1:16823` also works. `http://127.0.0.1:16823` can be opened in Chrome as well.

Edge 20.10240.16384.0, running from MsEdge + Win 10 Vagrant package downloaded from https://dev.windows.com/en-us/microsoft-edge/tools/vms/mac/.

oxtoacart:


Makefile
```diff
@@ -157,32 +157,27 @@ all: binaries
 android-dist: genconfig android
 
 # This is to be called within the docker image.
-docker-genassets: require-npm
+genassets: require-npm
```
oxtoacart: There's no reason genassets has to run in Docker, so running locally now.

Dockerfile
```diff
@@ -58,7 +58,7 @@ RUN yum -y install nodejs && yum clean packages
 RUN npm install -g gulp
 
 # Getting Go.
-ENV GO_VERSION go1.6rc1
+ENV GO_VERSION go1.6
```
oxtoacart: Using latest released Go in docker.

Makefile
```diff
@@ -157,32 +157,27 @@ all: binaries
...
 	DEST="src/github.com/getlantern/flashlight/ui/resources.go" && \
-	\
-	if [ "$$UPDATE_DIST" ]; then \
-			cd $$LANTERN_UI && \
-			npm install && \
-			rm -Rf dist && \
-			gulp build && \
-			cd -; \
-	fi && \
+	cd $$LANTERN_UI && \
```
oxtoacart: I got burned by a corrupted dist folder.  Now, we always update dist when generating assets.

Makefile
```diff
@@ -199,10 +194,10 @@ docker-linux-arm:
 	$(call build-tags) && \
 	CGO_ENABLED=1 CC=arm-linux-gnueabi-gcc CXX=arm-linux-gnueabi-g++ CGO_ENABLED=1 GOOS=linux GOARCH=arm GOARM=7 go build -a -o lantern_linux_arm -tags="$$BUILD_TAGS" -ldflags="$(LDFLAGS) $$EXTRA_LDFLAGS -linkmode internal -extldflags \"-static\"" github.com/getlantern/flashlight/main
 
-docker-windows-386:
+windows-386: require-assets
```
oxtoacart: There's no need to build windows in docker, so building it locally now.

oxtoacart: In fact, the only thing that still needs docker is the Linux build.

Makefile
```diff
@@ -374,8 +338,21 @@ package-windows: require-version windows
 	@echo "Generating distribution package for windows/386..." && \
 	if [[ -z "$$SECRETS_DIR" ]]; then echo "SECRETS_DIR environment value is required."; exit 1; fi && \
 	if [[ -z "$$BNS_CERT_PASS" ]]; then echo "BNS_CERT_PASS environment value is required."; exit 1; fi && \
-	$(call docker-up) && \
```
oxtoacart: package-windows can run locally as well.

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -179,3 +191,29 @@ func (client *Client) Configure(cfg *ClientConfig, proxyAll func() bool) {
 func (client *Client) Stop() error {
 	return client.l.Close()
 }
+
+func (client *Client) proxiedDialer(orig func(network, addr string) (net.Conn, error)) func(network, addr string) (net.Conn, error) {
```
oxtoacart: I consolidated the dialing logic into on place that's used by both intercept() and ReverseProxy.

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -179,3 +191,29 @@ func (client *Client) Configure(cfg *ClientConfig, proxyAll func() bool) {
...
+	var proxied func(network, addr string) (net.Conn, error)
+	if client.ProxyAll() {
+		proxied = orig
+	} else {
+		proxied = detour.Dialer(orig)
+	}
+
+	return func(network, addr string) (net.Conn, error) {
+		if isLanternSpecialDomain(addr) {
```
oxtoacart: This is the magic.  The UI is actually accessed via the LanternSpecialDomain, which we detect here and then dial to the actual UI server.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -325,11 +324,6 @@ func (updated *Config) applyFlags(flags map[string]interface{}) error {
 			updated.CloudConfigCA = value.(string)
 		case "instanceid":
 			updated.Client.DeviceID = value.(string)
-
-		// HTTP-server
-		case "uiaddr":
```
oxtoacart: There's no need for uiaddr to be in the config, and even stranger I noticed that we were pushing updates to it via the cloud.yaml.  uiaddr now just comes from flags at runtime.

src/github.com/getlantern/flashlight/main/flags.go
```diff
@@ -3,18 +3,25 @@ package main
...
+	configdir          = flag.String("configdir", "", "directory in which to store configuration, including flashlight.yaml (defaults to current directory)")
+	cloudconfig        = flag.String("cloudconfig", "", "optional http(s) URL to a cloud-based source for configuration updates")
+	cloudconfigca      = flag.String("cloudconfigca", "", "optional PEM encoded certificate used to verify TLS connections to fetch cloudconfig")
+	instanceid         = flag.String("instanceid", "", "instanceId under which to report stats to statshub. If not specified, no stats are reported.")
+	registerat         = flag.String("registerat", "", "base URL for peer DNS registry at which to register (e.g. https://peerscanner.getiantem.org)")
+	country            = flag.String("country", "xx", "2 digit country code under which to report stats. Defaults to xx.")
+	cpuprofile         = flag.String("cpuprofile", "", "write cpu profile to given file")
+	memprofile         = flag.String("memprofile", "", "write heap profile to given file")
+	uiaddr             = flag.String("uiaddr", "127.0.0.1:16823", "if specified, indicates host:port the UI HTTP server should be started on")
```
oxtoacart: Note how we default uiaddr

src/github.com/getlantern/flashlight/main/main.go
```diff
@@ -35,15 +34,6 @@ import (
 var (
 	log = golog.LoggerFor("flashlight")
 
-	// Command-line Flags
-	help               = flag.Bool("help", false, "Get usage help")
```
oxtoacart: Moved to flags.go for consistency.

src/github.com/getlantern/flashlight/main/main.go
```diff
@@ -217,16 +207,10 @@ func beforeStart(cfg *config.Config) bool {
 		//
 		// See: https://github.com/getlantern/lantern/issues/2776
 		log.Debug("Clearing proxy settings")
-		doPACOff(fmt.Sprintf("http://%s/proxy_on.pac", cfg.UIAddr))
+		doPACOff(fmt.Sprintf("http://%s/proxy_on.pac", *uiaddr))
 		exit(nil)
 	}
 
-	log.Debug("Starting client UI")
```
oxtoacart: Moved resolution into ui.go.

src/github.com/getlantern/pac/pac2go.sh
```diff
@@ -18,13 +18,12 @@ BINPATH=../pac-cmd/binaries
...
 mv $BINPATH/windows/pac.exe $BINPATH/windows/pac
 osslsigncode sign -pkcs12 "$BNS_CERT" -pass "$BNS_CERT_PASS" -in $BINPATH/windows/pac -out $BINPATH/windows/pac || die "Could not sign windows"
-go-bindata -nomemcopy -nocompress -pkg pac -prefix $BINPATH/windows -o pac_bytes_windows.go $BINPATH/windows    
+go-bindata -nomemcopy -nocompress -pkg pac -prefix $BINPATH/windows -o pac_bytes_windows.go $BINPATH/windows
 mv $BINPATH/windows/pac $BINPATH/windows/pac.exe
 
 go-bindata -nomemcopy -nocompress -pkg pac -prefix $BINPATH/linux_386 -o pac_bytes_linux_386.go $BINPATH/linux_386
 go-bindata -nomemcopy -nocompress -pkg pac -prefix $BINPATH/linux_amd64 -o pac_bytes_linux_amd64.go $BINPATH/linux_amd64
-go-bindata -nomemcopy -nocompress -pkg pac -prefix $BINPATH/linux_arm -o pac_bytes_linux_arm.go $BINPATH/linux_arm
+#go-bindata -nomemcopy -nocompress -pkg pac -prefix $BINPATH/linux_arm -o pac_bytes_linux_arm.go $BINPATH/linux_arm
```
oxtoacart: There is no linux_arm binary.

src/github.com/getlantern/tarfs/tarfs.go
```diff
@@ -152,8 +153,7 @@ func (fs *FileSystem) get(p string, ignoreempty bool) ([]byte, error) {
 	}
 	b, found := fs.files[p]
 	if !found {
-		err := fmt.Errorf("%v not found", p)
-		return nil, err
+		return nil, os.ErrNotExist
```
oxtoacart: When I had a bad dist folder, I was missing files.  However, the web server reported a 500 error, which was confusing.  Returning os.ErrNotExist here causes it to behave normally and return a 404.

Makefile
xiam: Not sure if this is an advantage, what's the main benefit of running this locally? On the other hand, the reason to run build tasks inside docker is that everyone has the same environment, the same version of node, npm and everything else, so it may look like we could skip docker but I think that having that guarantee is important, specially with node-npms's ecosystem which already caused *it-works-for-me* problems in the past.

Makefile
xiam: I don't think we benefit a lot from this. It already happened some months ago that I was running a beta version of Go locally (in OSX, which we cannot build inside docker) that created binaries that crashed upon starting, nobody had the same problem as me because they haven't updated Go versions to the latest beta, I don't run Go betas anymore. Using the docker image prevents the same problems on Windows and Linux, we can be sure we have exactly the same build. We cannot do that for OSX and that's a shame, but we can be on the safe side with Linux and Windows (Windows being the most used) and I really think that saves debugging time.

Makefile
xiam: Note that by running this locally it's also a requisite to have `osslsigncode` and `makensis` locally as well, and we are not providing instructions for that anywhere, the person who is building needs to figure out how to do that instead of just using the docker image which already has those binaries.

oxtoacart: Actually I provided instructions in the README.

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -5,18 +5,30 @@ import (
...
 	"github.com/getlantern/balancer"
+	"github.com/getlantern/detour"
 	"github.com/getlantern/eventual"
 	"github.com/getlantern/golog"
 )
 
+const (
+	// LanternSpecialDomain is a special domain for use by lantern that gets
+	// resolved to localhost by the proxy
+	LanternSpecialDomain          = "ui.lantern.io"
```
xiam: This is awesome!

Makefile
oxtoacart: I see your point about consistency.  Docker isn't perfect here either because we're installing using apt-get, which just gets the latest version of everything, but it's better than relying on the local environment.

The big advantage to not using docker for daily development is that people in resource constrained environments can build more easily/cheaply because they don't have to install and run a docker machine.

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -179,3 +191,29 @@ func (client *Client) Configure(cfg *ClientConfig, proxyAll func() bool) {
...
+			rewritten := rewriteLanternSpecialDomain(addr)
+			log.Tracef("Rewriting %v to %v", addr, rewritten)
+			return net.Dial(network, rewritten)
+		}
+		return proxied(network, addr)
+	}
+}
+
+func isLanternSpecialDomain(addr string) bool {
+	return strings.Index(addr, LanternSpecialDomainWithColon) == 0
```
xiam: What's the reason for the constant with a colon? Is this something we could check with https://golang.org/pkg/strings/#HasPrefix and `LanternSpecialDomain`?

src/github.com/getlantern/flashlight/main/main.go
```diff
@@ -305,17 +297,22 @@ func parseFlags() {
...
-	url := "http://" + tcpAddr + "/startup"
+func showExistingUi(addr string) error {
+	url := "http://" + addr + "/startup"
 	log.Debugf("Hitting local URL: %v", url)
 	resp, err := http.Get(url)
 	if err != nil {
 		log.Debugf("Could not hit local lantern")
-		if err = resp.Body.Close(); err != nil {
-			log.Debugf("Error closing body! %s", err)
+		if resp.Body != nil {
```
xiam: We checked `err != nil` above and that ensures that `resp.Body` is not nil.

xiam: Sorry, I take that back, this is an error, not the lack of it.

src/github.com/getlantern/flashlight/main/pac.go
```diff
@@ -76,31 +76,27 @@ func genPACFile() {
 	}
 	formatter :=
 		`var bypassDomains = %s;
-		function FindProxyForURL(url, host) {
-			if (isPlainHostName(host) // including localhost
-			|| shExpMatch(host, "*.local")) {
+	function FindProxyForURL(url, host) {
```
xiam: Should we hardcode `ui.lantern.io` here to make sure it always passes through Lantern or that's not required?

Makefile
xiam: Good point, docker has a cost and it takes a lot of space. So, we could use docker only when building a distribution package, not for the daily development. That way we can be sure the final build is the same for Linux/Windows. How does it sound?

Perhaps it would be better to keep track of the solution for this on a different issue, it looks like the UI address issue does not depend on this problem.

Makefile
xiam: Oh yes, right, I didn't see them initially.

Makefile
```diff
@@ -265,8 +242,8 @@ docker: system-checks
 	cp Dockerfile $$DOCKER_CONTEXT && \
 	docker build -t $(DOCKER_IMAGE_TAG) $$DOCKER_CONTEXT;
 
-update-dist:
-	UPDATE_DIST=true make genassets
+# Kept for backward-compatibility, but no different than genassets
+update-dist: genassets
```
xiam: I think we can get rid of this, nothing external depends on this target.

src/github.com/getlantern/pac/pac2go.sh
xiam: Do yo know anything about this @uaalto?

uaalto: I added this back when I did the port to Raspberry Pi. I don't remember the details, and it's been a while that I don't compile it, but it was certainly working.

xiam:
@oxtoacart This is awesome! so users will be able to access Lantern by typing `ui.lantern.io` and having it running, right? We then can have a default *real* page for `ui.lantern.io` with instructions for running Lantern and we can get rid of all the odd numbers appearing in the address bar.

Do you think we could we move the Makefile changes into another branch/issue and discuss them there?

src/github.com/getlantern/flashlight/client/client.go
fffw: Just a quick note: We were advertising getlantern.org for a long time, would some domain like `start.getlantern.org` or `launch.getlantern.org` be more familiar to user and help to strengthen the brand?

Makefile
myleshorton: I think that would probably be wise @xiam, as we also want to push out valencia ASAP so we can get back to pro, so we probably want to keep all pieces as stable as possible (i.e. this isn't on the critical path for release).

I would argue we ideally would have our dev environments be as close as possible to production to essentially get free QA, but I don't feel strongly about docker versus local dev environment at first glance.

oxtoacart: The only thing that put it on the critical path for me is that for some reason, I can't package Windows when running in Docker.  If it's okay with you guys, I think I have a solution that I can show in a couple of hours.

src/github.com/getlantern/pac/pac2go.sh
oxtoacart: Is it possible that the linux_arm executable never got checked in?

src/github.com/getlantern/flashlight/client/client.go
oxtoacart: Unfortunately I can't use a subdomain of getlantern.org because that domain was configured with HSTS, including for subdomains.  So, Chrome always switches over to an HTTPS connection, which our embedded UI server doesn't support.

src/github.com/getlantern/flashlight/client/client.go
oxtoacart: Including the colon is just a little bit stricter, so we wouldn't match something like `ui.lantern.io.justkidding:80`, only match `ui.lantern.io`.

src/github.com/getlantern/flashlight/main/main.go
oxtoacart: Yeah, I actually got a panic here during development, that's what prompted me to fix it.

src/github.com/getlantern/flashlight/main/pac.go
oxtoacart: I originally did that, but Microsoft Edge seems very sensitive to the PAC file and my change actually broke that browser.  It turns out that we don't need it, because ui.lantern.io will never get added to the exceptions list because it doesn't go through detour.

Makefile
oxtoacart: Will do.

oxtoacart:
@xiam Thanks for the review!  I agree about the Makefile changes, I'll move those into another PR.

I think the only thing about this change to use ui.lantern.io that's not optimal is that it requires traffic to go through our proxy.  If someone is using a proxy switcher that bypasses our proxy, ui.lantern.io won't work.

Makefile
xiam: I can take a look into that Windows build problem, it could be something wrong with the Dockerfile, could you open an issue with the error message and assign it to me?

Makefile
```diff
@@ -436,9 +436,9 @@ release-qa: require-version require-s3cmd
 	cp lantern-manoto_*amd64.deb $$BASE_NAME_MANOTO-64-bit.deb && \
 	cp lantern-installer.apk $$BASE_NAME.apk && \
 	for NAME in $$(ls -1 $$BASE_NAME*.*); do \
-		shasum $$NAME | cut -d " " -f 1 > $$NAME.sha1 && \
-		echo "Uploading SHA-1 `cat $$NAME.sha1`" && \
-		$(S3CMD) put -P $$NAME.sha1 s3://$(S3_BUCKET) && \
+		shasum -a 256 $$NAME | cut -d " " -f 1 > $$NAME.sha256 && \
```
oxtoacart: @xiam @myleshorton I don't know why this is showing as a diff, this is what's on valencia (I just copied the Makefile).

src/github.com/getlantern/pac/pac2go.sh
myleshorton: Yeah I think we might as well just keep it around. We could always publish it with releases to see if anyone plays with it.

oxtoacart: @uaalto If you can find the binary, you can add it and uncomment this line of the script.  As it stands, we still have the pac_bytes_linux_arm.go from whenever that got generated, so people can still run Lantern on arm.  Until then, I have to leave this commented-out since the script won't run successfully otherwise.

Makefile
myleshorton: I'm guessing you just created your branch before that change was merged into valencia yesterday @oxtoacart, but not to worry.

oxtoacart: I did, though I don't understand why GitHub would show this as a diff.  At some point I should probably read a book about git :)

myleshorton:
> I think the only thing about this change to use ui.lantern.io that's not optimal is that it requires traffic to go through our proxy. If someone is using a proxy switcher that bypasses our proxy, ui.lantern.io won't work.

That's interesting @oxtoacart. Especially with the change to allow users to set whether Lantern should set the system proxy with the changes in https://github.com/getlantern/lantern/pull/3628, we might want to just check to make sure Lantern is set as the system proxy, and if it's not just open the UI at 127. That would of course break things on Edge, but that at least limits the damage to users who are: 

1) running with Edge as their default browser 
2) have customized Lantern to not use the system proxy

That's probably a pretty small minority, especially since most users doing #2 are probably using a proxy switching extension, and I don't know of any of those for edge.

oxtoacart:
@myleshorton Yeah, edge doesn't actually allow any extensions right now, so there's definitely no proxy switchers for it :)

So for #3628, I'll use the loopback address for showing the UI if the setting to set Lantern as system proxy is off.  I'll address that as part of that ticket.

myleshorton:
Sounds great @oxtoacart

src/github.com/getlantern/flashlight/main/pac.go
```diff
@@ -76,31 +76,27 @@ func genPACFile() {
...
-			if (/^[0-9.]+$/.test(host)) {
-				if (isInNet(host, "10.0.0.0", "255.0.0.0") ||
-				isInNet(host, "172.16.0.0",  "255.240.0.0") ||
-				isInNet(host, "192.168.0.0",  "255.255.0.0") ||
-				isInNet(host, "127.0.0.0", "255.255.255.0")) {
-					return "DIRECT";
-				}
-			}
-			// Lantern desktop version proxies only http and https
-			if (url.substring(0, 4) != 'http') {
```
myleshorton: AFAICT this is the only actual change here to the PAC other than whitespace right @oxtoacart? ui.lantern.io is still over HTTP, so why is this change necessary?

oxtoacart: Hmm, there shouldn't be a change here at all.  Hang on.

oxtoacart: It's a confusing diff, but it's just whitespace changes.  I'm tempted to stick with this version since it worked in Edge and other browsers.

myleshorton: Wait, but the new version doesn't have this right:

```
			// Lantern desktop version proxies only http and https
			if (url.substring(0, 4) != 'http') {
				return "DIRECT";
			}
```

oxtoacart: Oh crap, you're right.  Okay, let me fix that and retest in Edge.

oxtoacart:
@myleshorton Pac file is reverted to original, works in Edge.

myleshorton:
OK cool. These changes LGTM! Will merge when travis finishes.

PR #3639, opened 2016-02-24
adding getiantem.org closes #3638

### LINKED ISSUE #3638: Not proxying getiantem.org by default
This means that the first time we try to access the config, detour will dynamically determine if it's blocked. We should probably just always proxy it.

myleshorton:
About as simple as a changelog itself can be, but it would be useful to review the underlying thinking here. I basically noticed in our logs that detour was dynamically determining what to do with getiantem.org

### DIFF
--- a/src/github.com/getlantern/flashlight/genconfig/proxiedsites/original.txt
+++ b/src/github.com/getlantern/flashlight/genconfig/proxiedsites/original.txt
@@ -1,3 +1,4 @@
+getiantem.org
 manototv.com
 firetweet.io
 herokuapp.com


myleshorton:
You want to give this a think @oxtoacart?

oxtoacart:
@myleshorton So this change automatically includes getiantem.org in the proxied sites list?  Should we go ahead and run genconfig along with this to pick that up?

myleshorton:
Yes we definitely need to @oxtoacart. I generally do that just before building any new release (unless we're releasing a bunch of versions very quickly like in the 2.0.13-2.0.16 batch).

myleshorton:
Building right before the release allows the masquerades to be as up to date as possible though.

oxtoacart:
@myleshorton Makes sense.

PR #3640, opened 2016-02-24
Make Docker optional

### LINKED ISSUE #3637: Unable to package windows with Docker
If I run the windows packaging script in docker, it fails with the below error.  Running the exact same logic locally works.  I am specifying SECRETS_DIR and BNS_CERT_PASS environment variables, and the SECRETS_DIR contains bns.pfx (not a symlink, but the actual file).  Note - I previously ran without that file, and got a different error (unable to open file or something like that).

```
Building windows/386...
Using /usr/local/bin/docker-machine
Build tags:
Extra ldflags: -X=github.com/getlantern/flashlight.compileTimePackageVersion=2.0.17
Generating distribution package for windows/386...
Using /usr/local/bin/docker-machine
Build tags:
Extra ldflags: -X=github.com/getlantern/flashlight.compileTimePackageVersion=2.0.17
/bin/bash: line 15:   898 Killed                  osslsigncode sign -pkcs12 "$BNS_CERT" -pass "$BNS_CERT_PASS" -n "Lantern" -t http://timestamp.verisign.com/scripts/timstamp.dll -in "lantern_windows_386.exe" -out "$INSTALLER_RESOURCES/lantern.exe"
Makefile:238: recipe for target 'docker-package-windows' failed
make: *** [docker-package-windows] Error 137
make: *** [package-windows] Error 2
```

oxtoacart:
@xiam @myleshorton This PR allows us to use docker or not, depending on an environment variable.

This isn't finished yet, but the basic concept is in place.  I'm going to pivot to higher priority stuff and revisit this later.

As a side-effect, this closes #3637.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -563,6 +563,9 @@ android-install: $(LANTERN_MOBILE_ANDROID_DEBUG)
 clean-assets:
 	rm -f $(RESOURCES_DOT_GO)
 
+# Provided for backward compatibility with how people used to use the makefile
+update-dist: clean-assets assets
+
 clean-bns-cert:
 	rm -f $(BNS_CERT)
 
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,9 @@ PACKAGE_URL := https://www.getlantern.org
 PACKAGED_YAML := .packaged-lantern.yaml
 MANOTO_YAML := .packaged-lantern-manoto.yaml
 
+RESOURCES_DOT_GO := ./src/github.com/getlantern/flashlight/ui/resources.go
+BNS_CERT := bns.pfx
+
 LANTERN_BINARIES_PATH ?= ../lantern-binaries
 
 GH_USER ?= getlantern
@@ -94,29 +97,6 @@ define build-tags
 	EXTRA_LDFLAGS=$$(echo $$EXTRA_LDFLAGS | xargs) && echo "Extra ldflags: $$EXTRA_LDFLAGS"
 endef
 
-define docker-up
-	if [[ "$$(uname -s)" == "Darwin" ]]; then \
-		if [[ -z "$(DOCKERMACHINE)" ]]; then \
-		  if [[ -z "$(BOOT2DOCKER)" ]]; then \
-  			echo 'Missing "docker-machine" command' && exit 1; \
-			fi && \
-			echo "Falling back to using $(BOOT2DOCKER), recommend upgrading to latest docker toolbox from https://www.docker.com/docker-toolbox" && \
-			if [[ "$$($(BOOT2DOCKER) status)" != "running" ]]; then \
-				$(BOOT2DOCKER) up; \
-			fi && \
-			if [[ -z "$$DOCKER_HOST" ]]; then \
-				$$($(BOOT2DOCKER) shellinit 2>/dev/null); \
-			fi \
-		else \
-		  echo "Using $(DOCKERMACHINE)" && \
-			if [[ "$$($(DOCKERMACHINE) status default)" != "Running" ]]; then \
-				$(DOCKERMACHINE) start default; \
-			fi && \
-			$$($(DOCKERMACHINE) env default 2>/dev/null | head -4 | tr -d '"'); \
-		fi \
-	fi
-endef
-
 define fpm-debian-build =
 	echo "Running fpm-debian-build" && \
  	PKG_ARCH=$1 && \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -117,6 +117,30 @@ define docker-up
 	fi
 endef
 
+# The dmake macro runs the current target in docker if and only if the
```
oxtoacart: @xiam @myleshorton This is the heart of the change.  This macro allows us to run any target either directly or in docker depending on the `USEDOCKER` environment variable.

Makefile
```diff
@@ -153,60 +133,85 @@ define fpm-debian-build =
...
+				$(DOCKERMACHINE) start default; \
+			fi && \
+			$$($(DOCKERMACHINE) env default 2>/dev/null | head -4 | tr -d '"'); \
+		fi \
+	fi
+endef
+
+# This implicit rule allows prefix any existing target with "docker-" to make it
+# run in docker.
+docker-%: system-checks
```
oxtoacart: @xiam I'm now using an implicit target that allows us to prefix any target with `docker-` which then runs it in docker.

Makefile
```diff
@@ -153,60 +133,85 @@ define fpm-debian-build =
...
+	cp Dockerfile $$DOCKER_CONTEXT && \
+	docker build -t $(DOCKER_IMAGE_TAG) $$DOCKER_CONTEXT && \
+	docker run -v $$PWD:/lantern -t $(DOCKER_IMAGE_TAG) /bin/bash -c 'cd /lantern && VERSION="'$$VERSION'" HEADLESS="'$$HEADLESS'" SECRETS_DIR="'$$SECRETS_DIR'" BNS_CERT_PASS="'$$BNS_CERT_PASS'" make $*';
+
 all: binaries
 android-dist: genconfig android
 
-# This is to be called within the docker image.
-docker-genassets: require-npm
+$(RESOURCES_DOT_GO): $(NPM)
```
oxtoacart: I made the assets a file-based rule so that the assets get built only when needed.  To force a refresh of the assets, we can use the new `clean-assets` rule.

Makefile
```diff
@@ -214,65 +219,61 @@ require-gh-token:
 	@if [[ -z "$$GH_TOKEN" ]]; then echo "GH_TOKEN environment value is required."; exit 1; fi
 
 require-secrets-dir:
-	if [[ -z "$$SECRETS_DIR" ]]; then echo "SECRETS_DIR environment value is required."; exit 1; fi
+	@if [[ -z "$$SECRETS_DIR" ]]; then echo "SECRETS_DIR environment value is required."; exit 1; fi
 
 require-secrets: require-secrets-dir
 	@if [[ -z "$$BNS_CERT_PASS" ]]; then echo "BNS_CERT_PASS environment value is required."; exit 1; fi
 
+$(BNS_CERT): require-secrets-dir
```
oxtoacart: I made this a file-based rule as well so that we can use it as a dependency that only fires if the file hasn't been copied.  Similarly to assets, this can be cleaned with `clean-bns-cert`.

Makefile
```diff
@@ -214,65 +219,61 @@ require-gh-token:
...
+	@cp -f $$SECRETS_DIR/$(BNS_CERT) $(BNS_CERT)
+
 require-lantern-binaries:
 	@if [[ ! -d "$(LANTERN_BINARIES_PATH)" ]]; then \
 		echo "Missing lantern binaries repository (https://github.com/getlantern/lantern-binaries). Set it with LANTERN_BINARIES_PATH=\"/path/to/repository\" make ..." && \
 		exit 1; \
 	fi
 
-docker-package-debian-386: require-version docker-linux-386
```
oxtoacart: These were with "debian" but when used as dependencies were named with "linux".  I renamed them using "linux" for consistency.

Makefile
```diff
@@ -418,9 +369,13 @@ package-darwin: package-darwin-manoto
 		echo "-> Skipped: Can not generate a package on a non-OSX host."; \
 	fi;
 
-binaries: docker genassets linux windows darwin
+binaries: docker-assets docker-linux docker-windows darwin
+
+packages: require-version require-secrets clean-desktop docker-assets docker-package-windows docker-package-linux package-darwin
```
oxtoacart: Few things here:

1. We don't need to clean everything, just the desktop related stuff, so I have a new rule `clean-desktop`
2. We don't need to independently build the binaries since the `package-*` targets take care of that
3. Note how we explicitly use docker for everything except package-darwin.  This gives consistent behavior with how this used to work, namely that everything except darwin builds in docker.

oxtoacart:
@xiam @myleshorton This is ready for review now.

Makefile
```diff
@@ -214,65 +220,61 @@ require-gh-token:
...
+	echo "-> lantern-installer.exe and lantern-installer-manoto.exe"
 
-windows: genassets windows-386
+# We overwrite the implicit docker-package-windows rule to include a dependency
+# on $(BNS_CERT)
+docker-package-windows: $(BNS_CERT)
+	make docker-do-package-windows
 
-darwin: genassets darwin-amd64
+linux: linux-386 linux-amd64 linux-arm
```
oxtoacart: Note that I added linux-arm to here.

Makefile
```diff
@@ -353,30 +328,7 @@ lantern: require-assets
...
-package-linux: require-version package-linux-386 package-linux-amd64
-
-package-windows: require-version windows
-	@echo "Generating distribution package for windows/386..." && \
-	if [[ -z "$$SECRETS_DIR" ]]; then echo "SECRETS_DIR environment value is required."; exit 1; fi && \
-	if [[ -z "$$BNS_CERT_PASS" ]]; then echo "BNS_CERT_PASS environment value is required."; exit 1; fi && \
-	$(call docker-up) && \
-	docker run -v $$PWD:/lantern -v $$SECRETS_DIR:/secrets -t $(DOCKER_IMAGE_TAG) /bin/bash -c 'cd /lantern && BNS_CERT="/secrets/bns.pfx" BNS_CERT_PASS="'$$BNS_CERT_PASS'" VERSION="'$$VERSION'" make docker-package-windows' && \
-	echo "-> lantern-installer.exe and lantern-installer-manoto.exe"
+package-linux: require-version package-linux-386 package-linux-amd64 package-linux-arm
```
oxtoacart: Note that this now includes linux-arm as well.

src/github.com/getlantern/flashlight/ui/resources.go
```diff
@@ -1,7 +0,0 @@
-// This file will be overwritten by make genassets.
```
oxtoacart: Instead of versioning resources.go, we just generate it as part of the build.  `make clean` removes it, which allows us to treat it like a typical file-based rule.

Makefile
myleshorton: so this is basically back to the manual assets generation @oxtoacart right? i.e. whenever someone changes the ui, they need to run this? i don't dislike that necessarily, but just making sure.

oxtoacart: Correct.  This was true with the original build file too - people had to run `update-dist` manually.  Now they run `clean-assets'.

Makefile
```diff
@@ -604,19 +560,16 @@ android-release: require-version require-secrets-dir $(LANTERN_MOBILE_ANDROID_RE
...
-	rm -f $(LANTERN_MOBILE_ANDROID_LIB) && \
-	rm -f $(LANTERN_MOBILE_ANDROID_SDK) && \
-	rm -f $(LANTERN_MOBILE_ANDROID_DEBUG) && \
-	rm -f $(LANTERN_MOBILE_ANDROID_RELEASE)
+clean-assets:
+	rm -f $(RESOURCES_DOT_GO)
 
-clean: clean-mobile
+# Provided for backward compatibility with how people used to use the makefile
+update-dist: clean-assets assets
```
oxtoacart: @myleshorton For backward compatibility, I've added this update-dist.  From the user's perspective, it works just as before.

Makefile
myleshorton: Right but you changed that on the current valencia to just run every time if I'm not mistaken @oxtoacart no? I don't honestly have a huge preference either way.

oxtoacart: Hehe, I never actually included that change because you and @xiam convinced me that it was best to handle makefile changes separately.  The important thing is that the packages target runs clean-desktop, which cleans this and forces the assets to be regenerated, so for official builds we're good either way.

myleshorton: Ha nice! OK I'm cool with all this. Any objections @xiam?

xiam: This sounds good to me as well.

Makefile
```diff
@@ -214,65 +220,61 @@ require-gh-token:
...
-docker-package-debian-386: require-version docker-linux-386
+package-linux-386: require-version linux-386
 	@cp lantern_linux_386 lantern_linux_i386;
 	@$(call fpm-debian-build,"i386")
 	@rm lantern_linux_i386;
 	@mv update_linux_i386.bz2 update_linux_386.bz2 && \
 	echo "-> lantern_$(VERSION)_i386.deb"
 
-docker-package-debian-amd64: require-version docker-linux-amd64
+package-linux-amd64: require-version linux-amd64
```
xiam: This target rename is probably not required as it still generates a debian-only package.

Makefile
```diff
@@ -214,65 +220,61 @@ require-gh-token:
...
+package-linux-arm: require-version linux-arm
 	@$(call fpm-debian-build,"arm")
 	@echo "-> lantern_$(VERSION)_arm.deb"
 
-docker-package-windows: require-version docker-windows-386
-	@if [[ -z "$$BNS_CERT" ]]; then echo "BNS_CERT environment value is required."; exit 1; fi && \
-	if [[ -z "$$BNS_CERT_PASS" ]]; then echo "BNS_CERT_PASS environment value is required."; exit 1; fi && \
+package-windows: $(BNS_CERT) do-package-windows
+
+do-package-windows: require-version windows
```
xiam: This target requires `BNS_CERT_PASS ` which does not have any effect without `BNS_CERT`, it looks a lot like `package-windows`, would it be possible to remove the `do-` and only have a `package-windows`?

Makefile
xiam: We don't currently distribute the arm version anywhere, it's kind of experimental, not sure if we should include it here as part of the linux binaries.

Makefile
```diff
@@ -35,6 +35,9 @@ PACKAGE_URL := https://www.getlantern.org
 PACKAGED_YAML := .packaged-lantern.yaml
 MANOTO_YAML := .packaged-lantern-manoto.yaml
 
+RESOURCES_DOT_GO := ./src/github.com/getlantern/flashlight/ui/resources.go
+BNS_CERT := bns.pfx
```
xiam: My concern with having this file here would be the risk of adding it to the repo [by accident](https://github.com/getlantern/gost/issues/2), while this makes signing a bit easier it's probably not so bad to keep the actual `bns.pfx` outside this path if that lowers the risk of adding it by mistake.

Makefile
xiam: Awesome!

Makefile
```diff
@@ -214,65 +220,61 @@ require-gh-token:
...
 
 require-secrets-dir:
-	if [[ -z "$$SECRETS_DIR" ]]; then echo "SECRETS_DIR environment value is required."; exit 1; fi
+	@if [[ -z "$$SECRETS_DIR" ]]; then echo "SECRETS_DIR environment value is required."; exit 1; fi
 
 require-secrets: require-secrets-dir
 	@if [[ -z "$$BNS_CERT_PASS" ]]; then echo "BNS_CERT_PASS environment value is required."; exit 1; fi
 
+$(BNS_CERT): require-secrets-dir
+	@cp -f $$SECRETS_DIR/$(BNS_CERT) $(BNS_CERT)
```
xiam: I think we should not do this automatically, this is not essential, or at least we could protect it with `.gitignore`.

src/github.com/getlantern/flashlight/ui/resources.go
xiam: The side effect of this is not being able to compile flashlight like this `source setenv.bash && go build github.com/getlantern/flashlight/main` without running genassets before. Perhaps we can find another way to check if assets have been generated instead of relying on the existence of this file? The `src/github.com/getlantern/lantern-ui/dist` directory should not exists if assets have not been generated, for instance.

Makefile
oxtoacart: Okay, I'll take it out.

src/github.com/getlantern/flashlight/ui/resources.go
oxtoacart: Yeah, @myleshorton mentioned this too.  I don't understand the benefit to building with go directly instead of using make, and there's a clear downside to how we used to do it.  The build script doesn't push an updated resources.go, so anyone who builds the way that you suggest will be running with an outdated UI.  For example, the resources.go that's on GitHub hasn't been updated since September of last year.  I think it's best for us to run with the most recent code while we develop so that we can catch any errors that we might be introducing, so using make this way seems helpful.

Makefile
oxtoacart: I'm just using the name of the existing package-linux-amd64 target, which is used as a dependency in various places.  What I did with all of the old docker and non-docker targets was to just collapse them and keep the name of the non-docker target.

Makefile
oxtoacart: No, because package-windows can't be run inside docker because the $(BNS_CERT) target doesn't work there (since the $SECRETS_DIR is not accessible from docker).

Makefile
oxtoacart: I agree, let me think about how to solve this.

Makefile
oxtoacart: Per your previous comment, I'm going to figure out a way to not even put the file in this directory at all.

Makefile
xiam: Could we mount it with `-v /full/path/to/local/file.txt:/path/to/file.txt`?

oxtoacart:
@xiam I think I've addressed all of your comments.

myleshorton:
OK LGTM @oxtoacart!

PR #3663, opened 2016-02-25
Issue3296 noioloop

oxtoacart:
@fffw @myleshorton Please don't pull this!  It's just a sketch of a different way to attack the detour refactoring.  Basically, instead of using in ioloop, it uses buffered connections that dial lazily.

I think the main things that would be left to do at this point is to:
1. Rename whitelist to blacklist, since I think that's more accurate
2. Implement a whitelist of direct sites that are known to work so that we don't bother loading a parallel detour connection every time
3. Implement a mechanism for periodically rechecking blacklisted direct sites in case they were blacklisted due to a transient network issue.  We could use a capped exponential backoff to control how frequently we recheck.


src/github.com/getlantern/detour/detour.go
```diff
@@ -1,400 +1,162 @@
...
-					log.Tracef("Detour connection to %s failed, removing from whitelist", dc.addr)
-					RemoveFromWl(dc.addr)
-					// no more connections, return directly to avoid dead lock
-					return n, err
-				}
+			log.Tracef("Dialing detour for %v", addr)
+			conn, err := dialDetour(network, addr)
+			if err == nil {
+				// Prefetch up to 1 MB
+				conn = newEagerconn(conn, 1024768)
```
oxtoacart: @myleshorton @fffw This is an optimization that causes the detour connection to load up to 1 MB of response before we've even failed over to detour.  That way, when we decide that the direct connection is no good, we've already done a lot of the detour reading and don't need to wait for that.

myleshorton:
Have you had a chance to look at these changes @fffw? I know I have not!!

fffw:
I had a glance on it but didn't actually look into the details. Now it's the time!

src/github.com/getlantern/detour/detour.go
```diff
@@ -1,400 +1,162 @@
...
-const (
-	connTypeDirect connType = iota
-	connTypeDetour connType = iota
+	// DirectDialTimeout controls the dial timeout for direct connections
+	// We make this relatively small because detour won't fail over to a proxied
+	// connection until this timeout is hit. The detoured connection is eagerly
+	// fetched, so once we fail over the request will complete very quickly, so
+	// 5 seconds is a good approximation of our maximum latency for initial
+	// detouring.
+	DirectDialTimeout = 5 * time.Second
```
fffw: We probably don't want to introduce timeout, but use whichever connection responds first. If detour connection responds in 1s, waiting for another 4s is unnecessary, but in extremely slow network (GPRS?), it's possible that a direct accessible site takes 10s to respond (and detour connection may take longer). That's the main reason we want to improve detour.

The idea of buffered connection is very inspirational. Having always available connections should eliminate most work to coordinate `Dial`, `Read` and `Write`.

lx3001:
不好

pengyuan:
请蓝灯客服留下联系方式。

我去年买了三年的蓝灯专业版，有不限制的高速流量，也有提示到期日期为2021年。
结果最新版本要求我升级到5.2，而且没有了到期日期，每次连接说我高速流量已用完，要求升级隐币。

以后还有谁敢买专业版，我之前的费用到哪里去了？？？
求解释！！！！！！

Issue #3685, opened 2016-03-01
Improve flashlight unit test coverage


oxtoacart:
Many of the packages used by flashlight have some unit test coverage, but a lot of the code inside flashlight itself isn't covered with unit tests right now.  Would be good to improve on that.

@myleshorton Let me know what you think is the priority on this.

myleshorton:
Ha! Yeah I've been thinking about that too and just noticed you opened a ticket on it! Yeah things like main.go and config.go have really low coverage but a high line count. I think it would be really nice to improve that.

PR #3686, opened 2016-03-01
Directing non-allowed CONNECT ports directly to destination closes #3684

### LINKED ISSUE #3684: For CONNECT traffic on unsupported ports, connect directly to the destination
See getlantern/http-proxy-lantern#35 and #2351.

Basically, our server-side proxy will reject outbound CONNECT requests on ports other than 80, 443, 8080, and 8443.  In these cases, we shouldn't even try to proxy the traffic but instead dial out to the destination directly in order to give the user agent a seamless experience.

If the user is trying to access something on a different port and that something is blocked, we can't help with that right now.

oxtoacart:


src/github.com/getlantern/flashlight/client/config.go
```diff
@@ -13,12 +13,13 @@ var (
...
-	FrontedServers []*FrontedServerInfo
-	ChainedServers map[string]*ChainedServerInfo
-	MasqueradeSets map[string][]*fronted.Masquerade
+	MinQOS              int
+	DumpHeaders         bool // whether or not to dump headers of requests and responses
+	DeviceID            string
+	FrontedServers      []*FrontedServerInfo
+	ChainedServers      map[string]*ChainedServerInfo
+	MasqueradeSets      map[string][]*fronted.Masquerade
+	ProxiedCONNECTPorts []int
```
oxtoacart: I made these ports part of the configuration so that we can quickly change them if we decide to add some.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -78,20 +88,33 @@ func (client *Client) intercept(resp http.ResponseWriter, req *http.Request) {
...
-		// to the chained server. We need to send that request from chained/dialer.go
-		// though because only it knows about the authentication token to use.
-		// We signal it to send the CONNECT here using the network transport argument
-		// that is effectively always "tcp" in the end, but we look for this
-		// special "transport" in the dialer and send a CONNECT request in that
-		// case.
-		return client.getBalancer().Dial("connect", addr)
-	})
-
-	connOut, err = d("tcp", addr)
```
oxtoacart: The heart of the change is below.

src/github.com/getlantern/flashlight/main/settings_test.go
```diff
@@ -11,6 +11,6 @@ func TestNotPersistVersion(t *testing.T) {
 	version := "version-not-on-disk"
 	revisionDate := "1970-1-1"
 	buildDate := "1970-1-1"
-	Load(version, revisionDate, buildDate)
+	LoadSettings(version, revisionDate, buildDate)
```
oxtoacart: This was a bonus.  I noticed that this test was failing because it was using the wrong name for this method.

oxtoacart:
cc: @myleshorton

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -78,20 +89,33 @@ func (client *Client) intercept(resp http.ResponseWriter, req *http.Request) {
...
+	sendToProxy := false
+	for _, proxiedPort := range client.ProxiedCONNECTPorts {
+		if port == proxiedPort {
+			sendToProxy = true
+			break
+		}
+	}
```
myleshorton: My one thought here @oxtoacart is testability. This class is one of our worst offenders in terms of test coverage (https://coveralls.io/builds/5259682), partly because handler_test.go I don't think actually tests anything in handler.go! A simple step would be to make this a separate sendToProxy(req *http.Request) method, something like that, which would be very easy to also write a test for. If you wanted to take that further, making handler_test.go really hit ServeHTTP would be pretty sweet, but we could also leave that for now.

oxtoacart: I agree that we need better coverage, and that's a good idea.  Do you mind if I handle this as part of #3685 so I can approach it holistically?

myleshorton: Yeah that's fine @oxtoacart!

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -437,6 +437,21 @@ func (cfg *Config) applyClientDefaults() {
...
+			// Standard HTTP(S) ports
+			80, 443,
+			// Common unprivileged HTTP(S) ports
+			8080, 8443,
+			// XMPP
+			5222, 5223, 5224,
+			// Android
+			5228, 5229,
+			// Google Hangouts (see https://support.google.com/a/answer/1279090?hl=en)
+			19302, 19303, 19304, 19305, 19306, 19307, 19308, 19309,
```
oxtoacart: Interesting question.  I think your description of what should happen is correct.  I don't know the performance characteristics of udpgw.

myleshorton: I partly asked because I think ports 19302, 19303, and 19304 are UDP-only. I think we want to include them here for the above reason, but I just want to synchronize our mental maps of how all this works!

oxtoacart: Oh, good point.  For the UDP-only ports, we don't actually care about the port number because what we're proxying is actually the udpgw connection, which is on a different port.  In theory, it shouldn't be hard to add udp support directly to Lantern, we mostly just need to stop hardcoding the "tcp" protocol everywhere and add a custom header to our CONNECT request that indicates the protocol :)

oxtoacart: Actually, take that back.  I bet it's not that easy, with UDP being connectionless.

fffw: IIRC Lantern mobile wraps all UDP traffic and forwards to udpgw waited at `127.0.0.1:7300` on server, so we should also add that port right? That will leave UDP traffic unlimited though, but we can add more restriction later.

myleshorton:
OK LGTM @oxtoacart!

PR #3693, opened 2016-03-02
Reconfigure client when proxyAll option changes fixes #3689

### LINKED ISSUE #3689: [2.1.0] Toggling "Proxy all traffic" option has no effect before restart Lantern

fffw:
Most of the code changes is to use map in settings.go, which makes it easier to add notifiers and entries in the future. We need them because that's how main.go get notified when proxyAll option changes.

The code has plenty of room to improve, but we may leave that for future refactor and fix https://github.com/getlantern/lantern/issues/3689 first.

@oxtoacart want take a look?

### DIFF
--- a/src/github.com/getlantern/flashlight/client/client.go
+++ b/src/github.com/getlantern/flashlight/client/client.go
@@ -43,7 +43,7 @@ type Client struct {
 	WriteTimeout time.Duration
 
 	// ProxyAll: (optional) proxy all sites regardless of being blocked or not
-	ProxyAll func() bool
+	ProxyAll bool
 
 	// MinQOS: (optional) the minimum QOS to require from proxies.
 	MinQOS int
@@ -154,25 +154,30 @@ func (client *Client) ListenAndServeSOCKS5(requestedAddr string) error {
 
 // Configure updates the client's configuration. Configure can be called
 // before or after ListenAndServe, and can be called multiple times.
-func (client *Client) Configure(cfg *ClientConfig, proxyAll func() bool) {
+func (client *Client) Configure(cfg *ClientConfig, proxyAll bool) {
 	client.cfgMutex.Lock()
 	defer client.cfgMutex.Unlock()
 
 	log.Debug("Configure() called")
 
-	if client.priorCfg != nil {
-		if reflect.DeepEqual(client.priorCfg, cfg) {
+	if client.priorCfg == nil {
+		var changed bool
+		if !reflect.DeepEqual(client.priorCfg, cfg) {
+			log.Debugf("Client configuration changed")
+			changed = true
+		} else if client.ProxyAll != proxyAll {
+			log.Debugf("Proxy all traffic flag changed to %v", proxyAll)
+			changed = true
+		}
+		if !changed {
 			log.Debugf("Client configuration unchanged")
 			return
 		}
-		log.Debugf("Client configuration changed")
 	} else {
 		log.Debugf("Client configuration initialized")
 	}
 
-	log.Debugf("Requiring minimum QOS of %d", cfg.MinQOS)
 	client.MinQOS = cfg.MinQOS
-	log.Debugf("Proxy all traffic or not: %v", proxyAll())
 	client.ProxyAll = proxyAll
 	client.DeviceID = cfg.DeviceID
 
@@ -194,7 +199,7 @@ func (client *Client) Stop() error {
 
 func (client *Client) proxiedDialer(orig func(network, addr string) (net.Conn, error)) func(network, addr string) (net.Conn, error) {
 	var proxied func(network, addr string) (net.Conn, error)
-	if client.ProxyAll() {
+	if client.ProxyAll {
 		proxied = orig
 	} else {
 		proxied = detour.Dialer(orig)
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -66,54 +66,63 @@ func Run(httpProxyAddr string,
 	socksProxyAddr string,
 	configDir string,
 	stickyConfig bool,
-	proxyAll func() bool,
+	proxyAll bool,
 	flagsAsMap map[string]interface{},
 	beforeStart func(cfg *config.Config) bool,
 	afterStart func(cfg *config.Config),
 	onConfigUpdate func(cfg *config.Config),
-	onError func(err error)) error {
+	onError func(err error)) (proxyAllUpdater func(bool), err error) {
+
 	displayVersion()
 
 	log.Debug("Initializing configuration")
 	cfg, err := config.Init(PackageVersion, configDir, stickyConfig, flagsAsMap)
 	if err != nil {
-		return fmt.Errorf("Unable to initialize configuration: %v", err)
+		err = fmt.Errorf("Unable to initialize configuration: %v", err)
 	}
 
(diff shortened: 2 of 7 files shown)


src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -43,7 +43,7 @@ type Client struct {
 	WriteTimeout time.Duration
 
 	// ProxyAll: (optional) proxy all sites regardless of being blocked or not
-	ProxyAll func() bool
+	ProxyAll bool
```
fffw: We now reconfigure each time, getter is no long required.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -66,54 +66,63 @@ func Run(httpProxyAddr string,
...
-		log.Debug("Preparing to start client proxy")
-		geolookup.Configure(client.Addr)
-		cfgMutex.Lock()
+	if !beforeStart(cfg) {
+		return
+	}
+
+	proxyAllUpdater = func(proxyAll bool) {
 		applyClientConfig(client, cfg, proxyAll)
-		cfgMutex.Unlock()
```
fffw: The diff here is messed up. Most of the code doesn't change indeed.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -66,54 +66,63 @@ func Run(httpProxyAddr string,
...
 	configDir string,
 	stickyConfig bool,
-	proxyAll func() bool,
+	proxyAll bool,
 	flagsAsMap map[string]interface{},
 	beforeStart func(cfg *config.Config) bool,
 	afterStart func(cfg *config.Config),
 	onConfigUpdate func(cfg *config.Config),
-	onError func(err error)) error {
+	onError func(err error)) (proxyAllUpdater func(bool), err error) {
```
fffw: It is ugly to add this function as return value, but limited by current code structure, it's a quick and dirty way to fix.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -66,54 +66,63 @@ func Run(httpProxyAddr string,
...
-				err = client.ListenAndServeSOCKS5(socksProxyAddr)
-				if err != nil {
-					log.Errorf("Unable to start SOCKS5 proxy: %v", err)
-				}
-			}()
-		}
-
-		log.Debug("Starting client HTTP proxy")
+	log.Debug("Starting client HTTP proxy")
+	go func() {
```
fffw: Wrap in goroutine so we can return immediately

src/github.com/getlantern/flashlight/main/main.go
```diff
@@ -132,6 +132,16 @@ func doMain() error {
 		return err
 	}
 
+	settings = LoadSettings(flashlight.Version, flashlight.RevisionDate, flashlight.BuildDate)
```
fffw: Split to two steps so notifiers added before `Start` can be called.

src/github.com/getlantern/flashlight/main/settings.go
```diff
@@ -2,198 +2,281 @@ package main
...
 )
 
+// Notifier is what caller register to get notified when a settings entry changed
+type Notifier func(prev, cur interface{})
+
+// BoolNotifier is a shortcut of Notifier on bool entries
+type BoolNotifier func(cur bool)
+
 const (
+	ProxyAll    id = "proxyAll"
```
fffw: Use predefined strings as key, but keep the type private so other packages can create keys on the fly. Not useful ATM.

src/github.com/getlantern/flashlight/main/settings.go
```diff
@@ -2,198 +2,281 @@ package main
...
-			log.Errorf("Unable to register settings service: %q", err)
-			return
+	settings.entries[Version] = version
+	settings.entries[BuildDate] = buildDate
+	settings.entries[RevisionDate] = revisionDate
+
+	return settings
+}
+
+func New(entries entries, persist bool) *Settings {
```
fffw: It's not meant to be used outside, but `new` is a keyword, so...

src/github.com/getlantern/flashlight/main/settings.go
```diff
@@ -2,198 +2,281 @@ package main
...
 		}
-		go settings.read()
-	})
+	}
 	return settings
 }
 
 type msg struct {
-	Settings   *Settings
+	Settings   map[string]interface{}
```
fffw: Have to use exactly this type.

src/github.com/getlantern/flashlight/main/main.go
```diff
@@ -140,7 +150,6 @@ func doMain() error {
 		}
 	})
 	addExitFunc(quitSystray)
-	addExitFunc(settings.Save)
```
fffw: Now we save settings whenever any entry changes, no need to save it anymore.

src/github.com/getlantern/flashlight/main/settings_test.go
```diff
@@ -11,6 +12,30 @@ func TestNotPersistVersion(t *testing.T) {
 	version := "version-not-on-disk"
 	revisionDate := "1970-1-1"
 	buildDate := "1970-1-1"
-	Load(version, revisionDate, buildDate)
-	assert.Equal(t, settings.Version, version, "Should be set to version")
+	LoadSettings(version, revisionDate, buildDate)
+	assert.Equal(t, settings.entries[ProxyAll], true, "Should load from file")
```
fffw: `ProxyAll` is false by default. If it's true, we successfully loaded file.

src/github.com/getlantern/flashlight/main/test.yaml
```diff
@@ -1,7 +1,7 @@
 version: test
```
fffw: With the `map` implemntation, the key names become camelCased in YAML. It will invalidate users' existing `settings.yaml`, but it's good to keep the names in same format from YAML file to UI.

oxtoacart:
@fffw I'm not a big fan of notification/listeners when that level of complexity is not needed.  Here is what I think is a simpler solution - https://github.com/getlantern/lantern/pull/3695

fffw:
@oxtoacart Seems your PR is solving a different problem.

The bug this PR trying to resolve is irrelevant to PAC but how Lantern choose dialer for plain HTTP requests: Though [client.newReverseProxy](https://github.com/getlantern/lantern/blob/release-2.1.0/src/github.com/getlantern/flashlight/client/reverseproxy.go#L17) calles [proxiedDialer](https://github.com/getlantern/lantern/blob/release-2.1.0/src/github.com/getlantern/flashlight/client/client.go#L195) which respects `settings.GetProxyAll`, it is only called once in `client.Configure`, so we need to call `client.Configure` each time `proxyAll` flag is toggled, that's why I added listener.

Other changes to `settings.go` seems more useful when we add more options in Pro UI.

oxtoacart:
Oh, interesting.  I think the bug is relevant to both, right?  Because the PAC file also wasn't getting updated when the setting changed.

For the reverse proxy problem, this seems like a much simpler solution - https://github.com/getlantern/lantern/commit/df930589f28f83a39337afc625472287a1b89076

fffw:
Oh I didn't realize that it requires such a simple fix. Will close this PR until we need the flexibility again.

PR #3695, opened 2016-03-02
Dynamically serving PAC file closes #3689

### LINKED ISSUE #3689: [2.1.0] Toggling "Proxy all traffic" option has no effect before restart Lantern

oxtoacart:
Basically the change is to generate the PAC file dynamically based on the current value of ProxyAll.

### DIFF
--- a/src/github.com/getlantern/flashlight/main/main.go
+++ b/src/github.com/getlantern/flashlight/main/main.go
@@ -267,8 +267,7 @@ func afterStart(cfg *config.Config) {
 func onConfigUpdate(cfg *config.Config) {
 	autoupdate.Configure(cfg)
 	proxiedsites.Configure(cfg.ProxiedSites)
-	log.Debugf("Proxy all traffic or not: %v", settings.GetProxyAll())
-	ServeProxyAllPacFile(settings.GetProxyAll())
+	ServeProxyAllPacFile(settings.GetProxyAll)
 }
 
 func i18nInit() {
--- a/src/github.com/getlantern/flashlight/main/main.go
+++ b/src/github.com/getlantern/flashlight/main/main.go
@@ -248,6 +248,7 @@ func beforeStart(cfg *config.Config) bool {
 
 func afterStart(cfg *config.Config) {
 	onConfigUpdate(cfg)
+	ServePACFile()
 	if settings.GetSystemProxy() {
 		pacOn()
 	}
@@ -267,7 +268,6 @@ func afterStart(cfg *config.Config) {
 func onConfigUpdate(cfg *config.Config) {
 	autoupdate.Configure(cfg)
 	proxiedsites.Configure(cfg.ProxiedSites)
-	ServeProxyAllPacFile(settings.GetProxyAll)
 }
 
 func i18nInit() {
--- a/src/github.com/getlantern/flashlight/main/pac.go
+++ b/src/github.com/getlantern/flashlight/main/pac.go
@@ -131,22 +131,27 @@ func watchDirectAddrs() {
 	go func() {
 		for {
 			addr := <-detour.DirectAddrCh
-			// prevents Lantern from accidently leave pac on after exits
-			if atomic.LoadInt32(&isPacOn) == 0 {
-				return
-			}
 			host, _, err := net.SplitHostPort(addr)
 			if err != nil {
 				panic("watchDirectAddrs() got malformated host:port pair")
 			}
-			if !directHosts[host] {
-				directHosts[host] = true
-				// reapply so browser will fetch the PAC URL again
-				doPACOff(pacURL)
-				doPACOn(pacURL)
-			}
+			addDirectHost(host)
 		}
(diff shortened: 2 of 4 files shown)


src/github.com/getlantern/flashlight/main/pac.go
```diff
@@ -118,42 +131,31 @@ func watchDirectAddrs() {
 	go func() {
 		for {
 			addr := <-detour.DirectAddrCh
-			// prevents Lantern from accidently leave pac on after exits
-			if atomic.LoadInt32(&isPacOn) == 0 {
-				return
-			}
```
oxtoacart: @myleshorton @fffw FYI, this is a bug that we introduced when we added the ability to disable Lantern as system proxy.  In that case, isPacOn will be false, so we ignore the new direct host.  This is an edge case and unlikely to happen, but I went ahead and fixed it while I was in here.  Specifically, I now check isPacOn only to determine whether to force the browser to reload the PAC by turning it off and then on.  We always apply the direct host to the map of direct hosts.

src/github.com/getlantern/flashlight/main/main.go
```diff
@@ -248,6 +248,7 @@ func beforeStart(cfg *config.Config) bool {
 
 func afterStart(cfg *config.Config) {
 	onConfigUpdate(cfg)
+	ServePACFile()
```
oxtoacart: We only need to do this once.

src/github.com/getlantern/flashlight/main/pac.go
```diff
@@ -20,22 +21,32 @@ import (
...
+
+func servePACFileIfNecessary() {
+	if pacURL == "" {
+		handler := func(resp http.ResponseWriter, req *http.Request) {
+			log.Trace("Serving PAC file")
+			resp.Header().Set("Content-Type", "application/x-ns-proxy-autoconfig")
+			resp.WriteHeader(http.StatusOK)
+			cfgMutex.RLock()
+			defer cfgMutex.RUnlock()
+			if _, err := genPACFile(resp); err != nil {
```
oxtoacart: Yes, this means we're generating the PAC file dynamically every time.  No it's not a performance problem, because the browser infrequently requests the PAC file.  Also, a boom test reveals that we can serve the dynamic file at about 20,000 per second.

src/github.com/getlantern/flashlight/main/pac.go
```diff
@@ -20,22 +21,29 @@ import (
...
-		atomic.StoreInt32(&shouldProxyAll, 0)
+func ServePACFile() {
+	cfgMutex.Lock()
+	defer cfgMutex.Unlock()
+	if pacURL == "" {
+		pacURL = ui.Handle("/proxy_on.pac", http.HandlerFunc(sercePACFile))
+	}
+}
+
+func sercePACFile(resp http.ResponseWriter, req *http.Request) {
```
myleshorton: Looking good just a spelling issue with sercePACFile

src/github.com/getlantern/flashlight/main/pac.go
```diff
@@ -20,22 +21,29 @@ import (
...
+	log.Trace("Serving PAC file")
+	resp.Header().Set("Content-Type", "application/x-ns-proxy-autoconfig")
+	resp.WriteHeader(http.StatusOK)
```
myleshorton: It strikes me we should be adding content-length her, but maybe the go HTTP internals do that automatically?

src/github.com/getlantern/flashlight/main/pac.go
myleshorton: Nice catch!

src/github.com/getlantern/flashlight/main/pac.go
oxtoacart: Fixed.

src/github.com/getlantern/flashlight/main/pac.go
oxtoacart: For HTTP/1.1, Go automatically uses chunked transfer encoding if no content-length header is specified.  That begs the question if any user agents request the PAC file using HTTP/1.0.  Incidentally, the previous version of this code also did not include the content-length header for the PAC file.

myleshorton:
This generally looks good @oxtoacart, although testing it revealed another issue, which is that `bypassDomains` in the pac file is never populated, either other this branch or on valencia! So something else broke there @fffw. I'll open a separate bug.

oxtoacart:
Interesting @myleshorton.  I do see sites get added to bypassDomains as I access them (presumably by detour).  Are you saying that the list should be pre-populated with some more site?

myleshorton:
Hey @oxtoacart sorry I'm actually seeing that now too -- for some reason it was taking them awhile to populate in my initial testing, but now I'm seeing it working on all branches.

myleshorton:
This is working well -- I'll merge as soon as Travis is happy.

PR #3726, opened 2016-03-05
Treat all net.OpError as being blocked

### LINKED ISSUE #3609: Fail to load Fastly sites sometimes
User reported that sites using fastly.net can be failed to load via Lantern in a period but can be accessed through VPN at the same time. From the logs, Lantern always connected to the site, but browser shows "Failed to load resource" or "net::ERR_CONNECTION_RESET". One such example is `travis-ci-org.global.ssl.fastly.net`.

fffw:
Resolves #3609.

We were too rigid on detecting the symptom of being blocked. In this specific case, due to recent changes of Go library code, `Read` returns `read: connection reset by peer` error instead of previous one `ECONNRESET ` (`connection reset by peer`), which makes `detour` failed to detect. This bug should affect lots of sites, so it probably worth adding to this release.

@myleshorton How do you think?

### DIFF
--- a/src/github.com/getlantern/detour/detect.go
+++ b/src/github.com/getlantern/detour/detect.go
@@ -3,7 +3,6 @@ package detour
 import (
 	"bytes"
 	"net"
-	"syscall"
 )
 
 // Detector is just a set of rules to check if a site is potentially blocked or not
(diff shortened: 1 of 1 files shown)


myleshorton:
Sounds good @fffw great catch! I'll review and add.

myleshorton:
LGTM @fffw. I'll add it to 2.1.0

PR #3740, opened 2016-03-07
Checking for nil connection in detour before closing

### LINKED ISSUE #3477: Lantern unable to proxy when using proxy-all-traffic after getting a SSL connection error for that particular site which is not in the default list - Windows 7
Steps:

1. Run lantern. 
2. Make sure proxy-all-traffic is unchecked. Go to a blocked site which is not in the default list(I used slack.com with fake DNS blocking). SSL connection error occurs.
3. Now, check proxy-all-traffic. The change doesn't impact on slack even if I refresh after few minutes of waiting time.
4. Open a new instance of Lantern by using Show Lantern. I couldn't open slack then too. But I was able to open other blocked sites which is not in the default list. I used chrome till here. I tried the site in Firefox and it was able to open that site without any problem. Looks like it is specific to the browser too.
5. After quitting lantern and running it again in chrome made it work.

After quitting, it looks like the new instance is able to identify it as a blocked site and starts proxying but the former didn't recognize..

oxtoacart:
This doesn't necessarily resolve #3477, but it fixes a panic that I encountered while testing it.

### DIFF
--- a/src/github.com/getlantern/detour/detour.go
+++ b/src/github.com/getlantern/detour/detour.go
@@ -309,7 +309,11 @@ func (dc *Conn) Close() error {
 			}
 		}
 	}
-	return dc.getConn().Close()
+	conn := dc.getConn()
+	if conn == nil {
+		return nil
+	}
+	return conn.Close()
 }
 
 // LocalAddr() implements the function from net.Conn


myleshorton:
LGTM!

PR #3747, opened 2016-03-08
Calculating versionCode dynamically based on date, eliminates need fo…

### LINKED ISSUE #3715: Automatically commit mobile version.properties with new releases
src/github.com/getlantern/lantern-mobile/app/version.properties currently gets updated with releases, but not committed and so not incremented across versions.

oxtoacart:
…r version.properties, closes #3715


myleshorton:
Nice @oxtoacart! And the internal update logic follows semantic versioning, is that right?

oxtoacart:
The internal update logic uses the integer versionCode and does plain integer comparisons.  As long as the versionCode is higher than the installed, then it upgrades.

atavism:
Do we want this to only apply when its a debug build? presumably we don't want release builds to use this type of versioning

oxtoacart:
@atavism I intended this to apply to all builds.  Why wouldn't we use it for release builds?

atavism:
The convention for version numbers in the Play store is the 1.x.x format (no more than three numbers typically). I've never seen a date or revision number included. So it just seems weird to me, but I guess it doesn't particularly matter though.

oxtoacart:
The [android docs](http://developer.android.com/tools/publishing/versioning.html) docs seem to leave the meaning of versionName pretty open

> The value is a string so that you can describe the application version as a <major>.<minor>.<point> string, or as any other type of absolute or relative version identifier.

PR #3755, opened 2016-03-09
Balancer always pick fastest dialer resolves #3358

### LINKED ISSUE #3358: When we have multiple servers available, use the fastest one

fffw:
Re-pull to `devel` branch. Resolves https://github.com/getlantern/lantern/issues/3358

### DIFF
--- a/src/github.com/getlantern/balancer/avg_bench_test.go
+++ b/src/github.com/getlantern/balancer/avg_bench_test.go
@@ -1,0 +1,54 @@
+package balancer
+
+/*
+Not a benchmark actually. It's just to give us an sense how different
+calculation of "average" looks like (Imagine the data points are random connect
+time of an 500ms link vs 200ms link). Example result:
+
+	benchmark_avg_test.go:50: d=450	avg=490	ravg1=584	ravg2=621	ravg3=622	||	d=110	avg=199	ravg1=92	ravg2=107	ravg3=119
+	benchmark_avg_test.go:50: d=882	avg=491	ravg1=733	ravg2=708	ravg3=687	||	d=264	avg=199	ravg1=178	ravg2=159	ravg3=155
+	benchmark_avg_test.go:50: d=190	avg=490	ravg1=461	ravg2=535	ravg3=562	||	d=109	avg=199	ravg1=143	ravg2=142	ravg3=143
+	benchmark_avg_test.go:50: d=942	avg=491	ravg1=701	ravg2=670	ravg3=657	||	d=227	avg=199	ravg1=185	ravg2=170	ravg3=164
+	benchmark_avg_test.go:50: d=238	avg=490	ravg1=469	ravg2=526	ravg3=552	||	d=11	avg=199	ravg1=98	ravg2=117	ravg3=125
+	benchmark_avg_test.go:50: d=291	avg=490	ravg1=380	ravg2=447	ravg3=486	||	d=324	avg=199	ravg1=211	ravg2=186	ravg3=174
+	benchmark_avg_test.go:50: d=204	avg=490	ravg1=292	ravg2=366	ravg3=415	||	d=123	avg=199	ravg1=167	ravg2=165	ravg3=161
+	benchmark_avg_test.go:50: d=876	avg=490	ravg1=584	ravg2=536	ravg3=530	||	d=128	avg=199	ravg1=147	ravg2=152	ravg3=152
+	benchmark_avg_test.go:50: d=35	avg=490	ravg1=309	ravg2=369	ravg3=406	||	d=127	avg=199	ravg1=137	ravg2=143	ravg3=145
+	benchmark_avg_test.go:50: d=350	avg=490	ravg1=329	ravg2=362	ravg3=392	||	d=356	avg=199	ravg1=246	ravg2=214	ravg3=197
+*/
+
+/*import (
+	"math/rand"
+	"testing"
+)
+
+func BenchmarkAvg(b *testing.B) {
+	total_A := 0
+	recentAvg_A := 0
+	recent2Avg_A := 0
+	recent3Avg_A := 0
+
+	total_B := 0
+	recentAvg_B := 0
+	recent2Avg_B := 0
+	recent3Avg_B := 0
+	for i := 0; i < b.N; i++ { //use b.N for looping
+		d_A := rand.Intn(1000)
+		total_A = total_A + d_A
+		recentAvg_A = (recentAvg_A + d_A) / 2
+		recent2Avg_A = (2*recent2Avg_A + d_A) / 3
+		recent3Avg_A = (3*recent3Avg_A + d_A) / 4
+
+		d_B := rand.Intn(400)
+		total_B = total_B + d_B
+		recentAvg_B = (recentAvg_B + d_B) / 2
+		recent2Avg_B = (2*recent2Avg_B + d_B) / 3
+		recent3Avg_B = (3*recent3Avg_B + d_B) / 4
+
+		if i > 1000 {
+			b.Logf("d=%d\tavg=%d\travg1=%d\travg2=%d\travg3=%d\t||\td=%d\tavg=%d\travg1=%d\travg2=%d\travg3=%d",
+				d_A, total_A/(i+1), recentAvg_A, recent2Avg_A, recent3Avg_A,
+				d_B, total_B/(i+1), recentAvg_B, recent2Avg_B, recent3Avg_B)
+		}
+	}
+}*/
--- a/src/github.com/getlantern/balancer/dialer.go
+++ b/src/github.com/getlantern/balancer/dialer.go
@@ -118,7 +118,7 @@ func (d *dialer) start() {
 }
 
 func (d *dialer) metrics() metrics {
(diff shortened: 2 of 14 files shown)


fffw:
@oxtoacart I removed `QOS` and `Weight` from balancer as we no longer use them. Tests are also changed a lot to align with strategy based heap operation. By always picking fastest server, the `QualityFirst` strategy has some sort of load balancing if the `avgDialTime` of multiple dialers are at same level but change slightly.

myleshorton:
Exciting @fffw! The one thing that strikes me is just naming -- i.e we're no longer really doing any "balancing" so maybe there's a better name for this whole subsystem?

fffw:
Hmm naming is not what I'm good at. "rotator" or "dispatcher"? Seems pointless...

myleshorton:
How you feeling on this one @oxtoacart?

oxtoacart:
@myleshorton @fffw Apologies about the delay, I'll take a look this evening.

src/github.com/getlantern/balancer/strategy.go
```diff
@@ -0,0 +1,104 @@
...
+		q1 := dialers[i].ConsecSuccesses() - dialers[i].ConsecFailures()
+		q2 := dialers[j].ConsecSuccesses() - dialers[j].ConsecFailures()
+		return q1 > q2
+	}}
+}
+
+// Fastest strategy always pick the dialer with lowest recent average connect time
+func Fastest(dialers []*dialer) dialerHeap {
+	return dialerHeap{dialers: dialers, lessFunc: func(i, j int) bool {
+		return dialers[i].AvgDialTime() < dialers[j].AvgDialTime()
```
oxtoacart: One risk with this is that if we're not setting dial timeouts, one unlucky dial can really skew the statistics (for example if the network happens to be down when we try one dialer).  After the one really long dial, it will take a while for the average to recover.

Two potential solutions to this are:

1. Use a moving average over the last X period
2. Use the median instead of the average 

Median can actually be accomplished memory efficiently using a [sketching algorithm](http://research.neustar.biz/2013/09/16/sketch-of-the-day-frugal-streaming/).

This will not prevent me from approving this PR, it's just something I noticed that we could improve at some point if we want.

src/github.com/getlantern/balancer/dialer.go
```diff
@@ -94,44 +78,94 @@ func (d *dialer) start() {
...
-					lastCheckSucceeded = time.Now()
-					timer.Reset(longDuration)
-					consecCheckFailures = 0
-				} else {
-					consecCheckFailures += 1
-				}
+			case <-d.errCh:
+				d.markFailure()
+			case <-d.checkTimer.C:
+				go func() {
```
oxtoacart: If checking takes longer than maxCheckTimeout, we will start piling up goroutines/connections.  Previously, checks happened on only one goroutine.  I think we should continue doing checks on just one goroutine.

src/github.com/getlantern/balancer/dialer.go
```diff
@@ -94,44 +78,94 @@ func (d *dialer) start() {
...
+	return atomic.LoadInt64(&d.avgDialTime)
+}
+func (d *dialer) ConsecSuccesses() int32 {
+	return atomic.LoadInt32(&d.consecSuccesses)
+}
+func (d *dialer) ConsecFailures() int32 {
+	return atomic.LoadInt32(&d.consecFailures)
+}
+
+func (d *dialer) CheckedDial(network, addr string) (net.Conn, error) {
```
oxtoacart: Now that we have CheckedDial, is Dial even used by anyone?  If not, let's just have  Dial method that does it all.

src/github.com/getlantern/balancer/dialer.go
oxtoacart: Also, now that we don't actually take dialers out of rotation anymore, what's the purpose of doing these checks?

oxtoacart:
@myleshorton You asked me to check something in relation to this at our standup.  I thought I understood your request, but I didn't.  What did you want me to check?

oxtoacart:
@fffw Generally looks good, nice and clean.  The only largish concern I have is around the dialer checking logic.  A smaller question I have is around use of average connect time vs median or something like that.  I've also submitted a small PR with some documentation/logging enhancements - #3780.

myleshorton:
@oxtoacart I was wondering if this PR fetches the pro config using the extra headers for pro.

fffw:
@myleshorton The X-Lantern-User-Id and X-Lantern-Pro-Token header should be absent for an not registered client. It's also not the responsibility of balancer to provide these headers, as they are same across servers assigned to specific user. IIRC the pro client code should have already persisted them to `settings.yaml` and use them.

myleshorton:
Cool thanks @fffw! Yeah I was wondering if you had added any of that header handling in config.go. I can easily do that.

fffw:
Oh sorry @myleshorton I misunderstood your previous comment. Seems changes to config.go should still be in a pro branch as we don't even have those fields supplied in `devel` branch.

It raises another issue that proxy should not remove relevant headers. Logged at https://github.com/getlantern/http-proxy-lantern/issues/44

src/github.com/getlantern/balancer/strategy.go
fffw: It's indeed a moving average in the form of `(current T + previous AvgT) / 2` though the name doesn't reflect it. If it's not smooth enough in practice we can change it to something like `(current T + (previous AvgT)*9) / 10`. I compared different calculation in `avg_bench_test.go`. Will change the name.

src/github.com/getlantern/balancer/dialer.go
fffw: The purpose is to calculate metrics even if the dialer is not picked recently by the strategy. Check in same goroutine will block the loop, maybe spawn a dedicated goroutine? Will try.

src/github.com/getlantern/balancer/dialer.go
fffw: `Dial` is a field supplied by caller.

src/github.com/getlantern/balancer/strategy.go
oxtoacart: Ahh, cool, I missed that!  Looks good.

src/github.com/getlantern/balancer/dialer.go
oxtoacart: Yeah, a dedicated goroutine sounds good.

src/github.com/getlantern/balancer/dialer.go
oxtoacart: Oh, I see.  It might be less confusing to change CheckedDial -> Dial and Dial -> DoDial or UnderlyingDial or something like that.  Dial is the typical name used by things like net.Dialer, so someone looking at this is more likely to use Dial than CheckedDial since they probably don't know what CheckedDial means.

fffw: Renamed `Dial` to `DialFN` and `CheckedDial` to `dial` (we don't call it outside this package anymore)

src/github.com/getlantern/balancer/dialer.go
```diff
@@ -94,44 +70,83 @@ func (d *dialer) start() {
 					d.OnClose()
 				}
 				return
-			case t := <-d.errCh:
```
fffw: I removed `errCh` and `markFailure` directly to avoid blocking caller, so goroutine is no longer required.

src/github.com/getlantern/balancer/dialer.go
```diff
@@ -94,44 +70,83 @@ func (d *dialer) start() {
...
-func (d *dialer) isActive() bool {
-	return atomic.LoadInt32(&d.active) == 1
+func (d *dialer) Stop() {
+	d.closeCh <- struct{}{}
+}
+
+// It's the Exponential moving average of dial time with an α of 0.5.
+// Ref https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average.
+// If it's not smooth enough, we can increase α by changing `updateEMADialTime`.
+func (d *dialer) EMADialTime() int64 {
```
fffw: Renamed `AvgDialTime` to `EMADialTime` to avoid confusion.

fffw:
@oxtoacart updated per your comments.

src/github.com/getlantern/balancer/avg_bench_test.go
```diff
@@ -0,0 +1,54 @@
+package balancer
```
oxtoacart: Why is this file commented out?

oxtoacart:
@fffw Your changes look good.  Just one remaining question about the commented-out avg_bench_test, and then we can pull.

fffw:
It's used as a comparision of different EWA, nothing is used in production,
so I comment it in case we need it in the future.

oxtoacart:
LGTM!

PR #3759, opened 2016-03-09
Moved server build instructions here

myleshorton:


myleshorton:
This looks ok @oxtoacart?

oxtoacart:
@myleshorton Looks good.  If the user is running the build on linux, they can actually just use `make linux` instead of `make docker-linux`.

myleshorton:
Oh good point @oxtoacart -- I'll add that.

myleshorton:
How's it look now @oxtoacart?

myleshorton:
Thanks @oxtoacart!

Issue #3769, opened 2016-03-11
Experimenting testdroid


fffw:
Manual / automatic test to make sure it's at least up and running on target devices, esp. those without Google Play service.

PR #3776, opened 2016-03-14
run Android tests locally and on testdroid cloud

fffw:
Added python scripts to run blackbox test. Currently only the "Quit" menu is tested. The value of `TESTDROID_APIKEY` is in `too-many-secrets/testdroid.com`.

@atavism You probably will find this useful.

Resolves https://github.com/getlantern/lantern/issues/3769

### DIFF
--- a/src/github.com/getlantern/lantern-mobile/test/lib/suite.py
+++ b/src/github.com/getlantern/lantern-mobile/test/lib/suite.py
@@ -37,7 +37,7 @@ class TestLantern(ParametrizedTestCase):
     def setUp(self):
         log("Starting Appium test using device '%s'" % self.device_name)
         tm = time.strftime("%Y%m%dT%H%M%S", time.gmtime())
-        self.screenshotDir = os.path.join('./screenshots', self.device_name+'_'+tm)
+        self.screenshotDir = os.path.join('./screenshots', tm + '_' + self.device_name)
         log("Will save screenshots at: " + self.screenshotDir)
         try:
             os.makedirs(self.screenshotDir)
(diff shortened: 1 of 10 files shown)


src/github.com/getlantern/lantern-mobile/test/lib/suite.py
```diff
@@ -0,0 +1,121 @@
...
+
+        log("WebDriver request initiated. Waiting for response, this typically takes 2-3 mins")
+        self.driver = webdriver.Remote(self.hub_url, self.desired_caps)
+        log("WebDriver response received")
+
+    def tearDown(self):
+        log("Quitting")
+        self.driver.quit()
+
+    def testStartAndQuit(self):
```
fffw: Here is the place to add more tests.

src/github.com/getlantern/lantern-mobile/test/lib/suite.py
```diff
@@ -0,0 +1,121 @@
...
+        log("Test: start and quit app")
+        self.driver.save_screenshot(self.screenshotDir + "/1_appLaunch.png")
+
+        # self.by_id('settings_icon').click() # it doesn't work on API 17 or below!
```
fffw: It's still not quite clear to me the various ways to locate an element on an Android app. @atavism You probably has more knowledge on that. Current test should work on API 17 (4.2.2) and above.

myleshorton:
Very cool @fffw. Can you give a more detailed overview of what this change does?

src/github.com/getlantern/lantern-mobile/test/lib/suite.py
atavism: These tests look awesome, @fffw! Elements are defined in the XML files in the ``layout`` resources directory. It would be cool if we could add a test here to toggle the on/off switch (the ToggleButton is defined in the ``activity_lantern_main.xml`` file).

myleshorton:
Can you document this such that @sunshine19090 can also run it?

fffw:
Good idea @myleshorton ! For that purpose, it would be better in a separate repo and well documented so anyone can use it to test any APK. Will make that change.

fffw:
Moved to https://github.com/getlantern/test-automation/tree/master/mobile-test

PR #3783, opened 2016-03-15
Issue 3745

oxtoacart:
Adds support for android push notifications.


pubsub-java/README.md
```diff
@@ -0,0 +1,6 @@
+# pubsub-java
```
oxtoacart: pubsub-java is a git subtree from https://github.com/getlantern/pubsub-java

PubSub/sdk/src/main/java/org/lantern/pubsub/AutoStart.java
```diff
@@ -0,0 +1,19 @@
+package org.lantern.pubsub;
```
atavism: For the Android app, I'm actually following the Sun convention and using the inverse of our domain for the package name: ``org.getlantern.lantern`` But I like ``org.lantern``, if only because it's shorter; the ``getlantern`` is unnecessary. I think after this is merged we could rename it to be consistent.

PubSub/local.properties
```diff
@@ -0,0 +1,11 @@
...
+# Do not modify this file -- YOUR CHANGES WILL BE ERASED!
+#
+# This file must *NOT* be checked into Version Control Systems,
+# as it contains information specific to your local configuration.
+#
+# Location of the SDK. This is only used by Gradle.
+# For customization when using a Version Control System, please read the
+# header note.
+#Fri Mar 11 12:42:32 CST 2016
+sdk.dir=/opt/adt-bundle-mac-x86_64-20130917/sdk
```
atavism: Noticed there's a gitignore entry for ``local.properties`` but maybe this file was committed by accident?

PubSub/sdk/src/main/java/org/lantern/pubsub/AutoStart.java
myleshorton: Yeah I kind of want to ditch getlantern in general honestly and to move everything to just lantern -- it confuses people constantly!

PubSub/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+    repositories {
+        jcenter()
+    }
+    dependencies {
+        classpath 'com.android.tools.build:gradle:1.5.0'
+
+        // NOTE: Do not place your application dependencies here; they belong
+        // in the individual module build.gradle files
+    }
+}
```
atavism: If an app goes to integrate pubsub, it's simple enough compiling the JAR and copying it to the libs directory, but it would also be really convenient to start publishing this and the Lantern SDK to Sonatype's snapshots repository. We don't have to do it in this PR, of course, but I figured its worth mentioning.

PubSub/sdk/src/main/java/org/lantern/pubsub/AutoStart.java
oxtoacart: So I'm okay to stick with `org.lantern`?

PubSub/local.properties
oxtoacart: Fixed, thanks!

PubSub/build.gradle
oxtoacart: Agreed.  Maybe we if have time in LA we can look at that together?

atavism: Yeah, absolutely!

PubSub/sdk/src/main/java/org/lantern/pubsub/AutoStart.java
myleshorton: Yeah given that it's a separate library outside of our Android app that sounds good to me @oxtoacart and @atavism

myleshorton: Just one less thing to change if we make the jump, which I was actually just thinking about earlier today...

atavism:
@oxtoacart I haven't investigated this further, but one issue I've encountered, integrating pubsub in a manner similar to the LanternMobileTestBed app, starting the service, and leaving it running for a bit:

```
             System.err  W  java.lang.RuntimeException: Unable to send message: Write error: ssl=0xa9611c60: I/O error during system call, Broken pipe
                         W      at org.lantern.pubsub.Client$Sendable.run(Client.java:289)
                         W      at org.lantern.pubsub.Client.doWithConnection(Client.java:185)
                         W      at org.lantern.pubsub.Client.process(Client.java:135)
                         W      at org.lantern.pubsub.Client.run(Client.java:127)
                         W      at java.lang.Thread.run(Thread.java:841)
                         W  Caused by: javax.net.ssl.SSLException: Write error: ssl=0xa9611c60: I/O error during system call, Broken pipe
                         W      at com.android.org.conscrypt.NativeCrypto.SSL_write(Native Method)
                         W      at com.android.org.conscrypt.OpenSSLSocketImpl$SSLOutputStream.write(OpenSSLSocketImpl.java:788)
                         W      at java.io.BufferedOutputStream.flushInternal(BufferedOutputStream.java:185)
                         W      at java.io.BufferedOutputStream.flush(BufferedOutputStream.java:85)
                         W      at org.msgpack.core.buffer.OutputStreamBufferOutput.flush(OutputStreamBufferOutput.java:99)
                         W      at org.msgpack.core.MessagePacker.flush(MessagePacker.java:157)
                         W      at org.lantern.pubsub.Client$Sendable.sendImmediate(Client.java:301)
                         W      at org.lantern.pubsub.Client$Sendable.run(Client.java:286)
                         W      ... 4 more
```

This causes the app to crash

oxtoacart:
@atavism Interesting.  Does the LanternMobileTestbed app crash running on the same device?  Does the crash happen with and/or without Lantern VPN enabled?

oxtoacart:
@atavism Also, is there a message directly related to the crash?  It's not immediately clear why that particular stack trace would cause the whole app to crash.

atavism:
@oxtoacart It hasn't happened with the LanternMobileTestbed app but with the Lantern app itself. And with the VPN disabled. But, actually, it looks the cause of the crash wasn't related to this. I fixed that issue, and I'm not seeing this error at all now. I'm conducting tests with the VPN turned on and off. So most likely a false alarm! I'll let you know if I encounter it again.

atavism:
BTW, it looks like some of our tests are failing now? but in ``balancer_test.go`` and unrelated to this PR.

oxtoacart:
@atavism Okay to merge this?

atavism:
Yep, working great for me!

Issue #3787, opened 2016-03-16
Settings are not propagated after change


xiam:
For instance, if we have two Lantern UIs running on different tabs, changing setting in one tab does not affect the other.

PR #3788, opened 2016-03-16
JSON format changes and removing unused UI code

xiam:
* Format changes to the JSON we were using to communicate between client and backend.
* Removed unused code from the UI.
* Allocating space for `user_id` and `token` to be saved in the settings file.

### DIFF
--- a/src/github.com/getlantern/flashlight/main/main.go
+++ b/src/github.com/getlantern/flashlight/main/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/getlantern/flashlight/client"
 	"github.com/getlantern/flashlight/config"
 	"github.com/getlantern/flashlight/logging"
+	"github.com/getlantern/flashlight/proxiedsites"
 	"github.com/getlantern/flashlight/ui"
 
 	"github.com/mitchellh/panicwrap"
@@ -265,6 +266,7 @@ func afterStart(cfg *config.Config) {
 
 func onConfigUpdate(cfg *config.Config) {
 	autoupdate.Configure(cfg)
+	proxiedsites.Configure(cfg.ProxiedSites)
 }
 
 func i18nInit() {
--- a/src/github.com/getlantern/flashlight/main/main.go
+++ b/src/github.com/getlantern/flashlight/main/main.go
@@ -24,7 +24,6 @@ import (
 	"github.com/getlantern/flashlight/client"
 	"github.com/getlantern/flashlight/config"
 	"github.com/getlantern/flashlight/logging"
-	"github.com/getlantern/flashlight/proxiedsites"
 	"github.com/getlantern/flashlight/ui"
 
 	"github.com/mitchellh/panicwrap"
@@ -266,7 +265,6 @@ func afterStart(cfg *config.Config) {
 
 func onConfigUpdate(cfg *config.Config) {
 	autoupdate.Configure(cfg)
-	proxiedsites.Configure(cfg.ProxiedSites)
 }
 
 func i18nInit() {
--- a/src/github.com/getlantern/flashlight/main/settings.go
+++ b/src/github.com/getlantern/flashlight/main/settings.go
@@ -87,11 +87,6 @@ func LoadSettings(version, revisionDate, buildDate string) *Settings {
 	return settings
 }
 
-type msg struct {
-	*Settings  `json:",inline"`
-	RedirectTo string `json:"redirect_to"`
-}
-
 // start the settings service that synchronizes Lantern's configuration with every UI client
 func (s *Settings) start() error {
 	var err error
@@ -101,7 +96,7 @@ func (s *Settings) start() error {
 		log.Debugf("Sending Lantern settings to new client")
 		s.Lock()
 		defer s.Unlock()
-		return write(&msg{Settings: s})
+		return write(s)
 	}
 	service, err = ui.Register(messageType, nil, helloFn)
 	return err
@@ -114,37 +109,44 @@ func (s *Settings) read() {
 
 		// We're using a map here because we want to know when the user sends a
 		// false value.
-		msg := (message).(map[string]interface{})
+		var data map[string]interface{}
+		var decoded bool
+
+		if data, decoded = (message).(map[string]interface{}); !decoded {
+			continue
+		}
 
 		var v, ok bool
 
-		if v, ok = msg["auto_report"].(bool); ok {
+		if v, ok = data["auto_report"].(bool); ok {
 			s.SetAutoReport(v)
 		}
 
-		if v, ok = msg["proxy_all"].(bool); ok {
+		if v, ok = data["proxy_all"].(bool); ok {
 			s.SetProxyAll(v)
 		}
 
-		if v, ok = msg["auto_launch"].(bool); ok {
+		if v, ok = data["auto_launch"].(bool); ok {
 			s.SetAutoLaunch(v)
 		}
 
-		if v, ok = msg["system_proxy"].(bool); ok {
+		if v, ok = data["system_proxy"].(bool); ok {
 			s.SetSystemProxy(v)
 		}
 
-		if msg["user_id"] != nil {
+		if data["user_id"] != nil {
 			// This is unmarshaled into a float64, I'm am converting it to string and
 			// then to float32 to catch the case when this is a float32.
-			if id, err := strconv.Atoi(fmt.Sprintf("%v", msg["user_id"])); err == nil {
+			if id, err := strconv.Atoi(fmt.Sprintf("%v", data["user_id"])); err == nil {
 				s.SetUserId(id)
 			}
 		}
 
-		if token, ok := msg["token"].(string); ok {
+		if token, ok := data["token"].(string); ok {
 			s.SetToken(token)
 		}
+
+		service.Out <- s
 	}
 }
 
@@ -234,7 +236,7 @@ func (s *Settings) SetSystemProxy(enable bool) {
 		preferredUIAddr, addrChanged := ui.PreferProxiedUI(enable)
 		if !enable && addrChanged {
 			log.Debugf("System proxying disabled, redirect UI to: %v", preferredUIAddr)
-			service.Out <- &msg{RedirectTo: preferredUIAddr}
+			service.Out <- map[string]string{"redirect_to": preferredUIAddr}
 		}
 	}
 }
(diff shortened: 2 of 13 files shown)


src/github.com/getlantern/flashlight/main/settings.go
```diff
@@ -106,18 +106,47 @@ func (s *Settings) read() {
...
+			if id, err := strconv.Atoi(fmt.Sprintf("%v", data["user_id"])); err == nil {
+				s.SetUserId(id)
+			}
+		}
+
+		if token, ok := data["token"].(string); ok {
+			s.SetToken(token)
+		}
+
+		service.Out <- s
```
xiam: We were not sending this change to all clients.

src/github.com/getlantern/flashlight/main/settings.go
```diff
@@ -195,7 +236,7 @@ func (s *Settings) SetSystemProxy(enable bool) {
 		preferredUIAddr, addrChanged := ui.PreferProxiedUI(enable)
 		if !enable && addrChanged {
 			log.Debugf("System proxying disabled, redirect UI to: %v", preferredUIAddr)
-			service.Out <- &msg{RedirectTo: preferredUIAddr}
+			service.Out <- map[string]string{"redirect_to": preferredUIAddr}
```
xiam: We don't need a field for sending this message.

src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	messageType = `ProxiedSites`
+	messageType = `proxied_sites`
```
xiam: I changed this for completion but we're not using this in the UI anymore.

src/github.com/getlantern/flashlight/ui/proxy.go
```diff
@@ -0,0 +1,24 @@
+// +build pro-experimental
```
xiam: This is an experiment, we don't need to compile this into released binaries.

src/github.com/getlantern/lantern-ui/app/js/app.js
```diff
@@ -155,23 +155,6 @@ var app = angular.module('app', [
       return methods;
     }
   ])
-  .factory('ProxiedSites', ['$window', '$rootScope', 'DataStream', function($window, $rootScope, DataStream) {
```
xiam: Unused code.

src/github.com/getlantern/lantern-ui/app/js/controllers.js
```diff
@@ -108,29 +108,6 @@ app.controller('RootCtrl', ['$rootScope', '$scope', '$compile', '$window', '$htt
 
 }]);
 
-app.controller('UpdateAvailableCtrl', ['$scope', 'MODAL', function($scope, MODAL) {
```
xiam: Unused code.

src/github.com/getlantern/lantern-ui/app/js/controllers.js
```diff
@@ -197,127 +162,3 @@ app.controller('MobileAdCtrl', ['$scope', 'MODAL', 'gaMgr', function($scope, MOD
   };
 
 }]);
-
```
xiam: Unused code.

src/github.com/getlantern/lantern-ui/app/js/services.js
```diff
@@ -56,97 +56,40 @@ angular.module('app.services', [])
...
-          if (settings.ProxyAll) {
-              model.settings.proxyAll = true;
-          }
 
-          if (settings.SystemProxy) {
-              model.settings.systemProxy = true;
-          }
+        if (settings.auto_report) {
+          model.settings.auto_report = true;
+          $rootScope.trackPageView();
```
xiam: I changed this as well, but is it correct to track pageviews only when auto_report is enabled? Also, each time this is changed we'll save a pageview, right? Perhaps we don't want this here.

src/github.com/getlantern/lantern-ui/app/js/services.js
```diff
@@ -56,97 +56,40 @@ angular.module('app.services', [])
...
-          }
-
-          applyPeer(data.data);
-        });
+        if (settings.redirect_to) {
+          console.log('Redirecting UI to: ' + settings.redirect_to);
+          window.location = settings.redirect_to;
+        }
+      },
+      'local_discovery': function(data) {
```
xiam: What do we need this for?

src/github.com/getlantern/lantern-ui/app/partials/modals.html
```diff
@@ -22,118 +10,6 @@
   </div>
 </div>
 
-<div ng-controller="ConfirmResetCtrl" ng-show="show" modal="show" class="modal">
```
xiam: More unused code.

xiam:
Hey @myleshorton, can you take a look at this one? Should close https://github.com/getlantern/lantern/issues/3787 and https://github.com/getlantern/lantern/issues/3786 and it's part of what I've been working on...

src/github.com/getlantern/flashlight/main/main.go
```diff
@@ -266,7 +265,6 @@ func afterStart(cfg *config.Config) {
 
 func onConfigUpdate(cfg *config.Config) {
 	autoupdate.Configure(cfg)
-	proxiedsites.Configure(cfg.ProxiedSites)
```
myleshorton: What's the thought behind removing this @xiam?

src/github.com/getlantern/flashlight/main/settings.go
```diff
@@ -27,15 +29,18 @@ var (
...
+	UserId    int    `json:"user_id,omitempty"`
+	UserToken string `json:"user_token,omitempty"`
+
+	Version      string `json:"version"`
+	BuildDate    string `json:"build_date"`
+	RevisionDate string `json:"revision_date"`
+	AutoReport   bool   `json:"auto_report"`
+	AutoLaunch   bool   `json:"auto_launch"`
+	ProxyAll     bool   `json:"proxy_all"`
+	SystemProxy  bool   `json:"system_proxy"`
```
myleshorton: Why do you want to switch all these names to using underscores @xiam?

src/github.com/getlantern/lantern-ui/app/js/controllers.js
```diff
@@ -138,40 +115,28 @@ app.controller('SettingsCtrl', ['$scope', 'MODAL', 'DataStream', 'gaMgr', functi
...
-  $scope.changeSystemProxy = function(systemProxy) {
-      var obj = {
-        systemProxy: systemProxy
-      };
-      DataStream.send('Settings', obj);
+  $scope.changeSystemProxy = function(value) {
+      DataStream.send('settings', {system_proxy: value});
   }
 
-  $scope.$watch('model.settings.systemProxy', function (systemProxy) {
```
myleshorton: Yeah just wondering again the reason for all these name changes @xiam..

src/github.com/getlantern/lantern-ui/app/js/services.js
myleshorton: Yeah ideally we'd put it somewhere else, but we do ideally only want to report them when there's an actual pageview and not on every click. That said, it's just recording a page view so won't impact unique users, which is the more important stat, so I wouldn't worry too much about it.

It looks like your version is the same as the previous except the name change though right @xiam?

xiam: Yeah, so there's no real reason to keep it here right? then we can move it to a better place. This part of code is only fired when the backend sends a message, and that should happen every time settings are changed, that's why it's pretty odd to have it here.

src/github.com/getlantern/flashlight/main/main.go
xiam: It's currently unused. We don't display not use this information in the UI.

src/github.com/getlantern/flashlight/main/settings.go
xiam: Because it was becoming pretty confusing that this was the only part in the (new) UI that used variables Like `AutoReport` instead of `auto_report` or `autoReport` which are more common in Javascript. With this change I stopped worrying about this detail in the UI.

src/github.com/getlantern/lantern-ui/app/js/controllers.js
xiam: Same thing, this is to reflect the name changes on backend.

src/github.com/getlantern/lantern-ui/app/js/services.js
xiam: It looks like we don't currently use this, right @uaalto?

src/github.com/getlantern/flashlight/main/main.go
myleshorton: Right, but we still use proxied sites as a general thing in Lantern @xiam! This change makes it so those aren't configured from the cloud config.

xiam: Oh right! that's true, I only meant to disable the ui service but this change is definitely wrong.

src/github.com/getlantern/flashlight/main/settings.go
myleshorton: yeah it seems like camel case is overall the most widely used standard for js (see, for example, http://www.w3schools.com/js/js_conventions.asp and https://google.github.io/styleguide/javascriptguide.xml), which I'm a little surprised Go doesn't just automatically convert to, but it also looks like on the frontend in a lot of cases you're converting *from camel case* to using underscores (https://github.com/getlantern/lantern/pull/3788/files#diff-8a1026309248e12a3231dfa77eb4755eR81), and there are only a few places where we actually have caps at the beginning. Not a huge deal, but just seems like a lot of code to change when in most cases we're actually moving to a *less* standard naming scheme!

src/github.com/getlantern/lantern-ui/app/js/services.js
myleshorton: Well, I'm guessing the original idea was that if a user had previously not had tracking on but just turned it on, then we should track a pageview

src/github.com/getlantern/flashlight/main/settings.go
xiam: Yes, it was more to avoid having an uppercase first letter (`AutoReport` instead of `auto_report` or `autoReport`). I'm gonna change them back to camelcase and first lowecase letter.

xiam:
Hey @myleshorton, I think it's better now, including the enabling/disabling anonymous stats and tracking a pageview only once. I added a few debugging messages too to make sure everything works as supposed.

myleshorton:
LGTM @xiam! Thanks.

PR #3796, opened 2016-03-16
Setting Host header in outgoing request, closes #3792

### LINKED ISSUE #3792: Lantern strips Host header, breaks sites that use virtual hosting
In particular http://connect.ancestry.com and http://dna.ancestry.com.

Various resources at c.mfcreative.com return 404 errors, like http://c.mfcreative.com//SSS_NAV/1.0.1080/Content/2.1/header.css.

We need to understand what's going on here to determine if it's potentially a broader issue.

screen shot 2016-03-16 at 11 19 53 am

oxtoacart:


src/github.com/getlantern/flashlight/client/reverseproxy.go
```diff
@@ -35,6 +35,9 @@ func (client *Client) newReverseProxy(bal *balancer.Balancer) *httputil.ReverseP
 		// field when upstream servers are trying to determine the client IP.
 		// We need to add also the X-Lantern-Device-Id field.
 		Director: func(req *http.Request) {
+			// Add back the Host header which was stripped by the ReverseProxy. This
+			// is needed for sites that do virtual hosting.
+			req.Header.Set("Host", req.Host)
```
myleshorton: It's definitely stripped before our Director function is called right @oxtoacart?

oxtoacart: Yes, it's stripped by the http.Server, which is using the ReverseProxy as a Handler, so by the time it gets to the ReverseProxy's logic, the header is already gone.

myleshorton:
LGTM.

PR #3804, opened 2016-03-17
detour using eventual connection

fffw:
It's yet another workable version of detour, and the most clean version to date. Inspired and powered by @oxtoacart 's brilliant insight of eventual connections in https://github.com/getlantern/lantern/pull/3663, and the idea of parallel access.

It is still incomplete for now (though not that far), just to track progress. Known issues:
1. Accumulated goroutines in `Read` and `Write`.
2. Data races in `eventual` after added `Stop` function.
3. Lacks of `LocalAddr`, `RemoteAddr` and deadline functions.


oxtoacart:
@fffw I'm looking forward to reviewing this, let me know when you want me to take a look.

oxtoacart:
@fffw Is this ready to review?  There's a bug with detour (#3900) and rather than fixing it on the old detour, I'd prefer to fix it on the latest and greatest.

fffw:
@oxtoacart unfortunately it is still not ready for review

oxtoacart:
@fffw Anything I can help with?

fffw:
@oxtoacart The main issue is accumulated go routines, and occasionally failed test. I should make the code clear enough to find bugs and prove the correctness.

fffw:
It's feature complete now, but three issues remaining:
1. `eventual.Get` accumulates goroutines (eventual.go#L104).
2. `CLOSE_WAIT` connections.
3. I forgot...

src/github.com/getlantern/eventual/eventual.go
```diff
@@ -51,34 +56,59 @@ func DefaultGetter(val interface{}) Getter {
...
+	// can still exit the goroutine.
+	valCh := make(chan interface{}, 1)
 	go func() {
 		v.wg.Wait()
 		valCh <- v.val.Load()
 	}()
-
 	select {
 	case val := <-valCh:
+		if val == nil { // when Stop() is called before value is set
```
oxtoacart: val == nil doesn't necessarily mean that the channel was closed.  Safer is:

```go
case val, open := <-valCh:
      if !open {
```

src/github.com/getlantern/eventual/eventual.go
```diff
@@ -51,34 +56,59 @@ func DefaultGetter(val interface{}) Getter {
...
+func (v *value) Get(timeout time.Duration) (ret interface{}, valid bool) {
+	if atomic.LoadInt32(&v.gotFirst) == intTrue {
 		// Short-cut used once value has been set, to avoid extra goroutine
 		return v.val.Load(), true
 	}
 
-	valCh := make(chan interface{})
+	// Make it buffered so if the caller no longer receives on the channel, we
+	// can still exit the goroutine.
+	valCh := make(chan interface{}, 1)
```
oxtoacart: @fffw You said that there's a problem with the below goroutine leaking, but this looks like you fixed it.  Are there still situations where the goroutine can leak?

oxtoacart: Oh, I guess if the value is never set, this goroutine will leak.

src/github.com/getlantern/eventual/eventual.go
fffw: Good catch @oxtoacart

src/github.com/getlantern/eventual/eventual.go
fffw: Yes exactly. I didn't find a way to fix it.

src/github.com/getlantern/eventual/eventual.go
```diff
@@ -51,34 +56,59 @@ func DefaultGetter(val interface{}) Getter {
...
+		v.updates <- val
+		v.muUpdates.RUnlock()
+	}
 }
 
 func (v *value) processUpdates() {
 	for val := range v.updates {
 		v.val.Store(val)
-		if v.gotFirst == FALSE {
+		if v.gotFirst == intFalse {
```
oxtoacart: To be safe, we should probably use atom.LoadInt32(&v.gotFirst).  In fact, we could just make a new utility function:

```go
func (v *value) signalFirst() {
    if atomic.CompareAndSwapInt32(&v.gotFirst, intFalse, intTrue) {
        v.wg.Done()
    }
}
```

And then use that both inside the loop and after the loop.

src/github.com/getlantern/eventual/eventual.go
```diff
@@ -51,34 +56,59 @@ func DefaultGetter(val interface{}) Getter {
...
 
-func (v *value) Get(timeout time.Duration) (interface{}, bool) {
-	if atomic.LoadInt32(&v.gotFirst) == TRUE {
+// Get waits the value to be set and returns it, or returns nil if times out or
+// Stop() is called. valid will be false in latter case.
+// TODO: Get should happen after Set if no timeout provided.
+func (v *value) Get(timeout time.Duration) (ret interface{}, valid bool) {
+	if atomic.LoadInt32(&v.gotFirst) == intTrue {
 		// Short-cut used once value has been set, to avoid extra goroutine
 		return v.val.Load(), true
```
oxtoacart: This isn't technically true if Stop() was called.  We should really be returning false, since we never did get a value.

oxtoacart: Oh, never mind, when stop is called we don't set gotFirst, so this is good!

src/github.com/getlantern/eventual/eventual.go
```diff
@@ -51,34 +56,59 @@ func DefaultGetter(val interface{}) Getter {
...
+		close(v.updates)
+		v.muUpdates.Unlock()
+	}
 }
 
-func (v *value) Get(timeout time.Duration) (interface{}, bool) {
-	if atomic.LoadInt32(&v.gotFirst) == TRUE {
+// Get waits the value to be set and returns it, or returns nil if times out or
+// Stop() is called. valid will be false in latter case.
+// TODO: Get should happen after Set if no timeout provided.
```
oxtoacart: I'm not sure what you mean by this?  If you want to wait forever for Set, shouldn't you just set a really high timeout?

src/github.com/getlantern/eventual/eventual_test.go
```diff
@@ -14,6 +15,7 @@ const (
 )
 
 func TestSingle(t *testing.T) {
+	goroutines := runtime.NumGoroutine()
```
oxtoacart: This way of testing is sensitive to previous tests.  If a previous test spawned a goroutine that ends after this line but before you call `runtime.NumGoroutine()` again, the count will be wrong.

oxtoacart: I only mention this because I see `TestNoRace` fail sporadically.

src/github.com/getlantern/eventual/eventual.go
oxtoacart: You know, it turns out that it doesn't matter.  It's not possible to set a nil value in an atomic.Value, so using nil here is fine.

src/github.com/getlantern/eventual/eventual.go
oxtoacart: I guess as long as someone calls Close(), then it's okay right?

oxtoacart: Sorry, I meant Stop().

src/github.com/getlantern/eventual/eventual.go
fffw: The purpose is to prevent below from happening. It's not quite easy. Maybe we can simply document it.
```
v.Set(1)
i, valid := v.Get(0) // valid can be false
```

src/github.com/getlantern/eventual/eventual_test.go
fffw: Would increasing sleep time work? Just want a way to capture leaked goroutines. Another way would be to `pprof.Lookup("goroutine")` and compare entries one by one.

src/github.com/getlantern/eventual/eventual.go
oxtoacart: The example code doesn't seem surprising to me.  The semantics of Get are that it waits up to the timeout and if it doesn't get a value by then, it returns false.  With a timeout of 0, it doesn't seem surprising that it fails to get a value.  For consistency with other methods in the standard library, we could add a feature that v.Get(-1) never times out.

src/github.com/getlantern/eventual/eventual_test.go
oxtoacart: Yes!  If you look at my PR, you'll see that I did this in with your detour tracker - https://github.com/getlantern/lantern/pull/3993/files#diff-dfac99737640cd22fdabf4fc7fb4312bR310

I'll refactor that into its own package and then use it from the eventual test.

oxtoacart: Done - https://github.com/getlantern/lantern/pull/3993/files#diff-8cea73b850590a3be8f02149d20e2d66R1

oxtoacart:
I tried running with this for a while and it seems to degrade over time.  I see a lot of EOF errors and `Unable to obtain connection within timeout`, even though I'm not running with any blocking.  I'm going to spend some time today looking at the detour code and debugging.

oxtoacart:
I just noticed that this is assigned to me. 1.5 years for a PR is probably a new record for me. Apologies for the slow response @joesis!

joesis:
Ha totally forgot it. It's obsolete anyway.

PR #3806, opened 2016-03-17
separate config fetching and add new headers closes #3800

### LINKED ISSUE #3800: Send pro tokens

myleshorton:
The underlying change here is to add the new headers necessary to fetch pro configs. Doing that, however, relies on fetching data from settings, which created a potential dependency cycle with config. Passing funcs gets around that, and I did a little refactoring while there to separate config fetching from the config itself, which seems to make things a bit saner all around.


myleshorton:
Hey @oxtoacart if you're free can you give this one a look?

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -127,49 +109,13 @@ func majorVersion(version string) string {
...
-				log.Errorf("Could not rename file from %v to %v: %v", path, configPath, err)
-			} else {
-				log.Debugf("Copied old config at %v to %v", path, configPath)
-				return true
-			}
-		}
-	}
-	return false
-}
-
```
myleshorton: This was unused.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -211,42 +157,6 @@ func Init(version string, configDir string, stickyConfig bool, flags map[string]
...
-				return cfg.updateFrom(bytes)
-			}
-		}
-	} else {
-		log.Errorf("Could not fetch cloud config %v", err)
-		return mutate, waitTime, err
-	}
-	return mutate, waitTime, nil
-}
-
```
myleshorton: moved to new fetcher..

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -349,13 +259,15 @@ func (cfg *Config) ApplyDefaults() {
 	}
 
 	if cfg.CloudConfig == "" {
-		cfg.CloudConfig = chainedCloudConfigUrl
+		cfg.CloudConfig = chainedCloudConfigURL
```
myleshorton: Just cleaning up some things the linter flagged...

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -349,13 +259,15 @@ func (cfg *Config) ApplyDefaults() {
...
 	}
 
-	if cfg.Client != nil {
-		cfg.applyClientDefaults()
+	if cfg.Client == nil {
+		cfg.Client = &client.ClientConfig{}
```
myleshorton: This one is interesting. It seems like whenever we want to apply the defaults we also want client defaults. Unless we add Client to config, though, we won't get those. I didn't have any immediate reason for this other than it repeatedly tripping up tests, but it just seems like what we want here.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -375,13 +287,6 @@ func (cfg *Config) ApplyDefaults() {
 	if cfg.TrustedCAs == nil || len(cfg.TrustedCAs) == 0 {
 		cfg.TrustedCAs = defaultTrustedCAs
 	}
-
-	if cfg.Client.DeviceID == "" {
-		// There is no true privacy or security in instance ID.  For that, we rely on
-		// transport security.  Hashing MAC would buy us nothing, since the space of
-		// MACs is trivially mapped, especially since the salt would be known
-		cfg.Client.DeviceID = base64.StdEncoding.EncodeToString(uuid.NodeID())
-	}
```
myleshorton: Moved this. Previously, if cfg.Client was nil, this would create a panic. Moved to within applyClientDefaults along with everything else that changes the client defaults.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -541,7 +385,7 @@ func (updated *Config) updateFrom(updateBytes []byte) error {
 			wlDomains[domain] = true
 		}
 		updated.ProxiedSites.Cloud = make([]string, 0, len(wlDomains))
-		for domain, _ := range wlDomains {
+		for domain := range wlDomains {
```
myleshorton: linter...

src/github.com/getlantern/flashlight/config/config.go
oxtoacart: Seems okay.  I think this was a holdover from when flashlight had a server mode.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -397,22 +302,7 @@ func (cfg *Config) applyClientDefaults() {
...
-					PoolSize:       0,
-					MasqueradeSet:  cloudflare,
-					MaxMasquerades: 20,
-					QOS:            10,
-					Weight:         4000,
-					Trusted:        true,
-				},
-			}
-
-		*/
```
oxtoacart: If we're not using FrontedServers anymore, perhaps we should remove them from the client config.

src/github.com/getlantern/flashlight/main/main.go
```diff
@@ -53,7 +54,12 @@ func init() {
 }
 
 func logPanic(msg string) {
-	cfg, err := config.Init(flashlight.PackageVersion, *configdir, *stickyConfig, flagsAsMap())
+	// Request the config via either chained servers or direct fronted servers.
+	// We have to do the full version here because we're actually trying to
+	// submit this panic to our logs.
+	cf := util.NewChainedAndFronted(client.Addr)
+	configFetcher := config.NewFetcher(settings.GetUserID, settings.GetToken, cf)
```
oxtoacart: What if SetUserID() hasn't been called before the fetcher tries to GetUserID?  Same with Token.

myleshorton: It's passing the funcs not the values. We then only pass the headers if the values are populated. See https://github.com/getlantern/lantern/pull/3806/files#diff-09edfa285ec7a6c867e5d7471691800aR105

src/github.com/getlantern/flashlight/config/config.go
myleshorton: Yeah good idea -- done!

src/github.com/getlantern/flashlight/config/config.go
myleshorton: Yup I think you're exactly right @oxtoacart! Good point.

src/github.com/getlantern/flashlight/main/main.go
oxtoacart: Yeah, my question was more around whether it breaks anything to request a config without these headers?

myleshorton: We'll only have those values if the user is pro AFAIK, so it will have to work without them for free (and does since we don't currently pass them in free).

myleshorton: The user ID is interesting though -- that's assigned by the pro server, is that correct @xiam? Does it also assign it to free users? Regardless it will get sent up whenever it's populated @oxtoacart, but we could trigger it too.

myleshorton: Adding @uaalto and @atavism too to make sure we're on the same page

xiam: The UserID is generated by the pro server with `POST /user-create`, this is currently triggered by the Pro UI. Free users also have need to have an UserID, that's how we can assign them a referral code.

oxtoacart: @xiam Interesting.  So Pro needs the UI in order to be able to function?  In other words, Pro can't run headless?

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -71,11 +71,12 @@ func Run(httpProxyAddr string,
 	beforeStart func(cfg *config.Config) bool,
 	afterStart func(cfg *config.Config),
 	onConfigUpdate func(cfg *config.Config),
+	configFetcher *config.Fetcher,
```
oxtoacart: This breaks liblantern.  In particular, github.com/getlantern/lantern doesn't pass a configFetcher.

oxtoacart: Consequently, this breaks mobile.

src/github.com/getlantern/flashlight/main/main.go
xiam: Lantern does require the UI to sign-up for pro, with that you generate an appropriate yaml file and with that yaml file you can run headless from that point on, if you want.

oxtoacart: Gotcha.  So basically you need a UI to log in.

src/github.com/getlantern/flashlight/flashlight.go
myleshorton: hmnn...ok I'll look at that @oxtoacart great catch.

testpackages.txt
```diff
@@ -22,6 +22,7 @@ github.com/getlantern/geolookup
 github.com/getlantern/golog
 github.com/getlantern/idletiming
 github.com/getlantern/keyman
+github.com/getlantern/lantern
```
myleshorton: Adding this @oxtoacart to make sure our tests catch issues with breaking liblantern

oxtoacart: Nice!

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -145,6 +146,7 @@ func applyClientConfig(client *client.Client, cfg *config.Config, proxyAll func(
 		Version, RevisionDate)
 	// Update client configuration
 	client.Configure(cfg.Client, proxyAll)
+	proxiedsites.Configure(cfg.ProxiedSites)
```
myleshorton: Including your suggested change on the proxied sites config here @oxtoacart, as this will at least ensure we use the client-configured proxied sites initially.

myleshorton:
Note I still haven't fixed the mobile compatibility problem @oxtoacart -- still need to figure out how to send in those new headers from mobile

myleshorton:
OK @oxtoacart made this at least not break mobile. We still need to wire up the user ID and token @todd, but there's at least some placeholder code at https://github.com/getlantern/lantern/pull/3806/files#diff-b8eeac50cbcb28f059813ada80be6c8fR82

src/github.com/getlantern/lantern/lantern.go
```diff
@@ -79,12 +79,25 @@ func AddLoggingMetadata(key, value string) {
...
+type userConfig struct {
+}
+
+func (uc *userConfig) GetToken() string {
+	return ""
+}
+
+func (uc *userConfig) GetUserID() int {
+	return 0
+}
```
atavism: Great! This is actually almost identical to what I already have so we could just merge it and I'll fix it on my side

oxtoacart:
LGTM.  detour test is failing, but it's failing on devel too.

myleshorton:
Thanks @oxtoacart! Yeah I was just looking at that detour test.

Issue #3812, opened 2016-03-18
Need new way of getting Android version as of NDK L


oxtoacart:
NDK L removed the __system_property_get() API, which osversion uses to get the Android version.  See [here](http://stackoverflow.com/questions/26722040/replacement-for-system-property-get-in-android-l-ndk) for discussion.

oxtoacart:
One option is to just pass it in at initialization time, which would allow us to get it from the Android Java app, where the OS version is easily accessible.

PR #3813, opened 2016-03-18
Update dependencies

oxtoacart:
This updates the dependencies for Lantern Desktop and Mobile.  That includes removing unused dependencies and upgrading remaining dependencies to latest versions (typically from master).  Needless to say, this is a pretty big change that we want to test thoroughly.

Note - there is some loss of functionality due to #3812.  Right now, I just return "Unknown 0.0" as the Android version.  This affects logging to Loggly.


myleshorton:
Looks like we only use github.com/mitchellh for github.com/mitchellh/panicwrap, but we include all of github.com/mitchellh. Maybe clean that up while you're at it @oxtoacart?

This PR is exciting to me for some strange reason.

myleshorton:
although maybe panicwrap internally relies on that..hmnn... maybe not to worry @oxtoacart

myleshorton:
actually it doesn't look like it does need it: 

https://github.com/mitchellh/panicwrap/blob/master/panicwrap.go

oxtoacart:
@myleshorton This PR already removes the non-panicwrap mitchelh stuff.  I basically removed everything from src/github.com other than getlantern and then refetched the needed dependencies using `go get`.

myleshorton:
Oh cool @oxtoacart. I must have misread the diff - nice.

myleshorton:
OK, this looks awesome to me, especially all the golang.org org updates that might actually fix a good deal. Any objections to me just merging this @oxtoacart?

oxtoacart:
@myleshorton It's just kind of scary, but I guess devel is going to get some pretty heavy testing anyway and we can/should fail forward on this stuff anyway.  Let me try running on mobile real quick and make sure it looks happy, then we can merge.

myleshorton:
Sounds good @oxtoacart! Yeah I agree it's a bit scary, although honestly I'm also a little frightened that we're using any of that older gomobile code in particular.

oxtoacart:
@myleshorton Okay, mobile works, so thumbs up on pulling.

myleshorton:
ok i'll wait for CI and then merge! detour test will probably fail again (was looking into it, and it's totally random), but that's ok. thanks @oxtoacart!

myleshorton:
OK merging...

Issue #3814, opened 2016-03-18
Not always proxying config requests through CloudFlare


myleshorton:
$ heroku logs -t | grep x.x.x.x

2016-03-18T14:28:22.112577+00:00 heroku[router]: at=info method=GET path="/cloud.yaml.gz" host=hostname request_id=3416694a-6bad-4229-8304-1e73300dd871 fwd="x.x.x.x,54.240.147.55" dyno=web.4 connect=0ms service=16ms status=200 bytes=62153

The config server only sees the CloudFront but not the CloudFlare request!!

oxtoacart:
@myleshorton Interesting.  I double-checked the CloudFlare configuration and it looks fine - it's set to Caching Level = No Query String, which means that it only serves from cache if URL doesn't include a querystring.  Ours does (from what I can see in the code).

oxtoacart:
The CloudFlare requests go through the chained servers.  Perhaps they're stripping out the querystring somehow - I wonder if this could be related to the changes I made for #3792.

oxtoacart:
I double checked, and Heroku will log the querystring as part of the Path if it's present.  So it looks like the request that came through CloudFront didn't have a querystring.

oxtoacart:
BTW, regarding #3792.  I just ran a test with the old http-proxy and the new http-proxy, and both pass the request to config-server including the querystring, so I don't think the problem is there.

oxtoacart:
Okay, the reason that I don't see a querystring for CloudFront is that CloudFront is configured not to pass querystrings to the origin server.

oxtoacart:
So, I'm not sure what's going on for you @myleshorton .  In my case, the client is sending a query string for both CloudFlare and CloudFront.  The http-proxy is passing the querystring to CloudFlare.  Both CloudFlare and CloudFront are issuing requests to the config-server.  So I seem to be in okay shape.

Can you send me the ip of your proxy server that's handling the CloudFlare request?

myleshorton:
Yeah I'm now on a new IP, @oxtoacart, but no matter what I keep getting assigned to 45.55.145.84.

Good find on CloudFront not passing the query string along. I'm now seeing both requests at least make it to the config server and am continuing to investigate!

oxtoacart:
@myleshorton Okay, cool.

myleshorton:
So this is odd @oxtoacart. Now the request is definitely making it through CloudFlare and is hitting the config server, but it looks like it might not be going through our proxy.

[...]

PR #3818, opened 2016-03-18
Force whitelisting getiantem.org in detour

oxtoacart:
This is a durable solution to close getlantern/lantern#3814, even if the fronted domain changes or happens to fall off the proxied sites list for some reason.


myleshorton:
It seems like it might be cleaner to use the existing permanent flag in `wlEntry` @oxtoacart and to modify `RemoveFromWl` to respect it, as in:

```go
func RemoveFromWl(addr string) {
  muWhitelist.Lock()
  defer muWhitelist.Unlock()
  entry, ok := whitelist[addr]
  if ok && !entry. permanent {
    delete(whitelist, addr)
  }
}
```

oxtoacart:
@myleshorton Oh yeah, I like that.  Let me do that.

oxtoacart:
@myleshorton Oh, the problem with your suggestion is that it breaks any previous logic that relies on RemoveFromWl even removing permanent ones.  So, I think it's safer to stick with what I had.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -69,6 +71,13 @@ type Config struct {
 
 // StartPolling starts the process of polling for new configuration files.
 func StartPolling() {
+	// Force detour to whitelist chained domain
+	u, err := url.Parse(chainedCloudConfigUrl)
+	if err != nil {
+		log.Fatalf("Unable to parse chained cloud config URL: %v", err)
```
myleshorton: Maybe Fatalf is extreme here @oxtoacart?

myleshorton:
Makes sense on `RemoveFromWl` @oxtoacart

src/github.com/getlantern/flashlight/config/config.go
oxtoacart: Well, this means that our code is fundamentally broken.  It's not like it'll happen spuriously, and this way we find out during development rather than possibly missing it.

myleshorton:
tlsdialer_test.go is still failing, but that one's also random...merging!

fffw:
Oh @oxtoacart @myleshorton it's indeed a bug that `RemoveFromWl` doesn't respect the `permanent` flag!

PR #3838, opened 2016-03-21
Displaying same version information in Mobile as in Desktop

oxtoacart:
This displays the same version string in Mobile as we do in the Desktop app.


atavism:
Sorry I somehow didn't notice this earlier.. LGTM!

PR #3843, opened 2016-03-22
Fix the latest android beta apk download link

netroby:


fffw:
Good catch! Thanks @netroby !

Issue #3842, opened 2016-03-22
make: *** [src/github.com/getlantern/flashlight/ui/resources.go] Error 1


fliar:
hi i've found lantern source is updated so try to build from source.
last time i built is when lantern 2.0.16 wai released and it was fine.

document shows it's even simpler (no docker required?), but it's not for me
this time i found i had to install npm, node, and gulp from npm, then i got this:
~~~
sudo env "PATH=$PATH" make lantern
linuxnpm WARN package.json colorspace@1.0.1 No repository field.
npm WARN package.json gulp-minify-css@1.2.3 No repository field.
npm WARN package.json gulp-minify-html@1.0.5 No repository field.
make: *** [src/github.com/getlantern/flashlight/ui/resources.go] Error 1
~~~
couldn't figure out the problem from the output

fliar:
alright, i found it, on ubuntu 14.04 x64, the doc needs modification
1. install nodejs & npm:
~~~
curl -sL https://deb.nodesource.com/setup_4.x | sudo -E bash -
sudo apt-get install -y nodejs
~~~
2. install gulp(maybe it's automatic, not sure)
~~~
sudo npm install gulf -g
~~~
3. if showed '', try this
~~~
sudo apt-get install libc6-dev-i386
~~~
4. build it
~~~
sudo env "PATH=$PATH" make linux
~~~

myleshorton:
Great care to make a pull request?

fffw:
Resolved by https://github.com/getlantern/lantern/pull/3860

PR #3860, opened 2016-03-24
fix prerequisites for building lantern issue #3842

fliar:


fffw:
Thanks @fliar. I'm going to merge it and perhaps rephrase afterwards.

PR #3868, opened 2016-03-25
idletiming correctly set deadline fix #3866

### LINKED ISSUE #3866: Error resend buffer: i/o timeout
New in 2.1.1. Reported by more than one user.

It's a bit unusual as `Write` should seldomly time out. It also times out too soon (less than 1 sec). We set write deadline somewhere?

```
detour: detour.go:249 Error while resend buffer to <masked>:80: write tcp <masked>:49482-><masked>:443: i/o timeout
```

fffw:
In https://golang.org/pkg/net/#Conn
> // A zero value for t means I/O operations will not time out.


oxtoacart:
LGTM!

PR #3869, opened 2016-03-25
Configuring correct proxiedsites package, closes #3864

### LINKED ISSUE #3864: Default whitelisted sites still get bypassed in PAC file
For example 'whatismyipaddress.com' should always get proxied, but it doesn't.

cc: @myleshorton

oxtoacart:


fffw:
LGTM

myleshorton:
Nice @oxtoacart! And this was definitely only an issue on devel right?

oxtoacart:
@myleshorton Yeah, I think it broke on [this commit](https://github.com/getlantern/lantern/commit/a1d3af508728f8c336fb046343dd88f41e10eed0) which is not in master.

PR #3884, opened 2016-03-29
delete "douyutv.com" from "proxiedsites" #3873

kehao95:
delete "douyutv.com" from "proxiedsites" #3873


fffw:
LGTM. Though it will not take effect until we run genconfig again next time.

myleshorton:
This needs to also be deleted in lantern_aws unfortunately @fffw -- otherwise the global sites will overwrite this while running lantern.

fffw:
Did that in https://github.com/getlantern/lantern_aws/commit/e29416ab386628bab13157b344a498424a9b135e. Thanks @myleshorton !

fffw:
As we plan to host common configuration separately on cloud, we can also always fetch latest configuration from cloud when building Lantern. By this way we don't need to have two copies of some list.

PR #3887, opened 2016-03-29
Configuring proxiedsites in main instead of flashlight, removes ui as dependency for mobile

oxtoacart:
This gets rid of the transitive mobile -> proxied sites -> ui dependency.

@myleshorton This undoes what you did here - https://github.com/getlantern/lantern/commit/e73ca0dd1bbe2f5f376c111a9a5cdb0ec7b96f73#diff-f51f873318b593f0a492e1abb11a57c2R149  What was the reason for that initial change?  Will it cause any problems to go back to just configuring proxiedsites in main?


oxtoacart:
cc: @atavism 

While testing this, I noticed that the ui dependency added almost 2 MB to the mobile release build, so I think we should definitely fix this.

myleshorton:
LGTM!

PR #3943, opened 2016-04-09
show feeds

fffw:


fffw:
@Derekf5 It's the WIP of newsfeed, in case you want to have a look.

Pending works:

1. Switch 'All' to ic-list icon
2. Properly fade out at the bottom of each item.
3. Only show if locale is zh_CN
...

myleshorton:
These generally look great to me @fffw -- merging in.

PR #3963, opened 2016-04-13
Added support for obfs4

oxtoacart:


myleshorton:
Hey so @oxtoacart is the test failure expected because of how you changed the merge to make it more readable?

myleshorton:
it's looking for that tor obfs4 package..

oxtoacart:
@myleshorton That's correct.  The full PR with dependencies is here - https://github.com/getlantern/lantern/pull/3971

src/github.com/getlantern/flashlight/client/chained.go
```diff
@@ -25,82 +22,50 @@ var ForceChainedProxyAddr string
...
 	Trusted bool
+
+	// PluggableTransport: If specified, a pluggable transport will be used
+	PluggableTransport string
+
+	// PluggableTransportSettings: Settings for pluggable transport
+	PluggableTransportSettings map[string]string
 }
 
 // Dialer creates a *balancer.Dialer backed by a chained server.
```
myleshorton: Minor nit, and this is only related to this PR in the sense that with pluggable transports there's more for the code reader to think about, but I think a comment for deviceID would be helpful here @oxtoacart -- i.e. make it clear it's not related to pluggable transports at all but just identifies this device to the server.

oxtoacart: Fixed.

PR #3968, opened 2016-04-14
Show feeds on mobile

atavism:


myleshorton:
Hey @oxtoacart you mind taking a look at this one?

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
```diff
@@ -0,0 +1,84 @@
...
+    public void onViewCreated(View view, Bundle savedInstanceState) {
+        super.onViewCreated(view, savedInstanceState);
+        Log.d(TAG, "Created view for " + this.feedName);
+        new Thread() {
+            public void run() {
+
+                Lantern.FeedByName(feedName, new Lantern.FeedRetriever.Stub() {
+                    public void AddFeed(String title, String desc, String image, 
+                            String url) {
+                        mFeedItems.add(new FeedItem(title, desc, image, url));
```
oxtoacart: mFeedItems is an ArrayList, which is not thread safe.  You can wrap it with Collections.synchronizedList() and then you should be fine.

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/model/FeedItem.java
```diff
@@ -0,0 +1,16 @@
+package org.getlantern.lantern.model;
+
+public class FeedItem {
+    public String Title;
```
oxtoacart: nit, but in Java variables should begin with lowercase letters.

src/github.com/getlantern/lantern/feed.go
```diff
@@ -0,0 +1,170 @@
...
+const (
+	// the Feed endpoint where recent content is published to
+	// mostly just a compendium of RSS feeds
+	feedEndpoint = `https://feeds.getiantem.org/%s/feed.json`
+)
+
+var (
+	feed Feed
+
+	httpClient *http.Client
```
oxtoacart: There's no need for this to be global.

src/github.com/getlantern/lantern/feed.go
```diff
@@ -0,0 +1,170 @@
+package lantern
```
oxtoacart: This seems like it could be its own package.

src/github.com/getlantern/lantern/feed.go
```diff
@@ -0,0 +1,170 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"math"
+	"net/http"
+	"strings"
+
+	"github.com/getlantern/eventual"
+	"github.com/getlantern/flashlight/util"
```
oxtoacart: If you put feed into its own package and remove the dependency on flashlight/util, that makes this a nicely isolated package.

oxtoacart: That said, I guess that makes packaging for Android more complicated, so never mind.

src/github.com/getlantern/lantern/feed.go
```diff
@@ -0,0 +1,170 @@
...
+
+	for _, s := range feed.Feeds {
+		for _, i := range s.Entries {
+			entry := feed.Entries[i]
+			// every feed item gets appended to a feed source array
+			// for quick reference
+			feed.Items[s.Title] = append(feed.Items[s.Title], entry)
+		}
+	}
+}
```
oxtoacart: This module could use a unit test.

oxtoacart:
@atavism I can't build this.  I get an error from ProGuard:

```
java.io.IOException: Can't write [/Users/ox.to.a.cart/lantern/src/github.com/getlantern/lantern-mobile/app/build/intermediates/transforms/proguard/release/jars/3/1f/main.jar] (Can't read [/Users/ox.to.a.cart/.gradle/caches/modules-2/files-2.1/commons-io/commons-io/1.3.2/b6dde38349ba9bb5e6ea6320531eae969985dae5/commons-io-1.3.2.jar(;;;;;;**/*.class)] (Duplicate zip entry [org/a/a/a/a.class == commons-io-1.3.2.jar:org/apache/commons/io/IOUtils.class]))
```

Does that look familiar at all?

src/github.com/getlantern/lantern/feed.go
```diff
@@ -0,0 +1,170 @@
...
+	if res, err = httpClient.Do(req); err != nil {
+		log.Errorf("Error fetching feed: %v", err)
+		return
+	}
+
+	defer res.Body.Close()
+	contents, err := ioutil.ReadAll(res.Body)
+
+	if err != nil {
+		log.Errorf("Error reading body: %v", err)
```
oxtoacart: If we're unable to load the feeds, shouldn't we let the client know instead of failing silently?

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/model/FeedItem.java
atavism: Fixed. The reason the variables were uppercase and public before is because I was previously using a library that necessitated it to parse JSON into FeedItem objects.

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
atavism: Thanks! Fixed.

src/github.com/getlantern/lantern/feed.go
atavism: Yeah, I agree. @Derekf5, what should we show in the UI whenever there's an issue fetching the latest feed?

atavism:
@oxtoacart I'm seeing that proguard issue with release builds too. I'll look into that now. If you want to keep testing, doing a debug build should work in the meantime.

oxtoacart:
Titles in mobile feed show <CDATA>

![cdata

atavism:
It's not an issue specific to mobile. There's a problem with the JSON now that's being published to the public feed. Do you know what's going on there, @fffw and @xiam?

fffw:
Yeah desktop even stops showing title at all with `CDATA`. Is it related to switching RSS parser @xiam ?

Maybe use `meta.og.title` and use `title` only as a fallback? Though I don't known what `og` actually means.

atavism:
@fffw ``og`` just means its an open graph meta tag.

src/github.com/getlantern/lantern/feed.go
atavism: @oxtoacart I updated the UI to display an error whenever we encounter an issue fetching the public feed. There's a refresh button that can be pressed to try retrieving it again. 

image

oxtoacart:
To test how this works in a blocked scenario, I set my wifi connection to use our fake DNS server () and then opened Lantern.  The feed view remained empty and eventually the app crashed with this error:

```
FATAL EXCEPTION: main
                         E  Process: org.getlantern.lantern, PID: 10995
                         E  java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.Iterator java.util.ArrayList.iterator()' on a null object reference
                         E      at org.getlantern.lantern.activity.LanternMainActivity.setupFeed(LanternMainActivity.java:256)
                         E      at org.getlantern.lantern.model.GetFeed.onPostExecute(GetFeed.java:53)
                         E      at org.getlantern.lantern.model.GetFeed.onPostExecute(GetFeed.java:13)
                         E      at android.os.AsyncTask.finish(AsyncTask.java:636)
                         E      at android.os.AsyncTask.access$500(AsyncTask.java:177)
                         E      at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:653)
                         E      at android.os.Handler.dispatchMessage(Handler.java:102)
                         E      at android.os.Looper.loop(Looper.java:135)
                         E      at android.app.ActivityThread.main(ActivityThread.java:5254)
                         E      at java.lang.reflect.Method.invoke(Native Method)
                         E      at java.lang.reflect.Method.invoke(Method.java:372)
                         E      at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
                         E      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
```

oxtoacart:
Okay, it looks like I actually had a bad IP address, which is what caused us to be unable to load the feed.  Still, if the feed can't load, we should not crash.

oxtoacart:
@atavism Build is failing now with this:

```
/Users/ox.to.a.cart/lantern/src/github.com/getlantern/lantern-mobile/app/src/main/res/layout/activity_lantern_main.xml:92:21-39: AAPT: No resource found that matches the given name (at 'layout' with value '@layout/feed_error').
```

Missing checkin?

oxtoacart:
This is generally working very well for me.  A couple of issues that I noticed:

1. The tab titles don't always match the feed that's actually shown (I think @percyalpha already noticed this)

2. NY Times articles won't load, giving either EOF or Bad Gateway (note that I am using our DNS blocking trick with DNS server 104.131.168.238)

eof

atavism:
> Okay, it looks like I actually had a bad IP address, which is what caused us to be unable to load the feed. Still, if the feed can't load, we should not crash.

Looks like we weren't checking for a null sources array whenever we setup the feed. With a recent change of mine, that should be impossible, but I added an additional check to make sure this doesn't happen.

atavism:
OK, that new proguard issue affecting release builds should be resolved now.

oxtoacart:
@atavism I've confirmed that I can build release now.  The NYTimes issue is still happening though.

atavism:
@oxtoacart What are you seeing in the logs when you try to load NYT pages? They're loading fine for me. Also, what about the issue with the tab header not matching the feed shown?

oxtoacart:
@atavism I don't see any errors in the log.  Just a "Gateway Timeout" or "EOF" displayed in the web view, after it hangs for a considerable amount of time.  This is on a Nexus 4, running with VPN mode off.

oxtoacart:
@atavism Interestingly, the link in question doesn't work from anywhere, so it's not an issue with our proxy.

http://rss.nytimes.com/c/34625/f/667200/s/4ef5a746/sc/13/l/0L0Snytimes0N0C20A160C0A40C160Cworld0Ceurope0Cpanama0Epapers0Espain0Bhtml0Dpartner0Frss0Gemc0Frss/story01.htm

oxtoacart:
rss.nytimes.com doesn't appear to be accepting connections:

`nc -vz rss.nytimes.com 80`

atavism:
Yeah, I'm seeing the same issue now. None of the NYT RSS links are working for me anymore. Looks like ``rss.nytimes.com`` is unreachable

atavism:
Now I'm actually getting "Not Found" for every NYT article I try to load =) Anyway, this seems unrelated to the current PR, and something we have to address w/ the feed server.

percyalpha:
The Chinese one is still up http://cn.nytimes.com/rss.html

Percy Alpha(PGP
<https://pgp.mit.edu/pks/lookup?op=vindex&search=0xF30D100F7FE124AE>)
(202) 780-5619
www.percya.com   @p3rcya <https://twitter.com/p3rcya>
Independent Internet freedom researcher
Former co-founder of GreatFire

oxtoacart:
@atavism LGTM, merging!

PR #3979, opened 2016-04-14
update translations from Transifex

fffw:


fffw:
These are 100% translated languages, including recent added strings. Other languages will fallback to English if specific key can't be found.

PR #3982, opened 2016-04-15
hide thumbnail when fail to load image fix #3981

### LINKED ISSUE #3981: Don't show image if fails to load

fffw:
@xiam mind have a look?

### DIFF
--- a/src/github.com/getlantern/lantern-ui/app/partials/feed.html
+++ b/src/github.com/getlantern/lantern-ui/app/partials/feed.html
@@ -1,6 +1,6 @@
 <a class="item" target="_blank" href="{{feed.link}}" ng-click="trackFeed(feedsTitle)">
   <div ng-if="feed.image" class="feed-thumbnail">
-    <img ng-src="{{feed.image}}"/>
+    <img ng-src="{{feed.image}}" on-error="hideImage(feed)"/>
   </div>
   <div class="content">
     <span class="feed-source"></span>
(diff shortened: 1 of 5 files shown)


src/github.com/getlantern/lantern-ui/app/partials/feed.html
```diff
@@ -1,6 +1,6 @@
 <a class="item" target="_blank" href="{{feed.link}}" ng-click="trackFeed(feedsTitle)">
   <div ng-if="feed.image" class="feed-thumbnail">
-    <img ng-src="{{feed.image}}"/>
+    <img ng-src="{{feed.image}}" on-error="hideImage(feed)"/>
```
myleshorton: shouldn't this be `onerror` instead of `on-error` @fffw? Or are they equivalent?

myleshorton: `on-error` doesn't seem to work in the example at http://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_onerror

myleshorton: then again, there could be something about angular directives I'm not grasping

fffw: Oops I didn't notice you are commenting @myleshorton . I should use more descriptive name but `on-error` corresponds to the `onError` directive, which knows about the `$scope`, so it can call `$scope.hideImage`. With `onerror` we can only manipulating element properties like `style`.

myleshorton:
LGTM @fffw!

Issue #3940, opened 2016-04-08
Encrypt configuration file


myleshorton:
Users are uploading their configuration files with Lantern servers to public sites. To get around this, we should encrypt it.

Arthur2e5:
// PR: #3985

oxtoacart:
@sunshine19090 The requirement to test is:

1. By default, the lantern.yaml should not be human readable
2. If you use the `-readableconfig` command-line flag, it should be human readable
3. You should be able to switch between using `-readableconfig` and not using it, and Lantern should still work

PR #3985, opened 2016-04-15
Added obfuscation of yaml config, closes getlantern/lantern#3940

oxtoacart:


src/github.com/getlantern/yamlconf/yamlconf.go
```diff
@@ -103,6 +103,14 @@ type Manager struct {
 	// example for fetching config updates from a remote server.
 	CustomPoll func(currentCfg Config) (mutate func(cfg Config) error, waitTime time.Duration, err error)
 
+	// Obfuscate: if true, the on-disk version of the config will be obfuscated by
+	// encrypting it using AES-128 encryption.
```
uaalto: Why this choice? Since it won't provide any kind of security, rot13+zip could've been sufficient. Not that important, but I wonder if a more lightweight solution could be possible.

src/github.com/getlantern/yamlconf/yamlconf.go
```diff
@@ -148,25 +156,25 @@ func (m *Manager) Init() (Config, error) {
...
-		if err == nil {
-			_, err = m.saveToDiskAndUpdate(copied)
-		}
-		if err != nil {
-			return nil, fmt.Errorf("Unable to perform initial update of config on disk: %s", err)
+	// Always save whatever we loaded, which will cause defaults to be
+	// applied and formatting to be made consistent
+	copied, err := m.copy(m.cfg)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to copy config: %v", err)
```
uaalto: We are using %v and %s inconsistently across this function for errors.

src/github.com/getlantern/yamlconf/yamlconf.go
oxtoacart: Easy - golang has a built-in library for doing AES-128.  There's no built-in rot13 library.  Reading/writing AES-128 turns out to be 6 lines of code each.  I can live with that :)

src/github.com/getlantern/yamlconf/yamlconf.go
oxtoacart: Fixed.

src/github.com/getlantern/yamlconf/yamlconf.go
uaalto: I mean lightweight computationally. We are adding up miliseconds to startup by each new feature that we add, and that's how heavy programs get born ;)

src/github.com/getlantern/yamlconf/yamlconf_disk.go
```diff
@@ -25,13 +31,30 @@ func (m *Manager) reloadFromDisk() (bool, error) {
...
+	var cfg Config
+	if m.Obfuscate {
+		log.Trace("Attempting to read obfuscated config")
+		var err1, err2 error
+		cfg, err1 = m.doReadFromDisk(true)
+		if err1 != nil {
+			log.Tracef("Error reading obfuscated config from disk, try reading non-obfuscated: %v", err)
+			cfg, err2 = m.doReadFromDisk(false)
+			if err2 != nil {
+				return false, err1
```
uaalto: In this case, why are we more interested in the first error? Perhaps return both concatenated?

src/github.com/getlantern/yamlconf/yamlconf_disk.go
```diff
@@ -55,6 +78,37 @@ func (m *Manager) reloadFromDisk() (bool, error) {
...
 
+func (m *Manager) doReadFromDisk(allowObfuscation bool) (Config, error) {
+	infile, err := os.Open(m.FilePath)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to open config file %v for reading: %v", m.FilePath, err)
+	}
+	defer infile.Close()
+
+	var in io.Reader = infile
+	if allowObfuscation && m.Obfuscate {
```
uaalto: Why do you check m.Obfuscate here? If this function is called by `reloadFromDisk` and m.Obfuscate isn't true it won't read the obfuscated version in case it fails to read as non-obfuscated one.

uaalto: I mean, it won't make line 53 very useful. Is that correct?

src/github.com/getlantern/yamlconf/yamlconf.go
oxtoacart: Wow, this actually turns out to make a big difference.  Without encryption we're looking at about 300ms to read the config, with encryption about 1.6 seconds!  I'll do a custom ROT13 thing.

src/github.com/getlantern/yamlconf/yamlconf_disk.go
oxtoacart: Good idea, will fix.

src/github.com/getlantern/yamlconf/yamlconf_test.go
```diff
@@ -184,17 +185,26 @@ func TestCustomPoll(t *testing.T) {
...
 }
 
-func assertSavedConfigEquals(t *testing.T, file *os.File, expected *TestCfg) {
+func assertSavedConfigEquals(t *testing.T, m *Manager, file *os.File, expected *TestCfg) {
 	b, err := yaml.Marshal(expected)
 	if err != nil {
 		t.Fatalf("Unable to marshal expected to yaml: %s", err)
 	}
-	bod, err := ioutil.ReadFile(file.Name())
+	stream, err := m.obfuscationStream()
```
uaalto: With this, we are now only testing the obfuscated version path. Perhaps adding both would improve coverage. What do you think?

src/github.com/getlantern/yamlconf/yamlconf_disk.go
oxtoacart: Good point.

src/github.com/getlantern/yamlconf/yamlconf.go
oxtoacart: Turns out those timings were a fluke.  I now see similar times for both (as low as 300ms and as high as 1.7 seconds).  So, I'm sticking with AES.

oxtoacart:
@uaalto This is ready for another review.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -145,18 +118,14 @@ func Init(userConfig UserConfig, version string, configDir string, stickyConfig
 		log.Errorf("Could not get config path? %v", err)
 		return nil, err
 	}
-	run := isGoodConfig(configPath)
```
oxtoacart: @uaalto Until today, I hadn't realized that we were breaking encapsulation and reading/writing the config from within flashlight without using yamlconf.  That scheme obviously broke once we introduced obfuscation.  So, rather than apply obfuscation in both places, I cleaned this up so that everything is handled in yamlconf.

src/github.com/getlantern/yamlconf/yamlconf.go
```diff
@@ -90,11 +90,20 @@ type Manager struct {
 	// FilePath: required, path to the config file on disk
 	FilePath string
 
-	// EmptyConfig: required, factor for new empty Configs
+	// ValidateConfig: optional. If specified, the config loaded from disk is
+	// validated with this function. If validation fails, the config is replaced
+	// with the result of DefaultConfig.
+	ValidateConfig func(cfg Config) error
```
oxtoacart: New function that allows validating the config (needed by flashlight).

src/github.com/getlantern/yamlconf/yamlconf.go
```diff
@@ -90,11 +90,20 @@ type Manager struct {
...
+
+	// DefaultConfig: optional. If specified, and there's a problem reading the
+	// config from disk or that , this is used.
+	DefaultConfig func() (Config, error)
```
oxtoacart: New function that allows application of a default config (in our case, the embedded bootstrap config).

src/github.com/getlantern/flashlight/config/bootstrap.go
```diff
@@ -102,14 +103,15 @@ func MakeInitialConfig(configPath string) error {
 	bytes, err := fs.GetIgnoreLocalEmpty("lantern.yaml")
```
uaalto: Are we ever reading the local filesystem for this? Why? Shouldn't the embedded be the default?

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -77,35 +75,13 @@ type CA struct {
...
-	if fi, err := os.Stat(file); os.IsNotExist(err) {
-		log.Debugf("File does not exist at %v", file)
-		return fi, false
-	} else {
-		log.Debugf("File exists at %v", file)
-		return fi, true
+// hasCustomChainedServer returns an error if the config doesn't have a custom
+// chained server.
+func hasCustomChainedServer(_cfg yamlconf.Config) error {
+	cfg, ok := _cfg.(*Config)
```
uaalto: We are assuming here that just casting to a Config will provide a chained server. The case where we have a valid cast without a valid chained server is not going to happen?

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -115,13 +91,10 @@ func hasCustomChainedServer(configPath, name string) bool {
...
 	// if it has been given a custom config with a custom chained server
 	// list
-	return nc > 0 && nc < 10
-}
-
-func isGoodConfig(configPath string) bool {
-	log.Debugf("Checking config path: %v", configPath)
-	fi, exists := exists(configPath)
-	return exists && hasCustomChainedServer(configPath, fi.Name())
+	if nc <= 0 || nc > 10 {
```
uaalto: ok I see. This is where probably my previous question gets answered. I would say that the function naming is a bit confusing, since what we are really checking there is if the config is valid, not if it has chained servers.

src/github.com/getlantern/yamlconf/yamlconf.go
uaalto: I'm interested in how you did these measurements. Are they comparative differentials between versions of Lantern with and without this or a profiling of the function itself?
Normally symmetric encryption has some overhead, that's why I'm interested in this aspect.

src/github.com/getlantern/yamlconf/yamlconf.go
uaalto: Thanks.

src/github.com/getlantern/flashlight/config/bootstrap.go
oxtoacart: This is using tarfs, which can include an embedded version.  This is existing code which I didn't change, so I'm assuming it does what it's supposed to.

src/github.com/getlantern/yamlconf/yamlconf_disk.go
```diff
@@ -55,6 +79,69 @@ func (m *Manager) reloadFromDisk() (bool, error) {
...
+		}
+	}
+	return cfg, nil
+}
+
+func doReadFromDisk(filePath string, allowObfuscation bool, emptyConfig func() Config) (Config, error) {
+	start := time.Now()
+	infile, err := os.Open(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to open config file %v for reading: %v", filePath, err)
```
uaalto: %v / %s

src/github.com/getlantern/flashlight/config/config.go
oxtoacart: Actually the only thing we've validating is that is has custom chained servers.  Still, at some point it may expand, so I've renamed it.

src/github.com/getlantern/flashlight/config/config.go
uaalto: I agree it's a better abstraction.

src/github.com/getlantern/yamlconf/yamlconf.go
oxtoacart: I added some timing to the loadFromDisk method and printed out the results.  I just wrote a benchmark for thoroughness.  The benchmark reads a realistic config file from disk and decodes the yaml.  Depending on the run, the encryption adds anywhere between 0 and 13% overhead.  Without the yaml decoding the differences are much more severe, but the yaml decoding dwarfs the other work.

oxtoacart: Oh, here's the benchmark - https://gist.github.com/oxtoacart/7e9b43efd08f6625a188bf48f8a2e0f3

src/github.com/getlantern/yamlconf/yamlconf_disk.go
oxtoacart: Left over from existing code, but I'll fix.

src/github.com/getlantern/yamlconf/yamlconf.go
uaalto: Impressive that you created the full benchmark. Thanks!!

oxtoacart: Looks like rot13 by itself (no zip) is less overhead, only about 1.5%, and much more consistent.  I'll switch to that.

oxtoacart:
Okay, I've switched to rot13 and merged from devel.  If tests all pass, I think this is good?

src/github.com/getlantern/rot13/rot13_test.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func TestRoundTrip(t *testing.T) {
+	var in, out, rt bytes.Buffer
+	for i := byte(0); i <= 255; i++ {
+		err := in.WriteByte(i)
+		if !assert.NoError(t, err, "Unable to write byte") {
+			return
+		}
+		if i == 255 {
```
uaalto: I'm curious, why did you decide to break at 255 instead of constraining the loop to `i < 255`?

oxtoacart: Because I wanted to test all 256 possible values of a byte.  If I don't break at 255, i++ brings wraps back to 0, which meets the boundary condition, so this becomes an infinite loop.

uaalto: Ah! I see, byte(0) actually creates a C char-like type. Of course, it makes sense.

uaalto:
LGTM! Kudos for the benchmarking and all the tests in the code!!

oxtoacart:
@uaalto Thanks for the review!

PR #3986, opened 2016-04-16
show scroll bar only on hover

fffw:
@myleshorton It looks better by not show scroll bar by default. On IE/Edge/Firefox the content will reflow when scroll bar is shown. That doesn't affect user experience too much I think.


myleshorton:
I'm actually not going to merge this for now @fffw, as I think the current behavior is actually the expected behavior, at least on OSX but I believe on Windows too. See http://www.w3schools.com/cssref/tryit.asp?filename=trycss_overflow

If there's a good reason for this change that I'm missing, though, we can easily add it in the next release.

myleshorton:
Closing for now @fffw.

PR #3993, opened 2016-04-18
Refactored eventual so that Stop() is no longer necessary

oxtoacart:
The benchmark results are still similar to the original implementation, so I think this is good to use.


src/github.com/getlantern/detour/detour.go
```diff
@@ -213,12 +213,12 @@ func (c *conn) Write(b []byte) (int, error) {
 
 func (c *conn) Close() error {
 	close(c.closed)
-	c.detourAllowed.Stop()
 	_ = c.direct.Close()
 	_ = c.detour.Close()
 
 	log.Tracef("%s: Should detour? %v - Detourable? %v", c.addr, c.direct.ShouldDetour(), c.detour.Detourable())
 	allowed, valid := c.detourAllowed.Get(0)
+	c.detourAllowed.Cancel()
```
oxtoacart: @fffw Due to the slightly different semantics of Cancel, I moved this to after I check detourAllowed.

oxtoacart:
@fffw This is ready for you now.  There was a problem with how I had implemented Cancel, but I've fixed that and all the tests are passing now.

src/github.com/getlantern/eventual/eventual.go
```diff
@@ -56,61 +55,115 @@ func DefaultGetter(val interface{}) Getter {
...
-	// Prevent sending on closed channel
-	if atomic.LoadInt32(&v.stopped) == intFalse {
-		v.muUpdates.RLock()
-		v.updates <- val
-		v.muUpdates.RUnlock()
+	v.mutex.Lock()
+	defer v.mutex.Unlock()
+
+	state := v.getState()
+	settable := state == nil || !state.canceled
```
fffw: We can set initial state in `NewValue` to avoid nil check each time.

src/github.com/getlantern/grtrack/grtrack.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	goroutineNumber = regexp.MustCompile(`goroutine ([0-9]+)`)
+)
+
+type Check func(t *testing.T)
+
+func Start() Check {
```
fffw: `grtrack` is neat! To be more apparent, we can have `Start` to return an interface with `Check(t)` and `CheckAfter(t, waitTime)`. Current implementation is good enough though.

src/github.com/getlantern/eventual/eventual.go
oxtoacart: I like it, done!

src/github.com/getlantern/grtrack/grtrack.go
oxtoacart: Good idea, done!

fffw:
Good. I'll go ahead and merge this. I can merge changes in https://github.com/getlantern/lantern/pull/4028 afterwards.

PR #4013, opened 2016-04-19
Track feed events on mobile

atavism:


src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/model/Utils.java
```diff
@@ -33,4 +43,27 @@ public static boolean isNetworkAvailable(final Context context) {
...
+        Log.d(TAG, "Logging feed event. Category is " + category);
+
+        getTracker(context).send(new HitBuilders.EventBuilder()
+                .setCategory(category)
+                .setAction("click")
+                .build());
+    }
+
+    private static Tracker getTracker(Context context) {
+        if (tracker == null) {
```
myleshorton: Is our context ever changing here? If it ever is, then we don't want to cache trackers and return the one for prior contexts. We could alternatively include a map of contexts->trackers, but how expensive are trackers really? Would it make more sense to just create a new one each time?

atavism: @myleshorton, we're passing in the application context. It's tied to the lifecycle of the entire application so we don't have to worry about it ever changing.

myleshorton: Also, it seems like ideally we'd adapt the code in https://github.com/getlantern/lantern/blob/devel/MobileSDK/sdk/src/main/java/org/lantern/mobilesdk/Lantern.java to be reusable for this @atavism

myleshorton: Sounds good on the application context @atavism. With the mobile SDK code, we are creating two analytics objects for the same tracking ID now though right or no? Seems like that could create some confusing data if so.

atavism: @myleshorton There is some overlap, but that analytics code is part of the SDK. I don't think it makes sense to intermix tracking events for Everfeed with it (no third-party that integrates Lantern would need any of it). So that's why I keep them separate.

atavism: @myleshorton Yes, but both trackers are logging different events entirely. With the tracker in the app itself, the only event we're currently recording is when a user clicks on an article (nothing related to new sessions or anything).

myleshorton: Like maybe make `StartResult` also include the Tracker or something:

https://github.com/getlantern/lantern/blob/devel/MobileSDK/sdk/src/main/java/org/lantern/mobilesdk/StartResult.java

atavism: Oh, that's a good idea actually. We could just have StartResult return the tracker already created. Then there's no need to have two. Ok, I'll make that change!

atavism: Oh, obtaining a ``StartResult`` though would (currently) necessitate starting Lantern (and we wouldn't want to make that a requirement). But I'll look into refactoring some of that code just to return a tracker

myleshorton: OK it's not a huge issue @atavism, especially if trackers definitely don't do anything until you tell them to, but just seemed like we could make it a little cleaner.

The other refactoring I can imagine would be to pass an analytics object into the SDK with another method. If the SDK gets that, then it would just use it instead.

atavism: @myleshorton No, I think it's a really good point. It's cleaner having only a single tracker.  What do you think of my latest changes?

oxtoacart:
@aranhoide @myleshorton Should this be going to release-2.2.0 instead of devel?

myleshorton:
OK LGTM @atavism!

PR #4027, opened 2016-04-20
fix #4019

### LINKED ISSUE #4019: Edge - The titles doesn't have the lower half of the lower case letters 'p','y','g'
image

fffw:
@myleshorton It makes feed title smaller to have room for the characters, without overflow. It's size in the original design, I just didn't correctly implement it.


myleshorton:
LGTM @fffw!

PR #4028, opened 2016-04-20
Updated eventual so that it doesn't need goroutines and therefore can…

oxtoacart:
…'t leak goroutines

@fffw I've separated this from the detour stuff since it's not strictly related.

### DIFF
--- a/src/github.com/getlantern/eventual/eventual.go
+++ b/src/github.com/getlantern/eventual/eventual.go
@@ -2,14 +2,10 @@
 package eventual
 
 import (
+	"math"
 	"sync"
 	"sync/atomic"
 	"time"
-)
-
-const (
-	FALSE = 0
-	TRUE  = 1
 )
 
 // Value is an eventual value, meaning that callers wishing to access the value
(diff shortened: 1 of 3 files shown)


fffw:
I'm wondering if we can use `sync.Cond` in place of slice of channels. Get() calls `cond.Wait()`, while Set/time.AfterFunc/Cancel calls `cond.Broadcast()`. It's unnecessary to pass the value through channel, as it can always be retrieved from `atomic.Value`.

Or, we can use single channel and always close it in `Set` and `Cancel`. All pending `Get` wait on it, and retrieve the actual value after `select`. Currently waiters are only notified once and get removed, single channel can do the same, but with less code and a little performance gain.

oxtoacart:
I don't think having time.AfterFunc call cond.Broadcast() because that'll make other waiters stop even if they have a longer timeout.

I'll try the single channel thing though.

oxtoacart:
Actually, a single channel won't work because we need to know how many waiters we have so that we send a message to each.  We could do a single channel plus an atomic counter of waiters, and then send a message for each waiter, but I don't know that this buys us much.

fffw:
Oh, I didn't think about it. LGTM!

PR #4032, opened 2016-04-20
switched link

myleshorton:


myleshorton:
I'm actually a little bit confused on this @fffw and @todd. I'm only seeing the `LANTERN_MOBILE_ANDROID` and `LANTERN_MOBILE_SHARE` keys for the en_US.json file. I swear I saw them earlier in the dist directory for other languages though. Either way, do we need to add those to the other locales?

myleshorton:
Any insights into why those other locales don't have those fields @fffw?

atavism:
@myleshorton We have similar strings on Android but for every locale already. Perhaps that's what you remember seeing? BTW, I substituted your ``bit.ly`` link with the one I was already using (on the pro branch).

fffw:
@myleshorton Those fields are in below files.
```
./locale/en_US.json
./locale/es.json
./locale/fa_IR.json
./locale/fr.json
./locale/zh_CN.json
```

Actually I think we should remove `LANTERN_MOBILE_ANDROID` from locale file, and change `LANTERN_MOBILE_SHARE` to a format string, as the URL itself doesn't need to be translated.

BTW: bit.ly was unblocked in 2010 in China, but the [greatfire test result](https://en.greatfire.org/bit.ly) is not that good.

fffw:
Hmm I guess you branched off from `devel`, as `LANTERN_MOBILE_ANDROID` only appears in `locale/en_US.json` in that branch. My previous grep is from release-2.2.0.

fffw:
Ok I see. I fetched Transifex translations in c9f4e6bd0f696fdc9a82b6337a5f4b891c5d6b1f, which is not merged to `devel` yet.

myleshorton:
yeah i was working on release-2.2.0 @fffw

myleshorton:
Weird -- I see the same thing you're seeing now @fffw. Not sure what happened there. I'll make those other changes as well.

myleshorton:
Never mind -- closing this!

fffw:
@myleshorton Did you see my PR for the same  issue ?

fffw:
Oh never mind I just saw you already merged it.

myleshorton:
@atavism can you put your bitly credentials in TMS?

Issue #4049, opened 2016-04-21
Mobile - Some descriptions of articles in Chinese show invalid characters


myleshorton:
For BBC Chinese, there is some decoding issues. In the abstract, some characters are not decoded correct and shown as a questions mark within a square. (It's not a Chinese character, thank you very much )

![image](https://cloud.githubusercontent.com/assets/1143966/14720560/df8abdfa-07b6-11e6-8707-d97ad7f59e10.png)

atavism:
This is an issue on the client side. I'm shortening article summaries to 150 characters, and that must be severing unicode characters towards the end.

PR #4059, opened 2016-04-22
Use content when meta description unavailable

atavism:
Whenever an article is missing a meta description tag, use the content entry instead. This also fixes https://github.com/getlantern/lantern/issues/4049


oxtoacart:
LGTM.  The build failure is unrelated to your changes.

Issue #4057, opened 2016-04-22
Mobile - Show a spinner while the feed loads


sunshine19090:
In my phone, it takes about 4-5 seconds when I quit and reopen lantern. Meanwhile, I am not even sure if it is going to load or not.

atavism:
Thanks, @sunshine19090. Good idea! Will add it now.

atavism:
Ok, I opened a PR for this [here](https://github.com/getlantern/lantern/pull/4064/files). 

This is what it looks like.. Look good to you, @Derekf5?

![image](https://cloud.githubusercontent.com/assets/6346213/14758154/ce0647de-08b0-11e6-8c2c-2f2d496a5d57.png)

Derekf5:
Yeah looks good @atavism the only think i would do is move the spinner to the middle of that section.

sunshine19090:
@Derek, For me it looks like the spinner is centered but the ON/OFF button is not. But am not sure as I couldn't take a proper pic when it appeared. Please check and close the issue if it looks good to you.


![screenshot_2016-04-27-10-33-22](https://cloud.githubusercontent.com/assets/11394369/14861854/cd0def4c-0c63-11e6-94ae-0dad6a5ec069.png)

atavism:
@sunshine19090 On/off looks off because I sent you a build without [this](https://github.com/getlantern/lantern/pull/4098) merged in. I will send you an updated version

sunshine19090:
Now it looks good @atavism. But I see spinner whenever I open the lantern app on top of the feeds text. The look of the spinner is not so good on top of the text. May be it is just because the spinner is big compared to the text in the background. Will it look better if the spinner is little small? 

Anyway moving this to P3 as this works fine. 

![screenshot_2016-04-27-12-01-26](https://cloud.githubusercontent.com/assets/11394369/14864727/0d178682-0c70-11e6-9981-648af5fb4b42.png)

myleshorton:
Is this one all set then @atavism and @sunshine19090?

oxtoacart:
At this point, the remaining requirement is that when reentering the app and the feeds are already loaded, we shouldn't show a spinner over the existing content.  We should however automatically reload the content as we do today, just not show the spinner.

myleshorton:
What's the status of this @atavism?

atavism:
[...]

PR #4064, opened 2016-04-23
Show progress dialog while the news feed loads

atavism:
This resolves https://github.com/getlantern/lantern/issues/4057

### DIFF
--- a/src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/model/GetFeed.java
+++ b/src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/model/GetFeed.java
@@ -25,6 +25,11 @@ public class GetFeed extends AsyncTask<String, Void, ArrayList<String>> {
         this.activity = activity;
         this.proxyAddr = proxyAddr;
         progressBar = (ProgressBar)activity.findViewById(R.id.progressBar);
+    }
+
+    @Override
+    protected void onPreExecute() {
+        super.onPreExecute();
         // show progress bar
         progressBar.setVisibility(View.VISIBLE);
     }
--- a/src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/model/GetFeed.java
+++ b/src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/model/GetFeed.java
@@ -25,14 +25,12 @@ public class GetFeed extends AsyncTask<String, Void, ArrayList<String>> {
         this.activity = activity;
         this.proxyAddr = proxyAddr;
         progressBar = (ProgressBar)activity.findViewById(R.id.progressBar);
+        // show progress bar
+        progressBar.setVisibility(View.VISIBLE);
     }
 
     @Override
     protected ArrayList<String> doInBackground(String... params) {
-
-        // show progress bar
-        progressBar.setVisibility(View.VISIBLE);
-
         String locale = Locale.getDefault().toString();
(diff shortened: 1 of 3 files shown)


src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/model/GetFeed.java
```diff
@@ -16,14 +19,20 @@
...
         this.activity = activity;
         this.proxyAddr = proxyAddr;
+        progressBar = (ProgressBar)activity.findViewById(R.id.progressBar);
     }
 
     @Override
     protected ArrayList<String> doInBackground(String... params) {
+
+        // show progress bar
+        progressBar.setVisibility(View.VISIBLE);
```
oxtoacart: Better to do this in preExecute so that it runs on the UI thread.

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/model/GetFeed.java
```diff
@@ -16,10 +19,14 @@
...
     private LanternMainActivity activity;
     private String proxyAddr = "";
+    private ProgressBar progressBar;
 
     public GetFeed(LanternMainActivity activity, String proxyAddr) {
         this.activity = activity;
         this.proxyAddr = proxyAddr;
+        progressBar = (ProgressBar)activity.findViewById(R.id.progressBar);
+        // show progress bar
+        progressBar.setVisibility(View.VISIBLE);
```
oxtoacart: FWIW, doing this in preExecute() makes sure it runs on the UI thread no matter how/when GetFeed is instantiated.

atavism: Ok, wasn't finished making the change yet.

atavism: That's where I moved it.

oxtoacart:
The progress indicator itself looks good, though it might look even better if centered within the feeds area rather than the screen.

Something unrelated to this change that I noticed is that if I go to another app and then come back to the Lantern app, it still has all of the feed data that was there before, but it loads the feeds again.  Without the loading indicator, I wouldn't have noticed this, but with the indicator it's quite noticeable.  Is this worth dealing with?  If so, should we log it separately or do you want to just tackle it as part of this PR?

atavism:
I do like the indicator appearing towards the top since that's where the feed headers show up. Is it aligned at the top of the container view for you or in an arbitrary spot on the screen? We could move it towards the center if that looks nicer.

That's actually semi-intentional since we don't have a refresh button. From a usability standpoint, if I leave the app running but navigate elsewhere, I like automatically seeing the latest content whenever I re-open it. I think it's OK too that the progress dialog appears on top of the previous content, since it lets the user know we're currently refreshing the feed. That said, I think we can leave it for now but ideally we'd add a refresh button to a future version which would eliminate the need for it. Does that sound good?

atavism:
Maybe this is something you want to chime in on, @Derekf5? Here's what it looks like now.. 
image

atavism:
Oh, just saw his feedback. He also recommended moving it to the middle of the screen. I'll make that fix now!

atavism:
Ok, I moved the progress dialog down towards the center of that section, @oxtoacart. Think it's in an ideal spot now.

oxtoacart:
Agreed on the behavior when reopening, new position of progress indicator looks great!

Issue #4022, opened 2016-04-19
Windows 7, all browsers - When Manage System proxy is turned off and if the sites are blocked, then lantern shows only the spinner and not loading the page


sunshine19090:
![image](https://cloud.githubusercontent.com/assets/11394369/14654475/bb06425e-0632-11e6-9145-17c431ee184e.png)

sunshine19090:
Works fine in windows 10 - able to see the news feed.

fffw:
I think it will fallback to bookmarks eventually but need to confirm . We probably only need to show alarm on home screen if Manage system proxy is off, so user know what caused trouble

sunshine19090:
I agree @fffw

myleshorton:
We'll handle this one with #4024

PR #4066, opened 2016-04-23
Show error page when feed fails to load fix #4024

### LINKED ISSUE #4024: If Lantern's not working, the newsfeed will never load

fffw:
@myleshorton I made a few changes based on your work.


src/github.com/getlantern/lantern-ui/app/partials/error.html
```diff
@@ -0,0 +1,46 @@
+<div id="error" ng-controller="ErrorCtrl" ng-show="showError">
+  <div id="oops">
+    {{ 'ERROR_OOPS' | translate }}
+  </div>
+  <div id="error-description">
+    {{ 'ERROR_DESCRIPTION' | translate }}
+  </div>
+  <div id="error-reasons">
+    <ol>
+      <li ng-show="!model.settings.systemProxy">
```
fffw: Related to https://github.com/getlantern/lantern/issues/4022

src/github.com/getlantern/lantern-ui/app/partials/error.html
```diff
@@ -0,0 +1,46 @@
...
+        <div class="error-action ng-hide" ng-show="showXunleiHelp">
+          {{ 'XUNLEI_HELP' | translate }}
+        </div>
+      </li>
+      <li>
+        <div>{{ 'ANOTHER_ISSUE' | translate }}</div>
+        <div ng-show="!showConnectionHelp"><a ng-click="toggleShowConnectionHelp()">{{ 'WHAT_DO' | translate }}</a></div>
+        <div class="error-action ng-hide" ng-show="showConnectionHelp">
+          <div>{{ 'CONNECTION_HELP' | translate }}</div>
+          <div>Windows Vista/7/8/10: C:\Users\USERNAME\AppData\Roaming\Lantern\logs</div>
```
fffw: It seems too detailed to show path names on this page. Can we just add the link to https://github.com/getlantern/lantern/wiki/Emailing-lantern-logs?

src/github.com/getlantern/lantern-ui/app/angular-feed/scripts/services.js
```diff
@@ -37,7 +37,8 @@ angular.module('feeds-services', []).factory('feedService', ['$q', '$http', '$sc
           deferred.reject(new Error("invalid data format"));
           return
         }
-        feedCache.set(feedURL, data);
+        // Always get fresh copy, in hope to be more reactive to network errors
+        // feedCache.set(feedURL, data);
```
fffw: It's still not unclear to me if we should keep the 5 minutes caching or not. Occasionally it took me around 30 seconds to get a response, even when the response is "304 Not modified". Caching would help for frequent refreshes.

src/github.com/getlantern/lantern-ui/app/partials/error.html
myleshorton: Yeah that was the original design. I could go either way @fffw. The other thing I was playing with was actually including the links in the translated version but didn't quite get it working.

fffw: @myleshorton I don't quite understand. Do you mean to set up different link for each translation?

myleshorton: No, just that we need "Facebook", "GitHub", and "Forums" to actually link to those places, but we're linking from translated text.

src/github.com/getlantern/lantern-ui/app/angular-feed/scripts/services.js
myleshorton: Sounds like maybe we should keep it then? Very odd it could sometimes take so long. Is your proxy generally slow at times when you've seen it take a long time to load the feed?

fffw: It doesn't happen to me now, could be caused by network glitches. It may take a bit longer to load feeds when Lantern starts up and old config doesn't work any more.

One possible way is to load cached feeds first, fetch feeds in background and replace it when succeeded, or show error page if failed. Not sure if I have a chance to make that change today.

src/github.com/getlantern/lantern-ui/app/partials/error.html
fffw: Ok I see. I can do it the same way we share mobile download link with different languages.

src/github.com/getlantern/lantern-ui/app/angular-feed/scripts/services.js
oxtoacart: +1 on caching on the client

cc @xiam

myleshorton:
Yeah aside from the links in the translations @fffw this seems good to go to me. The only thing I haven't been able to verify is that the on-error handler actually gets called when there's an error. When I just changed it to some bogus URL, it actually did not get called, but maybe there's a better way. Have you seen the on-error handler actually get called?

src/github.com/getlantern/lantern-ui/app/partials/error.html
myleshorton: Honestly I think there are also cleaner ways to do it with angular translate but maybe I'm wrong. I actually thought your `$sce.trustAsHtml` from services.js might do the trick no?

src/github.com/getlantern/lantern-ui/app/partials/error.html
myleshorton: I don't think we want this @fffw, as this is a very common issue with the system proxy on or off -- i.e. some issue with extensions right?

myleshorton: Oh hey sorry I misread this @fffw -- looks good!

myleshorton:
OK @fffw I just added the links and also threw in OS-detection so we can tell the user exactly where their logs are. Also removed d3 while in there.

myleshorton:
Hey @fffw I know you're on break, but you want to give this a look when you're back? Thanks!

fffw:
@myleshorton Device OS detection is a cool idea! 

I usually trigger error page by turning off Wifi temporarily. When changing to bogus URL, both `feedUrl()` and `fallback-url` need to be changed.

I added the logic to show cached feeds first if any, and fetch server in the background. It will re-render feeds after a successful fetching, or render the error page. I also adjusted the cache from 5 minutes to 24 hours, so users don't need to wait for the feeds to load in most cases.

myleshorton:
This all looks good to me @fffw! Merging.

fffw:
Thanks @myleshorton !

PR #4075, opened 2016-04-25
Support building development lantern on Windows

oxtoacart:
The packaging stuff won't work yet, but at least I can build the `lantern` target and run Lantern.


Makefile
```diff
@@ -200,13 +200,18 @@ $(RESOURCES_DOT_GO): $(NPM)
...
+	rm -f bin/tarfs && \
 	go install github.com/getlantern/tarfs/tarfs && \
+	mv bin/tarfs.exe bin/tarfs ; \
 	echo "// +build !stub" > $$DEST && \
 	echo " " >> $$DEST && \
-	tarfs -pkg ui $$DIST >> $$DEST && \
+	bin/tarfs -pkg ui $$DIST >> $$DEST
+
+# Generates a syso file that embeds an icon for the Windows executable
+generate-windows-icon:
```
oxtoacart: The syso is now back in git so I don't need to run this on Windows, which is good because it doesn't run.

xiam:
Gonna take this one, looks like we have a conflict @oxtoacart, could you rebase and push again?

oxtoacart:
@xiam Merged

oxtoacart:
@xiam I mean, I reverse merged to resolve the conflict :)

xiam:
Hey @oxtoacart, looks good, just made a few changes. What are the name changes required for?

oxtoacart:
@xiam By "name changes" do you mean all the moved files in lantern-ui?  With lantern-ui in the original location, the directory structure is too deep for Windows which has a maximum path length of 260.

xiam:
Yeah, that's it. Hmm, not sure if that justifies changing the path for the
whole project. Isn't it possible to use soft links or equivalent? I'll take
a look too trying to find another solution.

oxtoacart:
@xiam I only moved lantern-ui.  It's actually kind of strange to have it in the GOPATH in the first place, and I used git mv to move it, so it shouldn't cause any merge issues.

myleshorton:
Yeah I agree @oxtoacart it's a little weird to have lantern-ui in the GOPATH.

xiam:
Ok then @myleshorton & @oxtoacart, I think that having our source code on `src` makes more sense, but if you both agree let's leave it like that. Otherwise the build works for me.

oxtoacart:
@xiam thanks for the review and improving the Makefile!

myleshorton:
Nice @xiam!!

PR #4076, opened 2016-04-25
Update build tools revision and Readme (Android)

uaalto:
In order to be able to compile Lantern for Android, I had to specify the build tools revision to the latest.

Unfortunately I couldn't find a method to wildcard the build tools revision so it doesn't need to be updated manually. I also cannot find revision 2, so I had no other option than to update. In case you had it working already, this steps for updating the revision might be of reference.

`android list sdk -a`

Shows a list:

1- Android SDK Tools, revision 24.0.2
2- Android SDK Platform-tools, revision 21
3- Android SDK Build-tools, revision 21.1.2
...
Select using ID for build-tools 23.0.3

`android update sdk -a -u -t 3`

### DIFF
--- a/MobileSDK/sdk/build.gradle
+++ b/MobileSDK/sdk/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'android-library'
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "23.0.2"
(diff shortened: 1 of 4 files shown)


atavism:
Thanks, @uaalto! LGTM.

PR #4132, opened 2016-04-28
Update share links

atavism:


atavism:
@myleshorton This substitutes the APK link with ``https://bit.ly/lanternapk``

myleshorton:
Wow -- that's actually pretty crazy. Those were extremely old links right @atavism? Either way -- good change!

PR #4138, opened 2016-04-28
Added else to only load the feed if not in the cache closes #4108

### LINKED ISSUE #4108: Articles in Chinese feed are mixed up
I see Jandan.net, theinitium.com, chinadigitaltimes.com under a single channel and all other channels are also mixed up.

image

myleshorton:
I also reduced the cache time.


myleshorton:
I'm gonna pull this in so we get a new build out today.

fffw:
@myleshorton Actually I'm not sure if it's the cause of #4108. `render()` always replaces objects in `$scope` each time so old items won't mess with new one. I do think background fetching is important to provide better UX and reveal connectivity issue.

myleshorton:
Ok @fffw if you find a different cause feel free to change it!

Issue #4139, opened 2016-04-28
Mobile - Feeds should refresh automatically when it detects internet connection


sunshine19090:
I got the message "OOPs Error fetching feeds" when there is no internet connection but it isn't going away until I manually refresh after connecting to my wifi. It doesn't refresh even after switching on lantern.

sunshine19090:
@atavism I don't see any change. I still had to refresh manually when the network connection is back. However, it works fine in the case of detecting no network connection and shows the error message.

sunshine19090:
yep, this works fine now @atavism

PR #4146, opened 2016-04-28
Automatically refresh feed when connectivity is detected

atavism:
This resolves https://github.com/getlantern/lantern/issues/4139

Whenever a user switches WiFi on, the feed is automatically refreshed; whenever WiFi is switched off, the feed error view is immediately shown.


src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/activity/LanternMainActivity.java
```diff
@@ -397,13 +404,21 @@ public void onReceive(Context context, Intent intent) {
...
             } else if (action.equals(ConnectivityManager.CONNECTIVITY_ACTION)) {
-                // whenever a user disconnects from Wifi and Lantern is running
                 NetworkInfo networkInfo =
                     intent.getParcelableExtra(ConnectivityManager.EXTRA_NETWORK_INFO);
-                if (LanternUI.useVpn() && 
-                        networkInfo.getType() == ConnectivityManager.TYPE_WIFI &&
-                        !networkInfo.isConnected()) {
-                    stopLantern();
+
+                if (networkInfo.getType() == ConnectivityManager.TYPE_WIFI) {
```
myleshorton: What about other connectivity types here @atavism? Like `TYPE_ETHERNET` or even `TYPE_MOBILE`?

atavism: Well, we could refresh the feed whenever we get a connection regardless of the type. I'm just wondering, given how flaky and slow mobile networks are, if we want to avoid fetching the feed every time a connection is available. That's why I limited it to WiFi.

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/activity/LanternMainActivity.java
```diff
@@ -397,13 +404,21 @@ public void onReceive(Context context, Intent intent) {
...
+                    if (networkInfo.isConnected()) {
+                        // automatically refresh feed when connectivity is detected
+                        refreshFeed(null);
+                    } else {
+                        // whenever a user disconnects from Wifi and Lantern is running
```
oxtoacart: If I'm on a mobile network but disconnect from wifi, this will display the error page, right?  I don't think we want that.

atavism: Oh, you're saying if we disconnect, it's preferable being able to browse the previously loaded feed instead of seeing the error page right away? Yeah, good point. Let me just remove that line

oxtoacart: That too, but what I was actually saying is that if we disconnect from Wifi but are still connected to a cellular network, we should be allowed to continue browsing as well.

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/activity/LanternMainActivity.java
myleshorton: OK I think that makes sense @atavism

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/activity/LanternMainActivity.java
atavism: Ok, we aren't showing an error page whenever we disconnect or switch networks anymore. Only time the feed is automatically refreshed is when we connect to WiFi

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/activity/LanternMainActivity.java
atavism: Ok, this should be updated to the latest from ``release-2.2.0`` now

myleshorton:
LFTM @atavism! Merging.

Issue #4153, opened 2016-04-28
Sort feeds on mobile


atavism:
Instead of ordering the feeds alphabetically, a new ``sorted_feeds`` field specifies the order the feeds should appear in.

PR #4154, opened 2016-04-28
Sort feeds on mobile

atavism:
This resolves https://github.com/getlantern/lantern/issues/4153 and it also a fixes an issue where we were using "all" instead of the translated string resource for the feed containing every article.

### DIFF
--- a/src/github.com/getlantern/lantern/feed.go
+++ b/src/github.com/getlantern/lantern/feed.go
@@ -36,6 +36,7 @@ type Feed struct {
 	Feeds   map[string]Source    `json:"feeds"`
 	Entries FeedItems            `json:"entries"`
 	Items   map[string]FeedItems `json:"-"`
+	Sorted  []string             `json:"sorted_feeds"`
 }
 
 // Source represents a feed authority,
@@ -99,7 +100,7 @@ func handleError(err error) {
 // Lantern public feed for displaying on the home screen.
 // If a proxyAddr is specified, the http.Client will proxy
 // through it
-func GetFeed(locale string, proxyAddr string, provider FeedProvider) {
+func GetFeed(locale string, all string, proxyAddr string, provider FeedProvider) {
 	var err error
 	var req *http.Request
 	var res *http.Response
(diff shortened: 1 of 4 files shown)


src/github.com/getlantern/lantern/feed.go
```diff
@@ -199,10 +184,27 @@ func processFeed(provider FeedProvider) {
 		if desc == "" {
 			desc = entry.Content
 		}
-
 		feed.Entries[i].Description = desc
 	}
 
+	// the 'all' tab contains every article that's not associated with an
```
oxtoacart: @atavism Ah, I see my mistake now.  I didn't realize that we were using value types here.  If we switch to using pointers, we can avoid some unnecessary copying.  Should I give it a whirl?

src/github.com/getlantern/lantern/feed.go
```diff
@@ -199,10 +184,27 @@ func processFeed(provider FeedProvider) {
...
+			all = append(all, entry)
+		}
+	}
+	feed.Items[allStr] = all
+
+	// Get a list of feed sources & send those back to the UI
+	for _, s := range feed.Sorted {
+		if entry, ok := feed.Feeds[s]; ok {
+			log.Debugf("Adding feed source: %s", entry.Title)
+			provider.AddSource(entry.Title)
```
myleshorton: Since this relies on external data, maybe just make sure `entry.Title` isn't null here @atavism like you were before?

src/github.com/getlantern/lantern/feed.go
```diff
@@ -199,10 +184,27 @@ func processFeed(provider FeedProvider) {
...
+		}
```
myleshorton: Also would be useful to log an error here if we are trying to find a feed that's not in the map (would again indicate a pretty big issue with the feed).

src/github.com/getlantern/lantern/feed.go
myleshorton: Should probably log an error in the above case too!

src/github.com/getlantern/lantern/feed.go
oxtoacart: I went ahead and just did it - https://github.com/getlantern/lantern/pull/4157

src/github.com/getlantern/lantern/feed.go
atavism: @myleshorton Logging both of those errors, and I'm checking for a missing title now

src/github.com/getlantern/lantern/feed.go
atavism: Thanks, @oxtoacart!

src/github.com/getlantern/lantern/feed.go
```diff
@@ -160,35 +163,17 @@ func GetFeed(locale string, proxyAddr string, provider FeedProvider) {
...
 	}
 
-	processFeed(provider)
+	processFeed(allStr, provider)
 }
 
 // processFeed is used after a feed has been downloaded
 // to extract feed sources and items for processing.
-func processFeed(provider FeedProvider) {
+func processFeed(allStr string, provider FeedProvider) {
```
oxtoacart: @atavism Why did allStr become a parameter?  Is it ever anything other than "all"

atavism: Yes... It's the translated text now.

oxtoacart: Interesting.  Are the other things in feed.Items keyed to translated strings as well, or are they keyed to the feed names?

atavism: We are using the feed names we get back instead of a translated string for everything except "all"

oxtoacart: Oh, I see - the feed names are effectively already "translated" because they are what they are.  Thanks.

oxtoacart:
LGTM and works on my Nexus 4.

PR #4157, opened 2016-04-29
Using pointers for Source and FeedItem

oxtoacart:
@atavism I went ahead and just did this.  Test still passes, plus I added a test for the description issue that I had introduced.


atavism:
LGTM, thanks @oxtoacart!

PR #4162, opened 2016-04-29
Made feed test independent of network closes issue4114

oxtoacart:
cc: @atavism 

For anything that runs as part of our Travis build, I think it's best for it to be completely independent of the internet and our server infrastructure.  So, I've changed this test so that it serves up its own feeds (which I just grabbed from prod a few minutes ago).


src/github.com/getlantern/lantern/feed.go
```diff
@@ -157,6 +162,7 @@ func GetFeed(locale string, allStr string, proxyAddr string,
 		return
 	}
 
+	ioutil.WriteFile("/tmp/"+locale, contents, 0644)
```
atavism: If we wanted to write the feed contents to a temporary file on Android, we would want to store it in the ``getCacheDir()`` directory instead of /tmp. If this is just for testing purposes, do we want to only call this if ``runtime.GOOS != 'android'``?

oxtoacart: Oh, good catch!  That was just temporary to save off the test files, I'll remove.

myleshorton:
LGTM @oxtoacart!

PR #4167, opened 2016-04-29
Added fetching the feed url from the local client closes #4163

### LINKED ISSUE #4163: Support setting the feed by region as well as locale
Unfortunately many users in different regions don't have their browsers set to the expected locale. Very few users in Iran have their systems set to fa_IR, for example, and most users in Malaysia also have en_US set as their locale.

We can use the country detection within Lantern to override the locale setting to get around this.

myleshorton:


src/github.com/getlantern/flashlight/feedUrl.go
```diff
@@ -0,0 +1,33 @@
...
+func GetFeedURL(backupLocale string) string {
+	var locale = backupLocale
+	country := geolookup.GetCountry(time.Duration(10) * time.Second)
+	if country == "" {
+		// This means the country lookup failed, so just use whatever the default is.
+		log.Debug("Could not lookup country")
+		locale = backupLocale
+	} else if strings.EqualFold("ir", country) {
+		locale = "fa_IR"
+	}
```
atavism: Do we also want to check for China and Malaysia? Actually, while using a combination of the locale and country makes sense for determining the default feed to display, maybe to avoid this weird country to locale mapping, we could start publishing region-based feeds (``https://feeds.getiantem.org/<country>/feed.json``) also?

myleshorton: Yeah basically the two regions where the feed we want differs from the locale are Iran and Malaysia. Everywhere else the locale should signify the feed we want to show. Right now I don't think we have a Malaysia feed though right @xiam?

myleshorton: In Iran's case, since some people do have Farsi set on their systems, we probably want to keep supporting that as well. So I think just normalizing on the full locale is probably ok.

xiam: Yes, we do have one at https://feeds.getiantem.org/ms_MY/feed.json

src/github.com/getlantern/flashlight/ui/ui.go
```diff
@@ -97,7 +98,18 @@ func Start(requestedAddr string, allowRemote bool, extUrl string) (string, error
 		}
 		resp.WriteHeader(http.StatusOK)
 	}
+
+	// We use the backend to detect the user's country and redirect the browser
+	// to the correct URL that will itself be proxied over Lantern.
+	feedHandler := func(resp http.ResponseWriter, req *http.Request) {
+		vals := req.URL.Query()
+		defaultLang := vals.Get("lang")
+		url := feed.GetFeedURL(defaultLang)
```
atavism: It's possible ``defaultLang`` could be missing here, right? If so, ``determineLocale`` doesn't currently account for the empty string case.

myleshorton: OK @atavism I just added support for that within GetFeedURL and also added support for Malay when we detect the user's in Malaysia.

atavism:
LGTM now! Merging

myleshorton:
Thanks @atavism!

On Sun, May 1, 2016 at 10:04 AM, atavism <notifications@github.com> wrote:

> LGTM now! Merging
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/getlantern/lantern/pull/4167#issuecomment-216055677>
>



-- 
--
President
https://www.getlantern.org

Issue #4136, opened 2016-04-28
Mobile - BBC chinese feed is blank


sunshine19090:
It is working fine in desktop.

atavism:
It's showing up for me. It's just totally blank for you?

![image](https://cloud.githubusercontent.com/assets/6346213/14898187/71cc471e-0d3a-11e6-9786-412b3d2692d7.png)

sunshine19090:
It was like in the screenshot. I came across the same issue when I changed the language back to English feed but it went away after closing and reopening the app. 

![screenshot_2016-04-28-12-05-14](https://cloud.githubusercontent.com/assets/11394369/14898289/df3ad4b4-0d3a-11e6-9742-02f3b3c59b80.png)

sunshine19090:
@atavism I was trying to reproduce in the chinese feed and this time all were blank except for one feed at the end. Also, I didn't see the spinner when it took time to load. After quitting and reopening, I was able to see them again and the spinner too.

myleshorton:
Just a guess, but maybe just switching the language itself doesn't trigger a reload of the feed @atavism? If so that would be more of an issue in testing environments than in real world use.

myleshorton:
Is this one still an issue @sunshine19090 and @atavism?

atavism:
@myleshorton I opened a PR for this here: https://github.com/getlantern/lantern/pull/4176/files

sunshine19090:
This is good now.

PR #4176, opened 2016-05-02
Issue 4136

atavism:
This resolves https://github.com/getlantern/lantern/issues/4136 

With Lantern running, if you change your device locale and bring the app back to the foreground, the feed in the new locale will automatically load.

This updates the feed ViewPager to use a FragmentStatePagerItemAdapter instead of a FragmentPagerItemAdapter. The former allows us to preserve state and dynamically update feed headers in the tab layout.


src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
```diff
@@ -31,29 +31,22 @@
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        this.mFeedItems = new ArrayList<FeedItem>();
```
myleshorton: Doesn't this still need to be a synchronized list @atavism?

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
```diff
@@ -62,28 +55,48 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
...
+                }
+            });
+
+            return items;
+        }
+
+        @Override
+        protected void onPostExecute(List<FeedItem> items) {
+            super.onPostExecute(items);
+            mFeedItems.clear();
```
myleshorton: I believe you need to separately synchronize clear and addAll here @atavism.

myleshorton: Or potentially use a copyOnWriteArrayList

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
atavism: No. We only add elements to the feed items list now in ``onPostExecute`` of the AsyncTask (this is invoked on the UI thread).

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
atavism: It doesn't need to be synchronized any longer (that's why I moved the processing inside an AsyncTask). It happens on the UI thread now.

myleshorton: Ah cool I see that now - nice @atavism.

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
```diff
@@ -31,29 +31,22 @@
...
-                Log.d(TAG, String.format("Feed %s has %d items", feedName, 
-                            mFeedItems.size()));
-                if (adapter != null) {
-                    adapter.notifyDataSetChanged(); 
-                }
-            }
-        });
-    }   
+    public void setFeedName(String feedName) {
+        this.feedName = feedName;
```
myleshorton: This is called automatically by the JSON parser or something @atavism is that right?

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
```diff
@@ -62,28 +55,48 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
...
-                            String image, String url) {
-                        items.add(new FeedItem(title, desc, image, url));
-                    }
-                });
-
-                NotifyDataSetChanged(items);
-            }
-        }.start();
+        Log.d(TAG, "onViewCreated for " + this.feedName);
+        new LoadFeed().execute(this.feedName);
```
myleshorton: I'm worried about the feedName being potentially null or "" here @atavism. Is that possible?

Also, a unit test would be pretty amazing here.

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
```diff
@@ -62,28 +55,48 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
         View view = inflater.inflate(R.layout.feed_fragment, container, false);
 
         mList = (ListView)view.findViewById(R.id.feed);
-        adapter = new FeedAdapter(getActivity(), mFeedItems);
         mList.setAdapter(adapter);
         return view;
     }
 
+    protected class LoadFeed extends AsyncTask<String, Void, List<FeedItem>> {
```
myleshorton: Seems like a private class would make more sense here @atavism -- i.e. I don't think we're ever going to subclass this and nothing else in this package will user, or maybe more precisely I don't think we want to allow either of those to happen.

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
```diff
@@ -31,29 +31,22 @@
...
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        this.mFeedItems = new ArrayList<FeedItem>();
+        this.adapter = new FeedAdapter(getActivity(), mFeedItems);
 
         Bundle bundle = getArguments();
-        this.feedName = bundle.getString("name");
-        this.mFeedItems = Collections.synchronizedList(new ArrayList<FeedItem>());
+        if (bundle != null) {
+            this.feedName = bundle.getString("name");
```
myleshorton: Yeah I'm worried here about the feed name never getting set. Ideally to me this would be an invariant of the class -- i.e. make feedName `final` and check any arguments assigning it for validity.

myleshorton: And ditch `setFeedName` if possible -- make it immutable.

myleshorton:
Hey @atavism I feel like this one actually isn't critical to 2.2.0, but please correct me if I'm wrong. I'm going ahead and building a new 2.2.0 now, but maybe we can just shift this to 2.2.1 or something.

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
```diff
@@ -24,66 +24,79 @@
...
 
     private FeedAdapter adapter;
-    private String feedName;
+    private final String feedName;
     private ListView mList;
     private List<FeedItem> mFeedItems;
 
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
```
atavism: Whenever the feed fragment is created, we're always passing in a bundle with a name argument. There's also multiple checks before that to make sure we have a valid name. That said, I'm returning an empty list of feed items now whenever we encounter a null or empty string name. I also removed ``setFeedName`` (wasn't being used anyway).

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
atavism: Whenever the feed fragment is created, we're always passing in a bundle with a name argument. There's also multiple checks before that to make sure we have a valid name. That said, I'm returning an empty list of feed items now whenever we encounter a null or empty string name. I also removed setFeedName (wasn't being used anyway).

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
atavism: Agreed. No reason for the class not to be made it private. Should be updated now!

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
atavism: Yeah, definitely agree we should add a unit test. I'm actually working on improving test coverage overall on a separate branch so I made a note of it.

atavism:
@myleshorton Not super critical, but without this, we can't change the device locale to load different feeds without restarting the app. I addressed your comments and would say we should try to get it in if we can.

myleshorton:
OK LGTM @atavism!

atavism:
Thanks, @myleshorton!

PR #4181, opened 2016-05-02
Issue 4109

atavism:
This fixes a tiny discrepancy between desktop and mobile. Whenever an article meta description is missing, we use ``contentSnippetText`` instead of ``contentText`` now. The former field is the one updated by the feed server already to provide article descriptions for sites like Reddit.


myleshorton:
OK LGTM @atavism!

PR #4198, opened 2016-05-03
Lots of tweaks to not show the newsfeed by default closes #4185

### LINKED ISSUE #4185: Add link to newsfeed
Instead of throwing all users into the newsfeed, we'll add a link allowing them to try it out.

The main intent here is to avoid automatically showing content that some users might consider to be political in nature, which might make Lantern a bigger target for censors.

So instead, people will have to click on a link to open the newsfeed.

The UI should remember if someone has previously elected to view the newsfeed and automatically show it the next time they come into Lantern.

This change only applies to desktop.  Mobile will stay as is (we consider it private enough).

myleshorton:


myleshorton:
Hey @fffw want to take a look at this one?

fffw:
1. Should not show the box if the mobile ad is in place, or, clicking feedLink should hide the mobile ad.

2. Suggest to make "Take me back to the old look" less obvious. Commented at https://github.com/getlantern/lantern/issues/4185#issuecomment-216714742

fffw:
@myleshorton That's all I found

fffw:
LGTM! The #2 suggest is worth another PR.

myleshorton:
Thanks @fffw!

PR #4199, opened 2016-05-03
Added both hash and version reporting closes #4164

### LINKED ISSUE #4164: Add hash of binary to analytics
This will help us verify if a lot of the Lantern users we're seeing in analytics are using versions we've built.

myleshorton:
Might as well get this into 2.2.0 so we have this data sooner in GA!

### DIFF
--- a/src/github.com/getlantern/flashlight/analytics/analytics.go
+++ b/src/github.com/getlantern/flashlight/analytics/analytics.go
@@ -15,13 +15,16 @@ import (
 	"github.com/getlantern/flashlight/config"
 	"github.com/getlantern/flashlight/geolookup"
 	"github.com/getlantern/flashlight/util"
+	"github.com/kardianos/osext"
 
 	"github.com/getlantern/golog"
 )
 
 const (
-	trackingId  = "UA-21815217-12"
-	ApiEndpoint = `https://ssl.google-analytics.com/collect`
+	trackingID = "UA-21815217-12"
+
+	// endpoint is the endpoint to report GA data to.
+	endpoint = `https://ssl.google-analytics.com/collect`
 )
 
 var (
@@ -30,6 +33,7 @@ var (
 	maxWaitForIP = math.MaxInt32 * time.Second
 )
 
+// Start starts the GA session with the given data.
 func Start(cfg *config.Config, version string) func() {
 	var addr atomic.Value
 	go func() {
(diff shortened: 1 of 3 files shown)


myleshorton:
Note adding the version was already in there but we didn't actually have that custom dimension set on the server -- i added that.

myleshorton:
You mind giving this one a look too @fffw?

src/github.com/getlantern/flashlight/util/hash.go
```diff
@@ -0,0 +1,28 @@
...
+	"io"
+	"os"
+)
+
+// GetFileHash returns the hex encoding of the sha-256 hash of the
+// file at the specified path.
+func GetFileHash(path string) (string, error) {
+	log.Debugf("Hashing file at path %v", path)
+	if f, err := os.Open(path); err != nil {
+		log.Errorf("test %v", err)
```
fffw: Consider more accurate error string?

myleshorton: Ha yes good catch -- fixed!

src/github.com/getlantern/flashlight/analytics/analytics.go
```diff
@@ -73,25 +77,44 @@ func sessionVals(ip, version, clientId, sc string) string {
...
 	vals.Add("sc", sc)
 	return vals.Encode()
 }
 
-func endSession(ip string, version string, proxyAddrFN eventual.Getter, clientId string) {
-	args := sessionVals(ip, version, clientId, "end")
+// GetExecutableHash returns the hash of the currently running executable.
+// If there's an error getting the hash, this returns
+func getExecutableHash() string {
+	if lanternPath, err := osext.Executable(); err != nil {
```
fffw: Seems `osext` doesn't support Android. It doesn't halt the program anyway, but would be good to make it explicit.

fffw: Oh actually Android is Linux, but I don't know what it will get because Lantern is not an executable on Android.

myleshorton: Good point. That analytics code doesn't actually run on Android but we
should guard against it.

myleshorton:
Hmnn...that android issue is actually significant. Right now we're doing analytics from the java side there, so it might make sense to make it a different issue.

myleshorton:
OK -- I just guard against android for now @fffw. What do you think?

fffw:
LGTM! Waiting for CI

PR #4205, opened 2016-05-04
Added user agent handling to analytics closes #4046

### LINKED ISSUE #4046: Google Analytics Data for Lantern is missing OS and other client information
Google Analytics gets this information from the user-agent, and our Go-based library populates a generic user agent that doesn't provide useful information to Google Analytics.

Things that would be useful to capture:

- Operating System
- Browser (although a user could be using multiple browsers in one session)
- Lantern Version

cc: @myleshorton

myleshorton:


src/github.com/getlantern/flashlight/logging/logging_test.go
```diff
@@ -7,12 +7,36 @@ import (
...
+	userAgent := eventual.NewValue()
+
+	listener := func(a string) {
+		userAgent.Set(a)
+	}
+	AddUserAgentListener(listener)
+	RegisterUserAgent(agent)
+
+	// Just register again to hit the code path for duplicates.
+	RegisterUserAgent(agent)
```
uaalto: In order to test eventual assignation (the purpose of eventual), perhaps we can run `RegisterUserAgent` in a different goroutine that waits for couple hundred miliseconds to register the agent, so `received` needs to block for that time.
These can be two separate tests though. One with immediately available value and one in which eventual blocks.

uaalto:
Except for that comment for a potential improvement in the tests, everything LGTM @myleshorton 

I also wanted to ask, the issue stems from sending analytics with no known user agent? (I had no background of this issue, but that's what I understand from the code).

src/github.com/getlantern/flashlight/logging/logging_test.go
myleshorton: Yeah that makes sense. I think arguably that should be covered in the tests for `eventual` itself, but no harm in adding it here. I'll do it!

myleshorton:
Thanks @uaalto yeah GA actually uses the user agent to determine a lot of stuff, particularly the browser and the operating system. So without it we don't get browser and OS data from the backend analytics.

myleshorton:
OK @uaalto I finally made those changes and added an analytics test! Will comment more in the code.

src/github.com/getlantern/flashlight/analytics/analytics.go
```diff
@@ -34,47 +33,61 @@ var (
...
 	go func() {
 		logging.AddUserAgentListener(func(agent string) {
 			userAgent.Set(agent)
 		})
-		ip := geolookup.GetIP(maxWaitForIP)
+		ip := ipFunc(maxWaitForIP)
 		if ip == "" {
-			log.Errorf("No IP found within %v, not starting analytics session", maxWaitForIP)
-			return
+			log.Errorf("No IP found within %v", maxWaitForIP)
```
myleshorton: We used to not report any analytics if we couldn't resolve the IP, which doesn't really make sense because there's lots of other useful info to report. So now we just keep going and don't report the IP if we don't know it.

src/github.com/getlantern/flashlight/analytics/analytics.go
```diff
@@ -34,47 +33,61 @@ var (
...
 )
 
 // Start starts the GA session with the given data.
-func Start(cfg *config.Config, version string) func() {
+func Start(deviceID, version string) func() {
+	return start(deviceID, version, geolookup.GetIP, maxWaitForUserAgent, trackSession)
+}
+
+// start starts the GA session with the given data.
+func start(deviceID, version string, ipFunc func(time.Duration) string, uaWait time.Duration,
```
myleshorton: This is just for testability

src/github.com/getlantern/flashlight/logging/logging_test.go
```diff
@@ -23,12 +23,17 @@ func TestUserAgent(t *testing.T) {
...
-	// Just register again to hit the code path for duplicates.
+	// Do an initial register just to test the duplicate agent paths.
 	RegisterUserAgent(agent)
 
+	AddUserAgentListener(listener)
+
+	go func() {
+		time.Sleep(200 * time.Millisecond)
+		RegisterUserAgent(agent)
+	}()
```
myleshorton: Here's the change to test that setting the user agent asynchronously works fine @uaalto

src/github.com/getlantern/flashlight/main/main.go
```diff
@@ -236,7 +236,7 @@ func beforeStart(cfg *config.Config) bool {
 
 	// Only run analytics once on startup.
 	if settings.IsAutoReport() {
-		stopAnalytics := analytics.Start(cfg, flashlight.Version)
+		stopAnalytics := analytics.Start(cfg.Client.DeviceID, flashlight.Version)
```
myleshorton: No need passing the whole config when we just need a string -- makes testing easier.

uaalto: Saw that above. More clear.

src/github.com/getlantern/flashlight/logging/logging_test.go
uaalto: Great! Thank you for updating

src/github.com/getlantern/flashlight/analytics/analytics_test.go
```diff
@@ -0,0 +1,30 @@
...
+		return "127.0.0.1"
+	}, 1*time.Millisecond, func(args string, addr eventual.Getter) {
+		logger.Debugf("Got args %v", args)
+		params.Set(args)
+	})
+
+	args, ok := params.Get(40 * time.Second)
+	assert.True(t, ok)
+
+	argString := args.(string)
```
uaalto: This will fail if we don't get a value in `params` on time, but being a test we have control over that time and it is perfectly fine.

uaalto:
LGTM! Merging.

PR #4211, opened 2016-05-05
close response body when bad status code fix #3987

### LINKED ISSUE #3987: Too many open files running Lantern
So this is interesting. While testing pre lazy loading on scroll version of the newsfeed UI, I got too many open files on the backend presumably because of all the connections initiated from Lantern on the backend. 

chained proxy at x.x.x.x:443 to tcp://da.feedsportal.com:80: Unable to dial server (trusted) chained proxy at x.x.x.x:443: dial tcp x.x.x.x:443: socket: too many open files on pass 0...continuing

fffw:
@myleshorton It should be worth to add to 2.2.0. With this patch, no any `CLOSE` or `CLOSE_WAIT` seen after running for 10min, while previously, at least 2 will be seen in a few seconds.

### DIFF
--- a/src/github.com/getlantern/flashlight/client/handler.go
+++ b/src/github.com/getlantern/flashlight/client/handler.go
@@ -149,10 +149,11 @@ func pipeData(clientConn net.Conn, connOut net.Conn, closeFunc func()) {
 		closeFunc()
 	}()
 
-	// Then start coyping from proxy to client.
+	// Then start copying from proxy to client.
 	if _, err := io.Copy(clientConn, connOut); err != nil {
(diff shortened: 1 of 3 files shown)


myleshorton:
Nice @fffw! Looking now.

myleshorton:
Looking good @fffw and great catch. I just added a shadowing fix I've been meaning to make for awhile as well. I don't think it ends up being consequential, but this is certainly more correct. Mind taking a look?

fffw:
LGTM, waiting for CI

oxtoacart:
@fffw Nice debugging!

PR #4212, opened 2016-05-05
Adds a parameter to prevent the PAC URL from being cached by Windows

xiam:
Should fixes a number of issues, specially this one

* https://github.com/getlantern/lantern/issues/4175

And others that claimed that Lantern was not working after reboot or not working on Edge.

PR #4213, opened 2016-05-05
Adds a parameter to prevent the PAC URL from being cached by Windows (release-2.2.0)

xiam:
See https://github.com/getlantern/lantern/pull/4212


myleshorton:
I do think we should probably do this every time we set the pac to on @xiam (add a new cache busting argument). I have a feeling it would fix other stuff too.

myleshorton:
For now I'll merge and build though.

PR #4220, opened 2016-05-06
Switch to GravityCompat

atavism:
@myleshorton Just curious, does this fix the issue you were seeing with ``Gravity.START`` not existing?


myleshorton:
Yeah that'll definitely do it -- I'll build a new one.

Issue #4226, opened 2016-05-06
Mobile - Able to switch ON/OFF continuously and lantern crashes after two switches


sunshine19090:

atavism:
@sunshine19090 I'll send you a new build with this fixed.

sunshine19090:
The new build that @atavism sent me works fine. I will close it once it is moved to deployed.

PR #4228, opened 2016-05-06
Add back code for disabling ON/OFF switch temporarily

atavism:
This resolves https://github.com/getlantern/lantern/issues/4226 -- to prevent the on/off switch from being toggled continuously, we disable the switch temporarily while the VpnService updates the connection.


atavism:
@myleshorton This should be a quick fix. Just pulls in code from before to disable the on/off switch temporarily while the VpnService updates the connection. Resolves https://github.com/getlantern/lantern/issues/4226

myleshorton:
Yup working for me @atavism!

PR #4232, opened 2016-05-08
Changing no-cache argument everytime we set the PAC URL.

xiam:
Hey @myleshorton, this implements your suggestion of changing the pac URL everytime it is set, this effectively forces a reset, even on Windows. It looks like problems with Xunlei get somewhat _fixed_ after  resetting the pac file, users may trigger a reset by themselves with checking/unchecking the proxy all feature. On my experience, though, I had no problems with Xunlei 7.9 and latest QA version, only with [Xunlei 5 + IE](https://support.microsoft.com/en-us/kb/2520403) and current public beta.

### DIFF
--- a/src/github.com/getlantern/flashlight/main/pac.go
+++ b/src/github.com/getlantern/flashlight/main/pac.go
@@ -21,11 +21,11 @@ import (
 )
 
 var (
-	isPacOn     = int32(0)
-	pacURL      string
+	isPacOn       = int32(0)
+	pacURL        string
 	pacURLNoCache string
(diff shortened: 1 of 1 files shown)


myleshorton:
Very interesting @xiam thanks! 

Any chance you can make this against devel actually? I'd like to do a 2.2.1 (or maybe even 2.3.0?) soon from devel just to get that code and fixes released, and all the refactorings on devel are starting to make merging old stuff challenging.

myleshorton:
LGTM @xiam! I'll just merge this into devel myself.

PR #4237, opened 2016-05-08
structured error reporting

fffw:


src/github.com/getlantern/errlog/errlog.go
```diff
@@ -0,0 +1,529 @@
+/*
+Package errlog defines error types used across Lantern project and implements
+functions to manipulate them.
+
+  var elog = errlog.ErrorLoggerFor("package-name")
+  if n, err := Foo(); err != nil {
+    elog.Log(err, errlog.WithOp("foo"), errlog.WithField("foo": "bar"))
+  }
+
```
fffw: @oxtoacart @uaalto How do you think about above idiom to report errors? It follows the usage of `golog`. Besides reporting to globally supplied reporter, it calls `golog.Error()` internally, so we can completely replace `log.Errorf` with it. More detailed usage is in `errlog_test.go`. I also replaced `log.Errorf` of a few packages as examples.

src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -37,7 +38,10 @@ func (client *Client) initBalancer(cfg *ClientConfig) (*balancer.Balancer, error
 			log.Debugf("Adding chained server: %v", s.Addr)
 			dialers = append(dialers, dialer)
 		} else {
-			log.Errorf("Unable to configure chained server. Received error: %v", err)
+			elog.Log(err, errlog.WithOp("configure"), errlog.WithProxy(&errlog.ProxyingInfo{
```
fffw: `WithProxy` with struct looks ugly, I'm considering splitting it to `WithProxyType`, `WithProxyAddr`, etc.

src/github.com/getlantern/errlog/errlog.go
fffw: An alternative is free style fields similar to https://github.com/go-kit/kit/tree/master/log. As the `errlog` package is used only by us at this point, I chose more rigid syntax.

src/github.com/getlantern/flashlight/client/balancer.go
oxtoacart: You could improve the syntax by either providing a convenience method on ChainedServerInfo to create the ProxyingInfo, or you could actually make ProxyingInfo an interface and make sure that ChainedServerInfo implements that interface.  Both options make life for the user of this API easier.

src/github.com/getlantern/errlog/errlog.go
oxtoacart: Why not just use the struct directly, like this:

```golang
elog.Log(&Error{
    Err: err,
    Op: "foo",
    Extras: map[string]string {
        "foo": "bar",
    },
})
```

To me, that seems more readable and also more conventional.

src/github.com/getlantern/errlog/errlog.go
```diff
@@ -0,0 +1,529 @@
...
+}
+
+type StdReporter struct {
+}
+
+func (l StdReporter) Report(e *Error) {
+	fmt.Printf("%+v", string(toJSON(e)))
+}
+
+func parseError(err error) (op string, goType string, desc string, extra map[string]string) {
```
oxtoacart: Very cool!

src/github.com/getlantern/errlog/errlog.go
fffw: Because in most cases simply `elog.Log(err)` would be enough. Different set of extra parameters can be supplied in different cases. For example, user agent is useful only when error interacting with client connections.

oxtoacart: @fffw You could follow the same convention as the built-in log package, which accepts an interface{} and then treats the value appropriately.  Regarding optionality - the conventional thing in Go is to leave the fields that you're not interested out of your struct initializer and the receiving function then fills those in.  I'll do a little PR to show what I'm suggesting.

oxtoacart: #4239 shows what I mean

src/github.com/getlantern/errlog/errlog.go
```diff
@@ -0,0 +1,540 @@
...
+type UserAgentInfo struct {
+	UserAgent string `json:"userAgent,omitempty"`
+}
+
+func (ul *UserAgentInfo) String() string {
+	return fmt.Sprintf("UserAgent=%s", ul.UserAgent)
+}
+
+// Error wraps system and application errors in unified structure
+type Error struct {
```
oxtoacart: I still think it's important to have a unique error code corresponding to the logical condition, e.g. "CPCR" = "Chained Proxy Certificate Rejected".  As we build our queries and perform analysis over time, we can't rely on package and type names because as things get refactored, these may change.  In fact, I think the code is more important than the go package and type.

src/github.com/getlantern/errlog/errlog.go
```diff
@@ -0,0 +1,540 @@
...
+	"strconv"
+	"syscall"
+	"time"
+
+	"github.com/getlantern/golog"
+	"github.com/getlantern/jibber_jabber"
+	"github.com/getlantern/osversion"
+)
+
+type systemInfo struct {
```
oxtoacart: It would be good to include the version of our software here as well.

src/github.com/getlantern/errlog/errlog.go
oxtoacart: If it's more readable, we can use strings instead of codes, e.g. "ChainedProxyCertificateRejected".

src/github.com/getlantern/errlog/errlog.go
```diff
@@ -0,0 +1,540 @@
...
+		case *net.AddrError:
+			goType = "net.AddrError"
+			desc = actual.Err
+			extra["addr"] = actual.Addr
+		case *net.DNSError:
+			goType = "net.DNSError"
+			desc = actual.Err
+			extra["domain"] = actual.Name
+			if actual.Server != "" {
+				extra["dnsServer"] = actual.Server
```
oxtoacart: Rather than using camel-case for extras and JSON field names, how about using underscores like `dns_server`?  This data will be queried in InfluxDB and other databases, and the underscore convention is more common there.

oxtoacart:
It occurs to me that collecting all of the context for a particular error in one place may be difficult, because different pieces of information like the request ID, the server IP, the detailed error, and so on are buried in different parts of the stack.  If the caller of the library has to collect all of this information in one place to make a call, that could require a lot of changes to our APIs to make this information available.

Something like [this](https://github.com/jtolds/gls) or [this](https://github.com/tylerb/gls) could help.  It's basically a similar technique to Java's [ThreadLocal](https://docs.oracle.com/javase/7/docs/api/java/lang/ThreadLocal.html).

src/github.com/getlantern/errlog/errlog.go
fffw: The problem is that we can't predefine all possible codes. It's also tedious for the caller to find the correct error codes. For monitoring purpose, most operations are not to exactly match a few error codes, but list the top N errors, show the distribution, or send alert if the count exceeds some criteria. For troubleshooting, we have user id, device id, request id, ip address, etc to acurately filter the dataset.

Changed field value does little harm as they are same in a version.

src/github.com/getlantern/errlog/errlog.go
fffw: Yeah good catch

src/github.com/getlantern/errlog/errlog.go
fffw: In fact `DNS server` is better as it's what we expect to see on the dashboard!

src/github.com/getlantern/errlog/errlog.go
oxtoacart: We don't need to predefine all possible codes.  The caller simply needs to say "logically, what's happening here is that the chained proxy's certificate was rejected, so let's use the code "ChainedProxyCertificateRejected".  Having a consistent string corresponding to a logical error condition across Lantern versions makes life much easier for the person monitoring the system.  It also makes it possible to write queries like "find the clients whose certificate rejection rate exceeds the 95th percentile level" without having to change that query from one version of Lantern to the next.

What do you think @myleshorton?

src/github.com/getlantern/errlog/errlog.go
oxtoacart: Good idea on including the Go version too!

src/github.com/getlantern/errlog/errlog.go
oxtoacart: `DNS server` is horrible for querying because it has to be quoted, like:

```sql
select * from health group by "DNS server"
```

Also, I have to remember that it's `DNS server` and not `DNS Server` or `dns server` or `dns Server`.  Sticking to lower-cased names with underscores will make life much easier for writing queries.  If it's more convenient, I can do this automatically on the server side by lower-casing everything and replacing spaces and hyphens with underscores.

src/github.com/getlantern/errlog/errlog.go
```diff
@@ -0,0 +1,540 @@
...
+	Scheme     string    `json:"scheme,omitempty"`
+}
+
+func (pi *ProxyingInfo) String() string {
+	var buf bytes.Buffer
+	if pi.ProxyType != "" {
+		_, _ = buf.WriteString(" ProxyType=" + string(pi.ProxyType))
+	}
+	if pi.LocalAddr != "" {
+		_, _ = buf.WriteString(" LocalAddr=" + pi.LocalAddr)
```
uaalto: This serialization doesn't use any separator between fields. How is it going to be efficiently deserialized?

src/github.com/getlantern/errlog/errlog.go
uaalto: What you are proposing here @oxtoacart is to have another field with the error type? I can see that in the code already, so I might have misunderstood what you propose.

src/github.com/getlantern/flashlight/analytics/analytics.go
```diff
@@ -25,7 +27,8 @@ const (
 )
 
 var (
-	log = golog.LoggerFor("flashlight.analytics")
+	log  = golog.LoggerFor("flashlight.analytics")
+	elog = errlog.ErrorLoggerFor("flashlight.analytics")
```
uaalto: Making this differentiation is so some errors are reported using this mechanism and others using the traditional GoLog? I would propose using a common interface for both, unless there is a strong reason to keep these separate.

src/github.com/getlantern/flashlight/client/balancer.go
uaalto: In my opinion the convenience method approach is more flexible. In order to make an interface hierarchy of Info types we need to expose all its members as functions and may end up being more convoluted.

src/github.com/getlantern/errlog/errlog.go
```diff
@@ -0,0 +1,540 @@
...
+}
+
+type Reporter interface {
+	Report(*Error)
+}
+
+var currentReporter Reporter = &StdReporter{}
+
+func ReportTo(r Reporter) {
+	currentReporter = r
```
uaalto: This is where I assume I'll have to plug my event processing mechanism, correct @fffw ?

src/github.com/getlantern/errlog/errlog.go
oxtoacart: @fffw Do you mean the `GoType` field?  In the tests, it looks like we're populating this with the type of the error object.  What I want is a logical error code like "ChainedProxyCertificateRejected", which doesn't map directly to a go error type.  Should we put that in the `GoType` field?  Should we maybe rename it to `ErrorCode`?

src/github.com/getlantern/flashlight/analytics/analytics.go
oxtoacart: Yeah, I had a similar thought actually.

src/github.com/getlantern/flashlight/client/balancer.go
oxtoacart: @uaalto I don't understand your comment, can you please explain more, maybe with an example?

oxtoacart: @uaalto Never mind, I understand now.  I had forgotten the context of the comment :)  Yes, I think the convenience method approach is good.

uaalto: My knowledge of Go idioms is not very extense, you might have a cleaner idea of what a interface-base solution might look like. My impression is that subtyping with dumb objects Go might prefer embedding.

src/github.com/getlantern/errlog/errlog.go
oxtoacart: Another thing to consider is that we may want to keep the number of distinct ErrorCodes somewhat low, since we're planning to use these as tags in InfluxDB.

src/github.com/getlantern/errlog/errlog.go
fffw: It's for logging purpose, I should rename the method to make it explicit. The format to report depends on the actual reporter.

src/github.com/getlantern/flashlight/analytics/analytics.go
fffw: With the new suggestion by @oxtoacart we can define `Error` in `errors` package, and extend `golog` to simply recognize it.

src/github.com/getlantern/errlog/errlog.go
fffw: Yeah @uaalto , but it seems that we'll move the functionality to `golog`.

src/github.com/getlantern/flashlight/analytics/analytics.go
oxtoacart: @fffw I like your idea of enhancing golog to handle special reporting of errors.  Since the reporting logic probably just goes into the Reporter implementation, we could actually make this opaque to golog by just using the built-in error type.  For example:

```go
package golog

...

type Reporter interface {
    Report(error)
}

func (l *logger) Error(arg interface{}) {
    // Log just as before
    l.print(GetOutputs().ErrorOut, 4, "ERROR", arg)

    if l.reporter != nil {
        err, isError := arg.(error)
        if isError {
            l.reporter.Report(err)
        }
    }
}
```

src/github.com/getlantern/errlog/errlog.go
uaalto: Totally fine, just wanted to see if we are thinking the same way.

fffw:
Thanks @oxtoacart and @uaalto for all your insightful comments. I'll refactor based on them.

src/github.com/getlantern/errors/errors.go
```diff
@@ -0,0 +1,641 @@
+/*
+Package errors defines error types used across Lantern project.
```
fffw: So here's the new API. It seems more expressive and requires less keystoke. @oxtoacart @uaalto any thoughts?

Instead of changing `golog` package, I make `errors` package to call `golog.Error()` if logging flag is set globally. With the `stack` package, we can attach the call stack when error is created, and attach another call stack when error is reported, if necessary.

src/github.com/getlantern/errors/errors.go
```diff
@@ -0,0 +1,641 @@
...
+	// Any extra fields
+	Extra map[string]string `json:"extra,omitempty"`
+
+	*ProxyingInfo
+	*UserLocale
+	*UserAgentInfo
+}
+
+func (e *Error) Report() {
+	caller := stack.Caller(1)
```
oxtoacart: If I've manually created an Error and then call Report, we'll never have called applyDefaults() or attached the stack.

src/github.com/getlantern/errors/errors.go
```diff
@@ -0,0 +1,641 @@
...
+Wrap() method will try as much as possible to extract details from the error
+passed in, if it's errors defined in Go standard library. For application
+defined error type, at least the Go type name and what err.Error() returns
+will be recorded.
+
+Extra fields can be chained in any order, at any time.
+
+  func Foo() *Error {
+  	//...
+    return errors.New("some error").With("some_counter", 1).WithOp("connect")
```
oxtoacart: I'm okay with this fluent-style syntax, but given that Go has a nice struct initialization syntax and the Error type is exported, I would just manually create an &Error{...}.  I actually find that more readable too:

```go
return &Error{
    Desc: "some error",
    Op: "connect",
    Extra: {
        "some_counter": 1,
    }
}
```

The fluent syntax would be more readable without the `With` prefix, for example:

```go
return errors.New("some error")
                    .Op("connect")
                    .Extra("some_counter", 1)
```

oxtoacart: Looking at my previous comment, I think part of why the struct initializer looks more readable to me is because of the syntax highlighting and whitespace.

src/github.com/getlantern/errors/errors.go
```diff
@@ -0,0 +1,641 @@
...
+	}
+	if e.UserAgentInfo != nil {
+		_, _ = io.WriteString(w, e.UserAgentInfo.String())
+	}
+	for k, v := range e.Extra {
+		_, _ = io.WriteString(w, " "+k+"="+v)
+	}
+}
+
+func (e *Error) attachStack(skip int) {
```
oxtoacart: Good idea to attach stack!

src/github.com/getlantern/errors/errors.go
fffw: A function call is inevitable to `applyDefaults` and attach the stack, and should at the very place the Error is created (to get the accurate stack). It seems like we should avoid creating Error manually.

src/github.com/getlantern/errors/errors.go
fffw: I choose `WithOp` simply because `Op` is already a struct field, other method calls are without "With". Nothing significant, but I choose to lazily create ProxyingInfo/UserLocale/UserAgentInfo etc. Method call can hide the details of how the fields are structured. Only the reporter needs those details.

If we allow manually creating Error, then the struct should be flat, or we'll end up with:
```
return &errors.Error{
    Desc: "some error",
    Op: "connect",
    ProxyingInfo: &errors.ProxyingInfo{
      ProxyType: errors.NoProxy,
      ProxyAddr: "...",
    }
    Extra: map[string]string{
        "some_counter": 1,
    }
}
```
instead of
```
return errors.New("some error").WithOp("connect").
    ProxyType(errors.NoProxy).ProxyAddr("...").With("some_counter", 1)
```

oxtoacart: In that case, I think we should disallow manual creation of Error by making it private and instead exposing an `Error` interface backed by an `error` implementation.

fffw: Unfortunately, Reporter has to know the structure to selectively report certain fields or have some transformation. I did think about exporting a flat `map[string]interface{}` to Reporter, but that still losses static check of name and type.

fffw:
Remaining packages to modify  (`find . -name *.go | grep -v "_test" | xargs grep log.Error | cut -f 1 | rev | cut -d "/" -f 2- | rev | sort -u`)

- [ ] autoupdate
- [ ] balancer
- [ ] buuid
- [ ] chained
- [ ] checkfallbacks
- [ ] detour
- [ ] flashlight
- [ ] flashlight/app
- [ ] flashlight/autoupdate
- [ ] flashlight/genconfig
- [ ] flashlight/logging
- [ ] flashlight/main
- [ ] flashlight/proxiedsites
- [ ] flashlight/ui
- [ ] fronted
- [ ] lantern
- [ ] launcher
- [ ] profiling
- [ ] protected
- [ ] systray
- [ ] tarfs
- [ ] yamlconf

src/github.com/getlantern/errors/errors.go
uaalto: What you refer to as "reporter" here is, if I understand correctly, needs to be connected to what I've denominated the Event Emitter, in the events-pipeline library. This one will receive a Key and a list of values in the form `map[string]interface{}`. But the actual interface between your error reporter and the my emitter would convert the static types to a `map[string]interface{}`. Does that sound reasonable to you?

fffw:
What next:
1. New helpers to extract information from `net.Conn` or `http.Request/Response`.
2. data race

oxtoacart:
Closing in favor of https://github.com/getlantern/lantern/pull/4363

PR #4239, opened 2016-05-08
Suggested syntax for structured error logging

oxtoacart:


src/github.com/getlantern/errlog/errlog_test.go
```diff
@@ -53,22 +53,49 @@ func TestAnonymousError(t *testing.T) {
...
+		Source:    source,
+		GoPackage: "specified package",
+		GoType:    "specified type",
+		Desc:      "specified desc",
+		Op:        "specified op",
+		Extra: map[string]string{
+			"extra1": "specified extra1",
+		},
+	}
+	l.Log(err)
```
oxtoacart: Here is an example of logging a structured error.  In real usage, one could just do this in one line, and omit the fields that one doesn't care about, for example:

```
l.Log(errlog.Error{
    Source: originalError,
    Op: "proxying",
    Extra: map[string]string{
        "server": "127.0.0.1:443",
    },
})
```

The cool thing about modeling it this way too is that these structured errors can be propagated, since they're real errors, i.e.:

```
err := doSomething()
if err != nil {
    return errlog.Error{
        Source: originalError,
        Op: "proxying",
        Extra: map[string]string{
            "server": "127.0.0.1:443",
        },
    }
}
```

src/github.com/getlantern/errlog/errlog.go
```diff
@@ -63,6 +63,29 @@ type systemInfo struct {
...
+
+	defaultSystemInfo = &systemInfo{
+		OSType:    runtime.GOOS,
+		OSArch:    runtime.GOARCH,
+		OSVersion: version,
+	}
+
+	lang, _ := jibber_jabber.DetectLanguage()
+	country, _ := jibber_jabber.DetectTerritory()
+	userLocale = &UserLocale{
```
fffw: `userLocale` can change during Lantern lifecycle. When application code does log error with locale (which should happen rarely because most code doesn't care about locale), then the information should be accurate.

src/github.com/getlantern/errlog/errlog_test.go
fffw: Cool @oxtoacart ! Then we can rename it to `errors` package (which was the package name I initially use) and replace the standard one. I prefer `errors.New(string)` and `errors.Wrap(error)` instead of creating object directly. By changing the method signature to return `errors.Error`, we encourage caller to attach anything valuable to it with type casting.

src/github.com/getlantern/errlog/errlog.go
oxtoacart: Ah, good point!  I'll fix that.

src/github.com/getlantern/errlog/errlog_test.go
oxtoacart: Thanks!  I like errors too.  Do you think we also want `errors.Log()`, or should we have two packages, `errors` and `errlog`?

fffw:
I'll merge and keep going based on it.

PR #4260, opened 2016-05-11
Caching vetted masquerades to disk for up to 24 hours

oxtoacart:


myleshorton:
Hey I just went ahead and fixed the test @oxtoacart and otherwise continuing with the review

myleshorton:
I gotta say this seems like overkill here @oxtoacart. The ring buffer is cool, but all we really care about is persisting the set of masquerades that have fully passed the HEAD check and maybe the masquerades that have passed/failed the connection check right? Couldn't we just throw that in a slice and encode/decode from JSON or yaml? I'm just worried this is more code that can fail and more code that whoever works on this part of the app has to understand, whereas the above is super straightforward unless there's something I'm missing here.

oxtoacart:
@myleshorton Yeah, I think we can simplify.  Give me a bit.

oxtoacart:
FWIW, the RingBuffer is well tested :)

oxtoacart:
In hindsight, it would have been faster to just write the whole JSON out every time.  However, now that we have this, it is well tested and is quite a bit more efficient since it only writes the updated JSON and the updated index information.

oxtoacart:
Closing in lieu of #4272

PR #4266, opened 2016-05-12
Way to hide Error

oxtoacart:
@fffw Here's a way to keep people from manually instantiation an `Error`.  Basically, make `Error` not implement the `error` interface and instead provide a `Finish()` method that creates an `error` (which is of an unexported type `wrapped`).  It's one more thing to call, but it prevents abuse of the library.

I had to rename `Error` to `Structured` because I can't embed `Error` in `wrapped` and define an `Error()` method to implement the `error` interface.

I also suggest moving the errors package into flashlight because it's very specific to that.

### DIFF
--- a/src/github.com/getlantern/flashlight/errors/errors.go
+++ b/src/github.com/getlantern/flashlight/errors/errors.go
@@ -14,9 +14,9 @@ be recorded.
 
 Extra fields can be chained in any order, at any time.
 
-  func Foo() *Error {
+  func Foo() error {
   	//...
-    return errors.New("some error").With("some_counter", 1).WithOp("connect")
+    return errors.New("some error").With("some_counter", 1).WithOp("connect").Finish()
   }
   ...
   if err := Foo(); err != nil {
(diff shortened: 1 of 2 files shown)


oxtoacart:
In practice, I think calling Finish() is not so bad, and it will save lots of trouble by preventing people from using Error as an error.

fffw:
Thanks, @oxtoacart . There are a few concerns make me thinking about the necessarity of this change:

1. Coders usually follow previous practice or the document or check the tests to learn how to use a package, instead of creating an object just because it's public.
2. The damage of manually creating an `Error` is low. It doesn't affect the program's feature. From the reported information, we can know which line of code is reporting the error without necessary fields, and fix it.
3. The mandatory `.Finish()` call is tedious and couter-intuitive, especially considering `errors` package will be used frequently.
4. The code turns a bit harder to understand to handle this accidental complexity.

fffw:
Yeah it looks like the package should be under `flashlight`, but other packages like `detour` or `balancer` may also want to report error without return anything to caller. Error reporting is a general requirement. We can make the package extendable through reflection, but it feels like an over design ATM.

PR #4283, opened 2016-05-13
Dftransport

oxtoacart:
@uaalto You were right, the existing utility mostly had what we needed.  The main change is that instead of executing these requests in parallel first, we always try the chained server first and then fall back to the fronted server.

cc @fffw


src/github.com/getlantern/flashlight/util/http.go
```diff
@@ -36,10 +36,17 @@ func success(resp *http.Response) bool {
...
-// and direct fronted servers in parallel.
-func NewChainedAndFronted(proxyAddrFN eventual.Getter) *chainedAndFronted {
-	cf := &chainedAndFronted{proxyAddrFN: proxyAddrFN}
+// NewChainedAndFronted creates a new struct for accessing resources using
+// chained and direct fronted servers. If parallel is true, the request will be
+// sent to both servers. If false, we'll first try the chained option and only
+// fall back to direct fronting if chained fails. Once a chained requets
+// succeeds, we only used chained going forward unless it fails, at which point
+// we go back to the original regime.
+func NewChainedAndFronted(proxyAddrFN eventual.Getter, parallel bool) *chainedAndFronted {
```
uaalto: I'm not particularly fond of this API. I'd use a constant with a name that clearly states the strategy
http.ChainedAndFrontedParallelStrategy
http.ChainedAndFrontedSequentialStrategy

There might be other strategies in the future, but more importantly, I think we could gain in clarity, instead of using a large comment (which is only found in the definition of the function, not in every usage).

uaalto: Those names could be shorter ;)

uaalto:
```
util.ParallelStrategy
util.SequentialStrategy

// ... and perhaps for testing or some usage in the future
util.OnlyFrontedStrategy
util.OnlyChainedStrategy

// ...etc
```

What do you think?

uaalto:
Implementation looks good to me. Let me know what you think about the API proposal, and it'd be ready to merge.

src/github.com/getlantern/flashlight/util/http.go
fffw: Maybe shorter names: `Parallel` and `ChainedFirst`

oxtoacart:
@uaalto I liked your suggestion of cleaning up the API that I ran with it.  Comments in code.

While testing all this, I found some other issues that I just kind of cleaned up.

oxtoacart:
@myleshorton These changes might be interesting to you as well.

src/github.com/getlantern/flashlight/proxied/proxied.go
```diff
@@ -1,4 +1,8 @@
-package util
+// Package proxied provides  http.Client implementations that use various
+// combinations of chained and direct domain-fronted proxies.
+//
+// Remember to call SetProxyAddr before obtaining an http.Client.
+package proxied
```
oxtoacart: All the http-related stuff from util is now in the proxied package.  Fundamentally, it works the same, but it exposes a common and consistently named set of functions that all return an *http.Client in lieu of our custom HTTPFetcher interface (which is no longer necessary).

src/github.com/getlantern/flashlight/proxied/proxied.go
```diff
@@ -24,50 +28,86 @@ const (
...
+	}
+	return addr.(string), true
+}
+
+// ParallelPreferChained creates a new http.RoundTripper that attempts to send
+// requests through both chained and direct fronted routes in parallel. Once a
+// chained request succeeds, subsequent requests will only go through Chained
+// servers unless and until a request fails, in which case we'll start trying
+// fronted requests again.
+func ParallelPreferChained() *http.Client {
```
oxtoacart: This is the old "ChainedAndFronted".

src/github.com/getlantern/flashlight/proxied/proxied.go
```diff
@@ -24,50 +28,86 @@ const (
...
+		parallel: true,
+	}
+	cf.setFetcher(&dualFetcher{cf})
+	return &http.Client{Transport: cf}
+}
+
+// ChainedThenFronted creates a new http.RoundTripper that attempts to send
+// requests first through a chained server and then falls back to using a
+// direct fronted server if the chained route didn't work.
+func ChainedThenFronted() *http.Client {
```
oxtoacart: This is the new ChainedThenFronted for submitting metrics.

src/github.com/getlantern/flashlight/config/fetcher.go
```diff
@@ -49,8 +48,8 @@ type UserConfig interface {
 
 // NewFetcher creates a new configuration fetcher with the specified
 // interface for obtaining the user ID and token if those are populated.
-func NewFetcher(conf UserConfig, httpFetcher util.HTTPFetcher) Fetcher {
-	return &fetcher{lastCloudConfigETag: map[string]string{}, user: conf, httpFetcher: httpFetcher}
+func NewFetcher(conf UserConfig, httpClient *http.Client) Fetcher {
```
myleshorton: All looking great @oxtoacart just wondering on this change. The interface makes it more testable, and hey the [Go best practices post](https://peter.bourgon.org/go-best-practices-2016/) even explicitly recommends it!

src/github.com/getlantern/flashlight/proxied/proxied.go
```diff
@@ -300,22 +365,13 @@ func httpClient(rootCA string, proxyAddrFN eventual.Getter, persistent bool) (*h
...
-		log.Debug("Waiting for proxy server to come online")
-		proxyAddr, ok := proxyAddrFN(60 * time.Second)
+	tr.Proxy = func(req *http.Request) (*url.URL, error) {
+		proxyAddr, ok := getProxyAddr()
 		if !ok {
-			// Instead of finishing here we just log the error and continue, the client
-			// we are going to create will surely fail when used and return errors,
-			// those errors should be handled by the code that depends on such client.
-			log.Errorf("Proxy never came online")
+			return nil, ErrChainedProxyUnavailable
```
oxtoacart: Note - we no longer fall back to a direct connection.  As far as I can tell, we never used this, and the clients weren't expecting this to go direct.  Direct connections are also a fingerprinting/blocking threat, so I took it out, which simplified the code too.

src/github.com/getlantern/flashlight/proxied/proxied_test.go
```diff
@@ -0,0 +1,163 @@
+package proxied
```
oxtoacart: This is just based on the old util_test, minus the hash stuff which is still in the util package.

src/github.com/getlantern/flashlight/proxied/proxied.go
```diff
@@ -24,50 +28,86 @@ const (
...
 	return resp.StatusCode > 199 && resp.StatusCode < 400
 }
 
-// NewChainedAndFronted creates a new struct for accessing resources using chained
-// and direct fronted servers in parallel.
-func NewChainedAndFronted(proxyAddrFN eventual.Getter) *chainedAndFronted {
-	cf := &chainedAndFronted{proxyAddrFN: proxyAddrFN}
+// SetProxyAddr sets the eventual.Getter that's used to determine the proxy's
+// address. This MUST be called before attempting to use the proxied package.
+func SetProxyAddr(addr eventual.Getter) {
```
oxtoacart: Instead of passing the proxy address to the individual functions all the time, we just set it once.  More convenient, less verbose.  Yes there's some global state, but this is something that is set once at startup and never changes.

src/github.com/getlantern/fronted/test_support.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"crypto/x509"
+	"testing"
+
+	"github.com/getlantern/keyman"
+)
+
+// ConfigureForTest configures fronted for testing using default masquerades and
+// certificate authorities.
+func ConfigureForTest(t *testing.T) {
```
oxtoacart: We had a bunch of tests that all duplicated masquerade configuration.  That was annoying and the tests were starting to fail because they had out of date masquerades.  I've moved the default masquerades into the fronted package and provide a convenience ConfigureForTest method that any tests relying on fronted functionality can just use.

oxtoacart: This also makes the other test files nice and short again, so when we make future changes to them, GitHub will be able to display a proper diff.

src/github.com/getlantern/flashlight/util/hash_test.go
```diff
@@ -0,0 +1,21 @@
...
+func TestGetFileHash(t *testing.T) {
+	wd, _ := os.Getwd()
+	path := wd + "/hash.go"
+
+	hash, _ := GetFileHash(path)
+	//log.Debugf("Got hash! %x", hash)
+	log.Debugf("Got hash! %v", hash)
+
+	// Update this with shasum -a 256 hash.go
+	assert.Equal(t, "699b8a31a9c8c0b26ce1b4cbe79af1de97de9ef83b9ab717241ed53fb5d86df2", hash,
```
oxtoacart: The only thing that changed in this test is updating the hash since the contents of hash.go changed a little.

src/github.com/getlantern/flashlight/autoupdate/autoupdate.go
```diff
@@ -48,7 +47,7 @@ func Configure(cfg *config.Config) {
 func enableAutoupdate(cfg *config.Config) {
 	var err error
 
-	httpClient, err = util.HTTPClient(cfg.CloudConfigCA, client.Addr)
+	httpClient, err = proxied.ChainedNonPersistent(cfg.CloudConfigCA)
```
oxtoacart: This is a good example of the syntactic change.  In keeping with @uaalto's suggestion, this API looks cleaner and also eliminates the dependency on the client package for the Addr.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -290,7 +284,7 @@ func (cfg *Config) ApplyDefaults() {
 	}
 
 	if cfg.TrustedCAs == nil || len(cfg.TrustedCAs) == 0 {
-		cfg.TrustedCAs = defaultTrustedCAs
+		cfg.TrustedCAs = fronted.DefaultTrustedCAs
```
oxtoacart: Note that fronting defaults are now kept in the fronted package.

src/github.com/getlantern/flashlight/feed/feed.go
```diff
@@ -297,3 +289,20 @@ func determineLocale(defaultLocale string) string {
 	}
 	return defaultLocale
 }
+
+func getHTTPClient() (*http.Client, error) {
+	var err error
+	httpClientMutex.Lock()
+	if _httpClient == nil {
+		_httpClient, err = proxied.ChainedNonPersistent("")
```
oxtoacart: This was an opportunistic fix.  I noticed that feed was constructing a new http.Client every time it fetched feeds, meaning that it didn't take advantage of keep-alives.  By making the client a global and lazily initializing it, we can take advantage of persistent connections.

cc: @atavism

src/github.com/getlantern/flashlight/config/fetcher.go
oxtoacart: I dig it, it's just that previously we had some methods that returned an HTTPFetcher and some that returned an *http.Client.  http.Client is actually a superset of HTTPFetcher, and I value consistency, so I went with returning http.Client across the board.  What do you think @myleshorton ?

oxtoacart: Also, http.Client has the advantage of being defined in the standard library, so more likely to just be something that we can pass around.  If we go to an interface, I would use http.RoundTripper which is actually identical to our HTTPFetcher.

src/github.com/getlantern/flashlight/feed/feed.go
myleshorton: Cool. I don't love the laziness aspect of this in the sense that there's nothing we're really saving (creating the HTTPClient is super cheap), but the fact the ChainedNonPersistent can return an error makes a solution unclear. To me it would be significantly nicer if we could just say `_httpClient     = getHTTPClient()` at the beginning and skip all of this, but that would require something like calling `getProxyAddr` once at startup in `proxied.go` and then just always using that, failing fast potentially just at startup if that ever fails. That should never fail, though, since it's just the proxy listening at 127.

src/github.com/getlantern/flashlight/autoupdate/autoupdate.go
oxtoacart: @myleshorton this is a good example of where we need an actual *http.Client.  It gets passed to autoupdate, which passes it to go-update which actually does some stuff with the transport.  We could rewrite all that code to accept a RoundTripper instead, but IIRC it's pretty common for Go libraries to accept *http.Clients.

src/github.com/getlantern/flashlight/config/fetcher.go
myleshorton: Yeah I actually think the Fetcher (maybe there's a better name) adds a lot to testability. Is there anywhere we call anything but `Do` on that client? If not then to me the interface is the clear winner. If we do call something else on it, that might indicate the code there is too tightly coupled.

Maybe "HTTPDoer" straight from that article like:

```
type HTTPDoer interface {
    Do(*http.Request) (*http.Response, error)
}
```

src/github.com/getlantern/flashlight/feed/feed.go
oxtoacart: The cost is not in creating an HTTP client, the cost is in that HTTP client opening connections to our proxy.  Previously, we did that for every request made by GetFeed because it created a non-persistent http client every time it was called.  Now, we have a single persistent HTTP client that can reuse connections within a single call to GetFeed and across calls to GetFeed.

src/github.com/getlantern/flashlight/config/fetcher.go
oxtoacart: As mentioned in another comment, our autoupdate relies on actually having an http.Client.  It doesn't have to, and I could go in and refactor that too if you want.  Regarding the interface, I don't see why we should redefine what's already defined in `http.RoundTripper`.

oxtoacart: I have an idea for how to have our cake and eat it too - stay tuned.

src/github.com/getlantern/flashlight/proxied/proxied.go
```diff
@@ -65,23 +65,23 @@ func getProxyAddr() (string, bool) {
 // chained request succeeds, subsequent requests will only go through Chained
 // servers unless and until a request fails, in which case we'll start trying
 // fronted requests again.
-func ParallelPreferChained() *http.Client {
+func ParallelPreferChained() http.RoundTripper {
```
oxtoacart: @myleshorton I switched to returning RoundTripper.

src/github.com/getlantern/flashlight/autoupdate/autoupdate.go
```diff
@@ -45,13 +45,14 @@ func Configure(cfg *config.Config) {
...
 func enableAutoupdate(cfg *config.Config) {
-	var err error
-
-	httpClient, err = proxied.ChainedNonPersistent(cfg.CloudConfigCA)
+	rt, err := proxied.ChainedNonPersistent(cfg.CloudConfigCA)
 	if err != nil {
 		log.Errorf("Could not create proxied HTTP client, disabling auto-updates: %v", err)
 		return
 	}
+	httpClient = &http.Client{
```
oxtoacart: @myleshorton Where underlying libraries require an http.Client, I wrap the RoundTripper from proxied in a client.

While doing this, it occurred to me that we're really not taking advantage of the Timeout capability on http.Client.  Something to consider for another day.

myleshorton:
Nice -- I like the use of RoundTripper.

oxtoacart:
Manually tested mobile, and feeds are loading okay.

src/github.com/getlantern/flashlight/feed/feed.go
atavism: Awesome, @oxtoacart! I like the change to make the client global to take advantage of persistent connections. There is one issue though: I'm using the ``proxyAddr`` to indicate whether or not to use a proxied vs ordinary HTTP client. If none is specified, that means full-device VPN mode is enabled. (if it is enabled, we don't want to proxy any of those requests, otherwise those connections end up being rerouted back through the ``VpnService``). BTW, it would be useful to put this functionality in a separate package too (or maybe ``util``?) since its being duplicated in a few places now!

oxtoacart: Good catch @atavism, thanks!  I've fixed this.

src/github.com/getlantern/flashlight/proxied/proxied.go
uaalto: This decision is polemical. Globals vs. argument passing percolating into different layers are at the opposite of the spectrum, which clear differences. But not so clear which one is best.
Obviously globals introduce convenience, but generally reduce testability and encapsulation.
I guess my main question here would be: is this 100% the case where we want to set this only once and is never going to change throughout the life of the program? If so, it should be ok, including for testing.

src/github.com/getlantern/flashlight/proxied/proxied.go
uaalto: I think this  is a good change. If we want direct connection at some point we should be able to do it at the library user side. Just providing the proper error signal from the Chained/Fronted API should be fine.
Besides, it wasn't obvious that it was doing that (i.e. too "smart" behavior)

uaalto:
I see that @myleshorton's suggestions have been integrated.
I just added a couple of comments, but overall I see the refactoring as a real improvement in the API. I'll wait for @myleshorton opinion on this before merging, as he's been reviewing as well.

src/github.com/getlantern/flashlight/proxied/proxied.go
oxtoacart: Correct, it never changes after being set the first time.

-- 
Cheers,
Ox

PGP Key - https://pgp.mit.edu/pks/lookup?op=get&search=ox@getlantern.org

- Marge Piercy

src/github.com/getlantern/flashlight/config/fetcher.go
```diff
@@ -131,7 +130,7 @@ func (cf *fetcher) fetchCloudConfig(cfg *Config) ([]byte, error) {
 	// successive requests
 	req.Close = true
 
-	resp, err := cf.httpFetcher.Do(req)
+	resp, err := cf.rt.RoundTrip(req)
```
myleshorton: This is actually kind of amazing that just calling RoundTrip essentially in place of Do accomplishes the same thing. We don't technically need the HTTP client itself at all here - nice!

oxtoacart: It just depends on what features we need.  http.Client adds redirect handling, cooky handling and probably some other stuff, but I don't think that's needed in this particular spot.

uaalto:
Merging!

PR #4290, opened 2016-05-13
[WIP] Borda reporting

uaalto:
Very simple Borda client.

Pending tasks:

- [x] API to add a single measurement at a time
- [x] Basic Borda client and
- [x] Pre-aggregation
- [ ] Integrate with chained->fronting fallback mechanism 
- [ ] Sending chunks in a single HTTP request remains pending of the Borda API. Seems secondary for an MVP, or at least can be considered a feature addition.


src/github.com/getlantern/flashlight/logging/borda.go
```diff
@@ -0,0 +1,130 @@
...
+		if err := b.sendMeasurement(b.mBuf[i]); err != nil {
+			// Log as usual
+			log.Errorf("Error sending Measurement to Borda: %v", err)
+		}
+	}
+	return nil
+}
+
+func (b *BordaReporter) sendMeasurement(m *Measurement) error {
+	jbytes, err := json.Marshal(*m)
```
oxtoacart: You could save memory and allocation time by using a pipe as described [here](https://medium.com/stupid-gopher-tricks/streaming-data-in-go-without-buffering-3285ddd2a1e5#.x192d0g0l).  That example misses proper error reporting, but you could add that with channels.

src/github.com/getlantern/flashlight/logging/borda_test.go
```diff
@@ -0,0 +1,90 @@
...
+	sent, err := bc.AddMeasurement(m)
+	assert.Nil(t, err, "Shouldn't return an error")
+	assert.True(t, sent, "Should had sent the measurements")
+}
+
+func newMockServer() *httptest.Server {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		httputil.DumpRequest(r, true)
+
+		dump, err := httputil.DumpRequest(r, true)
```
oxtoacart: Would be nice to actually check that the request contained what's expected.

oxtoacart:
Nice!  The other big thing we need to add is pre-aggregation.  We could do that very specifically for the Error type that @fffw defined and build a map of Measurements.

src/github.com/getlantern/flashlight/logging/borda.go
uaalto: Great tip! I'll do it.

src/github.com/getlantern/flashlight/logging/borda_test.go
uaalto: Indeed!

uaalto:
The aggregation is simple and different approaches could be taken. I've tried to follow the simplest solution that makes sense: build a new error type that can share common data between instances and report that with the count of instances.
Some tests for it would be nice too.

uaalto:
We can see this on Monday @oxtoacart  and @fffw  . It isn't complete, but it is an integrated first approach that depends only on freezing/merging Borda + Structured Errors.

src/github.com/getlantern/flashlight/logging/borda.go
```diff
@@ -0,0 +1,139 @@
...
+		mBuf:    make([]*Measurement, opts.MaxChunkSize),
+	}
+}
+
+func (b *BordaReporter) AddMeasurement(m *Measurement) (sent bool, err error) {
+	b.mBuf[b.nMeas] = m
+	b.nMeas++
+
+	sent = false
+	if b.nMeas >= b.options.MaxChunkSize {
```
fffw: To output in time and accurate report, we probably also need a timer that for every 10 minutes or so, errors should be sent even the buffer is not full.

src/github.com/getlantern/flashlight/logging/error_aggregator.go
```diff
@@ -0,0 +1,176 @@
+package logging
```
fffw: Aggregating errors is tricky as different error has different set of fields available. Error volume should be very low in normal time, but will burst when there's connectivity issue or server is blocked. Rate limiting may be more useful in that case.

fffw:
I'm thinking of a rate limiter that itself reports an error entry with the error it supresses with a count.
```
[
  {"desc": "EOF", "go-type: "io.EOF", "proxy-addr": "x.x.x.x", "device-id": xxx,...}...
  {"desc": "error supressed", "error": "EOF", "supress-count": 35, "device-id": xxx}
]
```
An overall statistics doesn't include the supressed count, which is ok because same error appears 5 times or 50 times doesn't affect the qualitative result. When we want to list errors for one device or user chronologically, the supressed count is there.

We probably don't need it or the aggregator ATM.

src/github.com/getlantern/flashlight/logging/borda.go
uaalto: Regarding "chunking" we still have to decide whether Borda is going to support multiple measurements at the same time or not. If it doesn't, we can just remove chunking altogether. This part is definitely not done.
The timer part goes into error aggregation actually. So, if there is no endpoint for multiple errors, we can just remove this code.

src/github.com/getlantern/flashlight/logging/error_aggregator.go
uaalto: You are more aware of the types of errors there are. Do you think aggregating some classes is worth it? Can we identify them and work with those?

oxtoacart:
@fffw @uaalto Sorry about my recent silence, been heads down integrating errors and context.  My next step this morning will be to pull in this to have a complete solution.  I'm not aiming to have all errors instrumented, just a few of the most interesting ones.  We could even go live with that and see what we learn.

oxtoacart:
Closing in favor of https://github.com/getlantern/lantern/pull/4363

PR #4294, opened 2016-05-15
[WIP] Using better gls library to provide context rather than attaching it to errors

oxtoacart:
This is an improved version of #4293 that doesn't leak context information and uses a more maintained gls library from https://github.com/jtolds/gls/

cc: @fffw @uaalto

### DIFF
--- a/src/github.com/getlantern/flashlight/client/handler.go
+++ b/src/github.com/getlantern/flashlight/client/handler.go
@@ -3,12 +3,15 @@ package client
 import (
 	"fmt"
 	"io"
+	"math/rand"
 	"net"
 	"net/http"
 	"net/http/httputil"
 	"strconv"
 	"sync"
 	"time"
+
+	"github.com/getlantern/golog"
 
 	"github.com/getlantern/flashlight/logging"
 )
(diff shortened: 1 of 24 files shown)


src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -21,8 +24,20 @@ const (
 // handler available from getHandler() and latest ReverseProxy available from
 // getReverseProxy().
 func (client *Client) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
-	logging.RegisterUserAgent(req.Header.Get("User-Agent"))
+	userAgent := req.Header.Get("User-Agent")
+	logging.RegisterUserAgent(userAgent)
+
+	golog.WithContext(golog.Map{
```
oxtoacart: Instead of using defer syntax, this uses callback functions.  This has the benefit of making the context boundaries clear and not leaking data.

Also, this version of gls creates nested contexts that contain parent context information but that don't leak their own information back into the parent.

src/github.com/getlantern/golog/golog.go
```diff
@@ -55,9 +59,9 @@ type Logger interface {
 	Debugf(message string, args ...interface{})
 
 	// Error logs to stderr
-	Error(arg interface{})
+	Error(arg interface{}) error
```
oxtoacart: Small convenience change to interface.  This allows us to log and return the error in one step.  The Errorf version of this replaces:

```golang
err := fmt.Errorf("Oh oh %v", arg)
log.Error(err)
return nil, err
```

with

```golang
return nil, log.Errorf("Oh oh %v", arg)
```

oxtoacart:
Closing in lieu of #4295.

PR #4295, opened 2016-05-15
[WIP] Contextual logging with nested contexts and nice defer() syntax

oxtoacart:
cc: @fffw @uaalto 

The callback functions in #4294 were annoying.  I ended up creating our own context library based on the other two that gives us properly nested contexts but using a `defer context.Exit()` syntax.  I think that's the best of both worlds.

See #4293 for my rationale in going in this direction.


oxtoacart:
Here's what the log looks like now.

```
May 15 07:38:00.104 - 8m15s ERROR balancer: balancer.go:81 Unable to dial via (trusted) chained proxy at 45.55.130.104:443 [] to connect://logs-01.loggly.com:443: Unable to dial server (trusted) chained proxy at 45.55.130.104:443 []: EOF on pass 2...continuing [detour=false op=proxy origin=logs-01.loggly.com:443 request_id=8203977569896484726 user_agent=Go-http-client/1.1]
May 15 07:38:00.105 - 8m15s DEBUG flashlight.client: handler.go:132 Could not dial Still unable to dial connect://logs-01.loggly.com:443 after 3 attempts [op=proxy origin=logs-01.loggly.com:443 request_id=8203977569896484726 user_agent=Go-http-client/1.1]
```

fffw:
What's the purpose of `Global`? If it's only used in `golog`, why not just set it as a variable of `golog` itself?

oxtoacart:
'global' is just a concept of global contextual state.

src/github.com/getlantern/context/context.go
```diff
@@ -0,0 +1,232 @@
...
+	return m
+}
+
+func (c *Context) fill(m Map) {
+	for ctx := c; ctx != nil; {
+		ctx.mx.RLock()
+		fill(m, ctx.data)
+		next := ctx.parent
+		if next == nil {
+			next = ctx.branchedFrom
```
fffw: `branchedFrom` is only used here, and only get set when spawning new Goroutine, in which case `parent` is always nil. Merge to single field?

fffw:
My understanding is that `global` is what unrelated to any `context`. It feels to me like global variables. Without `global` related methods the API can be much more concise.

Say if we want to retrieve current user settings before report error. Why not call a `Settings` service in a central place and merge the result with error details? It's more explicit than `PutGlobalDynamic` somewhere else and retrieve the value through a string key.

src/github.com/getlantern/context/context.go
fffw: In fact I need to carefully scan the code to find the meaning of `branchedFrom` and `parent`.

oxtoacart:
> Say if we want to retrieve current user settings before report error. Why not call a Settings service in a central place and merge the result with error details? It's more explicit than PutGlobalDynamic somewhere else and retrieve the value through a string key.

Theoretically, something may want to add a non-exported value to the global context.  This way, it can do that without having a coupling between the error logging service and whatever knows about that context.

The way I think about it, the global scope is the base of any context, and then additional contextual information is added on top of that.  Originally, I was actually going to have a single global context in the app and just Enter/Exit from there, but passing that around the application would be difficult.

oxtoacart:
It's useful to have global context in the same module as the other context stuff because it follows the same scoping rules.  That is to say, for a given key, the value indicated is looked up via the following chain:

Contextual interface -> Current Context -> Parent Context (etc) -> Global

src/github.com/getlantern/flashlight/client/pluggable_transports.go
```diff
@@ -82,10 +87,21 @@ func obfs4DialFactory(s *ChainedServerInfo, deviceID string) (dialFN, error) {
...
 	return func() (net.Conn, error) {
-		return cf.Dial("tcp", s.Addr, net.Dial, args)
+		defer proxyContext(s.Addr, "obfs4").Exit()
+		conn, err := cf.Dial("tcp", s.Addr, net.Dial, args)
+		return conn, log.IfError(err)
 	}, nil
 }
+
+func proxyContext(addr string, protocol string) *context.Context {
+	ctx := context.Enter().Put("proxy_protocol", protocol)
```
uaalto: Why there is no `.Exit()` in this one or deferred?

uaalto: I see, you need to do it in the calls to proxyContext(). However, I find this unintuitive and potentially prone to developer errors, as it seems an inconsistency. What's the origin of this?

src/github.com/getlantern/context/context.go
oxtoacart: We can't merge them.  We use parent when unwinding the stack on the current goroutine.  Once we reach parent == nil, we stop unwinding the stack.  Each goroutine needs to have its own stack, because parent goroutines can terminate before the child ones.

src/github.com/getlantern/flashlight/client/pluggable_transports.go
oxtoacart: You can't call defer ctx.Exit() in here because it will exit the context as soon as proxyContext returns.  It's moot anyway because the most recent implementation of this code on another branch doesn't use the proxyContext method anymore.

PR #4311, opened 2016-05-17
[WIP] Error reporting from client to borda working end to end

oxtoacart:
@fffw @uaalto Sorry for all the PRs.  This will be the final one.  It integrates @fffw's errors package and @uaalto's borla reporting enhancements, plus my context package.

In its current state, I was able to confirm that errors 

Here's what changed from the original error and borda code:

- Errors automatically capture the context in which they were raised
- Helper methods moved from errors to flashlight/context
- Aggregation in borda is no longer specific to the error data model and it now has counting logic (however the counting logic is hardcoded)
- Borda submissions are based on a timer rather than a batch size
- The borda buffer is bounded so that if we get more during the time window than we can buffer, we start discarding stuff. With this plus the counting logic, there is no need for additional rate limiting.
- I changed the borda REST API to accept lists of measurements which is what the client now passes
- I'm capturing fewer fields than we had originally so that aggregation in borda actually has the opportunity to collapse data.  For example, if we captured request_id, pretty much every error would result in its own measurement and there would be no point to pre-aggregating.  I really think the pre-aggregation is important because otherwise we are sending too much data to the server.

What's left:

- [ ] Make sure Loggly logging works in practice
- [x] Update/fix unit tests for errors
- [x] Update/fix unit tests for borda
- [x] Clean up field names and identify appropriate dimension in borda
- [x] Harmonize error messages and op names (maybe use a convention of codes as originally discussed)
- [ ] Add flag to disable use of borda so that we can merge as is
- [x] Generally review logged errors to make sure that stuff works
- [x] Reenable stack information on errors
- [x] Make counting logic in borda more generic
- [x] Add measurements for success counts
- [x] Add proxy-side logging
- ???

@myleshorton FYI, @fffw and @uaalto worked over the weekend to get us to a point where we could actually integrate all this stuff and have at least a working POC.


oxtoacart:
Fields from which to choose indexed dimensions:

```
application
appversion
clientip
country
detour
domain
error
error_count
error_type
goversion
hostname
http_request_host
http_request_method
http_request_proto
instanceid
locale_country
locale_language
op
origin_host
origin_port
osarch
osname
osversion
process
proxy_host
proxy_port
proxy_type
timezone
user_agent
```

oxtoacart:
Closing in favor of https://github.com/getlantern/lantern/pull/4363/files

PR #4322, opened 2016-05-18
check chained server through ping or recently proxied site fix #4268

### LINKED ISSUE #4268: Re-enable chained proxy checks
Checks are disabled [here](https://github.com/getlantern/lantern/blob/devel/src/github.com/getlantern/balancer/dialer.go#L78) but should be enabled certainly for pro.

They were disabled because we were effectively DDOSing servers and were blacklisting clients. Some limited discussion at #4261

fffw:


src/github.com/getlantern/balancer/dialer.go
```diff
@@ -60,7 +56,7 @@ type dialer struct {
 func (d *dialer) Start() {
 	d.consecSuccesses = 1 // be optimistic
 	d.closeCh = make(chan struct{})
-	d.checkTimer = time.NewTimer(maxCheckTimeout)
+	d.checkTimer = time.NewTimer(MaxCheckTimeout)
```
oxtoacart: While we're in here, it would be nice to randomize this like we do with the config fetcher.

src/github.com/getlantern/flashlight/client/chained.go
```diff
@@ -4,18 +4,24 @@ import (
...
+	"github.com/getlantern/withtimeout"
 )
 
 // Close connections idle for a period to avoid dangling connections.
 // 1 hour is long enough to avoid interrupt normal connections but short enough
 // to eliminate "too many open files" error.
 var idleTimeout = 1 * time.Hour
 
+// Lantern internal sites won't be used as check target.
+var internalSiteSuffix = "getiantem.org"
```
oxtoacart: Should probably look for getlantern.org and lantern.io as well.

src/github.com/getlantern/flashlight/client/chained.go
```diff
@@ -103,6 +105,79 @@ func (s *ChainedServerInfo) Dialer(deviceID string) (*balancer.Dialer, error) {
...
+	ok, timedOut, _ := withtimeout.Do(60*time.Second, func() (interface{}, error) {
+		resp, err := rt.RoundTrip(req)
+		if err != nil {
+			log.Debugf("Error testing dialer %s to %s: %s", s.Addr, url, err)
+			return false, nil
+		}
+		if err := resp.Body.Close(); err != nil {
+			log.Debugf("Unable to close response body: %v", err)
+		}
+		good := resp.StatusCode < 500
```
oxtoacart: This is a little unusual, but I guess the point is that you're just looking for any response from the origin.  Might be good to put a comment here about it.

oxtoacart: One problem here is that we don't check the status code when setting the checkTarget, so if we originally got a bad gateway, we're just going to keep getting a bad gateway error.

oxtoacart:
These changes look good, and the idea of checking using the last site is a clever way to distribute the load.

The reason that we disabled checking originally is that in some cases, the checker was creating TCP connections but not actually doing meaningful work, which caused the client to eventually get blacklisted for looking like it's DOS'ing us.  Did you figure out how/why that's happening and if so, how does this PR fix that?

src/github.com/getlantern/flashlight/client/chained.go
fffw: Thanks! Will add them.

src/github.com/getlantern/flashlight/client/chained.go
fffw: Nice suggestion. I should add comments for both. To reduce the chance of an unreachable site, we can remember recent sites, check against them in round robin, and remove those failed ones.

fffw:
> Did you figure out how/why that's happening and if so, how does this PR fix that?
Not yet but it's my priority in next days. I carefully scanned the code but didn't find a clue. I couldn't reproduce either.

src/github.com/getlantern/flashlight/client/chained.go
oxtoacart: YW

src/github.com/getlantern/flashlight/client/chained.go
oxtoacart: That's a great solution!

fffw:
It's not easy to test the check logic unless we launch a real chained server during the test, which falls into integration test. I chose to skip it. It looks like that coveralls failed due to a decreased coverage?

oxtoacart:
Yeah, that decreased coverage warning is a little annoying.  I've updated our configuration to only complain if coverage decreases more than 1%.  You can ignore the current warning.

> I chose to skip it.

What does this mean?

fffw:
> > I chose to skip it.

> What does this mean?

I simply mean to not add test for the check logic.

fffw:
The next plan is to stop checking dialer regularly if the connection is idle. Instead, having balancer to initiate a check for all dialers when its `Dial` is called after a period without activity. Dialer check is only meaningful to comparing metrics among dialers, when there's some activity.

oxtoacart:
> Dialer check is only meaningful to comparing metrics among dialers, when there's some activity.

I like that.

src/github.com/getlantern/balancer/balancer_test.go
```diff
@@ -20,6 +20,7 @@ var (
 
 func init() {
 	rand.Seed(time.Now().UnixNano())
+	nextCheckFactor = 100 * time.Millisecond
```
fffw: Have to put it here, or we'll get race condition during test.

src/github.com/getlantern/balancer/balancer_test.go
```diff
@@ -150,26 +151,53 @@ func TestTrusted(t *testing.T) {
...
+			return nil, nil
+		},
+		Check: func() bool {
+			wg.Done()
+			return atomic.LoadUint32(&failToDial) == 0
+		},
+		Trusted: true,
+	}
+	bal := New(Sticky, d)
+
```
fffw: So below is various conditions that a dialer is checked.

fffw: So below is various conditions when a dialer is checked.

src/github.com/getlantern/balancer/dialer.go
```diff
@@ -127,61 +135,40 @@ func (d *dialer) updateEMADialTime(t time.Duration) {
...
 	newCF := atomic.AddInt32(&d.consecFailures, 1)
 	log.Tracef("Dialer %s consecutive failures: %d -> %d", d.Label, newCF-1, newCF)
-	nextCheck := time.Duration(newCF*newCF) * 100 * time.Millisecond
-	if nextCheck > maxCheckTimeout {
-		nextCheck = maxCheckTimeout
+	// Don't bother to recheck if dialer is constantly failing.
+	// Balancer will recheck when there's traffic after idle for some time.
+	if newCF < ServerBlacklistingThreshold/2 {
+		atomic.StoreInt32(&d.consecSuccesses, 0)
+		nextCheck := randomize(time.Duration(newCF*newCF) * nextCheckFactor)
```
fffw: When you run tests, you'll see that `nextCheck` will rapidly increase with `newCF`, even without the dialer actually has a chance to be checked. I think it's fine. More consecutive failures mean that it's more likely to fail again recently, while only one consecutive failure may be temporary network glitch.

If network temporarily goes down for a few seconds and comes up again, there can be a maximum of 5(`ServerBlacklistingThreshold/2`)*5*10 (`nextCheckFactor`) = 250 seconds delay for a dialer to refresh its metrics if other dialers take precedence. It should rarely happen as in that situation, other dialers also fail.

fffw:
@oxtoacart I made the changes per the plan. Could you have a look when you have time?

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -63,6 +73,13 @@ func (b *Balancer) OnRequest(req *http.Request) {
 // either manages to connect, or runs out of dialers in which case it returns an
 // error.
 func (b *Balancer) Dial(network, addr string) (net.Conn, error) {
+	lastDialTime := b.lastDialTime.Load().(time.Time)
+	idled := time.Since(lastDialTime)
+	if idled > RecheckAfterIdleFor {
+		log.Debugf("Balancer idled for %s, start checking all dialers", idled)
+		b.checkDialers()
```
oxtoacart: This results in the same dialer that's used by the current Dial immediately being checked.  Seems redundant, though not particularly bad.

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -17,14 +19,20 @@ const (
 )
 
 var (
+	// When Dial() is called after an idle period larger than
+	// RecheckAfterIdleFor, Balancer will recheck all dialers to make sure they
+	// are alive and have up-to-date metrics.
+	RecheckAfterIdleFor = 1 * time.Minute
```
oxtoacart: Does this need to be exported?

oxtoacart: It implies that it can be changed.  To change it, we'd want to protect it with a mutex, right?

src/github.com/getlantern/flashlight/client/chained.go
```diff
@@ -96,6 +118,7 @@ func (s *ChainedServerInfo) Dialer(deviceID string) (*balancer.Dialer, error) {
 				return nil, err
 			}
 
+			s.addCheckTarget(addr)
```
oxtoacart: Just because we were able to dial the proxy doesn't mean that this is a good check target.  I don't think it matters that much, but wanted to make sure that's clear.

src/github.com/getlantern/flashlight/client/chained.go
```diff
@@ -105,6 +128,117 @@ func (s *ChainedServerInfo) Dialer(deviceID string) (*balancer.Dialer, error) {
...
+	rt := &http.Transport{
+		DisableKeepAlives: true,
+		Dial:              dial,
+	}
+	var url string
+	checkTarget := s.checkTargets.get()
+	if checkTarget == "" {
+		url = "http://ping-chained-server"
+	} else {
+		url = fmt.Sprintf("http://%s/index.html", checkTarget)
```
oxtoacart: I don't think this will work if the original request was an HTTPS request.

oxtoacart: One solution would be to just issue a CONNECT request to the check target and consider it successful if the proxy returns 200.  At that point, we know the proxy could establish a connection to the origin, which is probably good enough.  And that works no matter what the user-agent was actually originally doing with the check target.

src/github.com/getlantern/flashlight/client/chained.go
```diff
@@ -105,6 +128,117 @@ func (s *ChainedServerInfo) Dialer(deviceID string) (*balancer.Dialer, error) {
...
+}
+
+func (s *chainedServer) addCheckTarget(addr string) {
+	host, port, e := net.SplitHostPort(addr)
+	if e != nil {
+		log.Errorf("failed to split port from %s", addr)
+		return
+	}
+	if port != "80" {
+		log.Tracef("Skip setting non-HTTP site %s as check target", addr)
```
oxtoacart: Haha, just saw this.  That addresses my previous comment about non-HTTP check targets :)

oxtoacart:
@fffw This looks good to me (and works in my testing).  I had a couple of little comments that you can address if you want, otherwise please feel free to merge at your convenience.

oxtoacart:
@fffw I like that you added a defensive condition to prevent checks from triggering a blacklisting condition.

src/github.com/getlantern/balancer/balancer.go
fffw: Ha I'm just too lazy to exclude current dialer from the list. `Dial` may also use multiple dialers if first one fails.

src/github.com/getlantern/balancer/balancer.go
fffw: Good catch. It's ok to keep it private and simple ATM.

src/github.com/getlantern/flashlight/client/chained.go
fffw: Yeah. It's even irrelevant if the proxy is dialable or not.

src/github.com/getlantern/flashlight/client/chained.go
fffw: You made me think that we probably want to check both HTTP and HTTPS. HTTPS is way more popular while HTTP forwarding is more sophisticated. Maybe not over implement this time.

fffw:
Thanks @oxtoacart ! Merging.

PR #4363, opened 2016-05-24
Final borda/error reporting PR

oxtoacart:
Really, I mean it this time!

Here's what we have:

* errors package is used to provide nice errors with contextual information and stack traces
* golog package now shows stacktraces when available
* context package is used to provide general context
* ops package (which wraps context) is used to track nested operations like proxy -> dial chained and record whether or not they failed to borda
* borda/client provides a client that sends pre-aggregated, rate-limited measurements to the borda server

Unlike the previous version of this, reporting to borda does not go through golog because after discussing with @uaalto, we realized that borda is not a log aggregator like Loggly but rather a health metrics monitoring system.

There's a related server-side PR - https://github.com/getlantern/http-proxy-lantern/pull/77.  These can be merged independently, but the most interesting data comes from having both in play.


fffw:
Tremendous! Need some time to scan carefully and nitpick.

oxtoacart:
Thanks @fffw, it is a big one.  I think @myleshorton would say that the priority right now is on pro stuff, so I wouldn't worry about this one too much unless you have extra time available :)

src/github.com/getlantern/flashlight/proxied/proxied.go
```diff
@@ -154,11 +163,16 @@ func (df *dualFetcher) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	op := ops.Enter("dualfetcher").Request(req)
+	defer op.Exit()
+
+	parallel := df.cf.parallel
+	frontedURL := req.Header.Get(lanternFrontedURL)
+	req.Header.Del(lanternFrontedURL)
 
 	if frontedURL == "" {
-		return nil, errors.New("Callers MUST specify the fronted URL in the Lantern-Fronted-URL header")
+		return nil, op.Error(errors.New("Callers MUST specify the fronted URL in the Lantern-Fronted-URL header"))
```
fffw: `op.Enter()` will change current context right? `errors.New()` also takes current context into consideration. IIUC `op.Error()` here is unnecessary.

I do prefer to avoid any implicit correlation between `errors` and `ops`, but if we can't avoid it, we'd better disallow explicit link between them to avoid confusing.

fffw: Ok, I see. `op.Error()` is not to construct an error object, but to mark the op as failed. I suggest to rename it to `op.Failed()` and without chaining. Reducing the mental burden is way more important than saving lines of code.

src/github.com/getlantern/ops/ops.go
```diff
@@ -0,0 +1,147 @@
...
+		_failure := o.failure.Load()
+		ctx := o.ctx.AsMap(_failure, true)
+		if _failure != nil {
+			failure = _failure.(error)
+			_, errorSet := ctx["error"]
+			if !errorSet {
+				ctx["error"] = failure.Error()
+			}
+		}
+		for _, reporter := range reportersCopy {
```
fffw: Without scanning the code, I never know that `op.Exit` reports failure. We probably need to totally separate those concept entities.

fffw:
My main concern is that the interrelation among those packages adds confusion. So here's my suggestion:

1. Add a new method `context.Current()`, to retrieve current context at any place, so we don't need to pass `op` around ([here](https://github.com/getlantern/lantern/pull/4363/files#diff-4a9300068d35a2ca92619d266ea30e2fR147) for example).

2. Substitute `ops` and `flashlight/ops` with `flashlight/context`, to wrap `context` package with a few helper methods. `context` has a clearer meaning than `ops` IMO.

3. Have a separate package, say, `errpt`, to report error at certain place. It should know nothing about `context`.

4. Add a `WithContext` method to `errors.Error`. I don't hold a strong opinion on it though.

src/github.com/getlantern/flashlight/proxied/proxied.go
oxtoacart: Error doesn't necessarily fail, only if there's actually an error.  How about naming it `FailIferror`?

oxtoacart:
@fffw Thanks for taking a look!

Funnily enough, most of these were deliberately decided opposite of what you suggest.  The purpose of "op" isn't just to provide context, but to actually model operations on whose success or failure we want to report.  A single Op can either succeed or fail, and we count it as one success or failure.

Let me use a concrete example.  Proxying a request is an op which I call `proxy`.  Within this, we have a `dial_for_balancer` op which dials a chained proxy for the balancer.  Because we can have multiple proxies, and because balancer is willing to retry dialing a few times, a single `proxy` op corresponds to multiple `dial_for_balancer` ops.  The results of these could look like this:

```
proxy -> success
    dial_for_balancer (server a) -> fail
    dial_for_balancer (server b) -> fail
    dial_for_balancer (server a) -> success
```

In borda, I want these to be counted as 2 successes and 2 failures, under each of the respective ops.

With that context, let me address your specific points:

> Add a new method context.Current(), to retrieve current context at any place, so we don't need to pass op around (here for example).

I intentionally don't provide access to the current context because I want to be explicit about which op we're marking as failed.

> Substitute ops and flashlight/ops with flashlight/context, to wrap context package with a few helper methods. context has a clearer meaning than ops IMO.

Yes, I've also considered collapsing the two.  I leaned towards keeping them separate because context is a very useful low-level package for tracking contextual information, while ops actually encapsulates logic about reporting on the success and failure of operations.

> Have a separate package, say, errpt, to report error at certain place. It should know nothing about context.

I think you're thinking of this as a reporting thing, but I don't think of this as reporting, I think of it as counting metrics on successes and failures of operations in the system.  When we're in the context of an operation, I want to explicitly mark the operation as failed and then report that as a single failure when the op finishes.  Also, if the op was never marked as failed, I want to report that as a success when the op finishes.  Counting successes if important so that we can understand failure rates, not just see that failures occurred.  That's one of my big frustrations with Loggly - it tells me that we had X dial failures during some time period, but I have no idea whether that's important or not because I don't know how many dial successes we had.

> Add a WithContext method to errors.Error. I don't hold a strong opinion on it though.

I don't understand this one.  What would that achieve?

fffw:
Cool! Having an `op` abstraction sounds great to me. We probably can have the method names to reflect the intention more apparently, say, `op := ops.Start("dial_for_balancer")`, `defer op.Complete()`, `op.Succeed()` (by default but can also call it explicitly), `op.Fail(err)`, etc.

If we always report error when an op is completed with failure, we can decouple `errors` package from `context` completely. Whenever we need to attach context to an error, the purpose is to call `op.Fail()` or `log.Error` immediately. We don't need to propagate errors with context, as `op.Fail()` can serve that purpose: it won't report until `op.Complete()` is called.

Please ignore my last point. It becomes irrelevant in this new context.

oxtoacart:
@fffw I like your suggestion for the naming.

On the subject of propagating context in errors - we can't control when someone logs an error.  Code at a higher level of the call stack may choose to log an error, in which case we want to see its context.  I suggest that we leave the context attached to the error so that it's available there.

I'll make an update to the names and will also reverse merge some stuff from devel and let you know when it's ready.

oxtoacart:
@fffw I thought about it, and I actually want to stick with the naming of ops methods as is.  I tried using something like "Begin" and "End", but that sounded too much like we were actually starting and ending something that does actual work.  Keeping names like "Enter" and "Exit" make it clear that the op is really just a contextual thing, not an actual operation.  What do you think?

fffw:
@oxtoacart Maybe due to language barrier, `Enter` and `Exit` an `op` makes no sense to me. FMPOV, we can optionally attach some `context` to an `op` but the `op` itself is not contextual thing. `Begin`, `End` and `Failed` are very clear semantically. They are pretty much like what in Basic or Pascal: They just mark a block of code which does actual work.

src/github.com/getlantern/flashlight/ops/ops.go
```diff
@@ -27,12 +27,12 @@ type Op struct {
 }
 
 // Enter mimics the similar method from ops.Op
-func (op *Op) Enter(name string) *Op {
+func (op *Op) Begin(name string) *Op {
```
oxtoacart: @fffw You convinced me.  I've applied the changes here in flashlight, and I'll apply them to the core ops library next.

src/github.com/getlantern/flashlight/ops/ops.go
```diff
@@ -52,40 +52,40 @@ func Go(fn func()) {
...
 
 // PutGlobalDynamic mimics the similar method from ops
-func PutGlobalDynamic(key string, valueFN func() interface{}) {
+func SetGlobalDynamic(key string, valueFN func() interface{}) {
 	ops.PutGlobalDynamic(key, valueFN)
 }
 
 // Error mimics the similar method from ops.op
-func (op *Op) Error(err error) error {
+func (op *Op) FailIf(err error) error {
```
oxtoacart: @fffw Note that this is named `FailIf` since it only fails if there's an error.

oxtoacart:
@fffw I've made the API name updates to ops.  Anything else I need to do?

fffw:
Sorry @oxtoacart I lagged behind the plan today. Will look at it tomorrow!

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -107,7 +108,7 @@ func (b *Balancer) Dial(network, addr string) (net.Conn, error) {
 		log.Tracef("Dialing %s://%s with %s", network, addr, d.Label)
 		conn, err := d.dial(network, addr)
 		if err != nil {
-			log.Errorf("Unable to dial via %v to %s://%s: %v on pass %v...continuing", d.Label, network, addr, err, i)
+			log.Error(errors.New("Unable to dial via %v to %s://%s: %v on pass %v...continuing", d.Label, network, addr, err, i))
```
fffw: It's very impressive seeing a single line of error to produce comprehensive context and stack traces.

src/github.com/getlantern/borda/borda.go
```diff
@@ -0,0 +1,246 @@
...
+	//            "connected_to_internet": true }
+	Dimensions map[string]interface{} `json:"dimensions,omitempty"`
+}
+
+// WriteFunc is a function that writes a batch to the database
+type WriteFunc func(client.BatchPoints) error
+
+// Collector collects Measurements
+type Collector interface {
+	http.Handler
```
fffw: It would be better to comment here why it embeds `http.Handler`

fffw: As there's no strong tie between `Collector` and `http.Handler`, would it be better to have a method `AsHTTPHandler() http.Handler` instead of embedding it directly?

src/github.com/getlantern/borda/borda.go
```diff
@@ -0,0 +1,246 @@
...
+
+	// Submit submits a measurement to the Collector
+	Submit(*Measurement)
+
+	// Wait waits up to timeout for the Collector to finish running and returns
+	// the error that caused the Collector to terminate. A timeout of -1 causes
+	// Wait to block indefinitely.
+	Wait(timeout time.Duration) error
```
fffw: In my understanding, a `Collector` should continuously run once started and until being told to stop. I saw `Wait()` only called in standalone application and tests. If something prevents collector from working, say, when network is totally down for a while, it's better to drop what can't be submitted and continue working, or the caller has to care about `Wait` and restart it again.

src/github.com/getlantern/borda/borda.go
```diff
@@ -0,0 +1,246 @@
...
+					if isDimension && c.indexedDimensions[key] {
+						var stringValue string
+						switch v := value.(type) {
+						case string:
+							stringValue = v
+						default:
+							stringValue = fmt.Sprint(v)
+						}
+						tags[key] = stringValue
+						fields["_"+key] = value
```
fffw: It would increase the size of the measurement a lot. Unlike tags which is de-duplicated compound by nature, fields are are stored line by line. If the key is designated as dimension, it's probably safe to just store it as tag.

src/github.com/getlantern/errors/hide.go
```diff
@@ -0,0 +1,47 @@
...
+	binary.BigEndian.PutUint64(b, nextID)
+	e.id = nextID
+	e.hiddenID = hidden.ToString(b)
+	hiddenErrors[idxForID(nextID)] = e
+	nextID++
+	hiddenMutex.Unlock()
+}
+
+func get(hiddenID []byte) *structured {
+	id := binary.BigEndian.Uint64(hiddenID)
```
fffw: It will panic if `hiddenID` is less than 8 bytes. Just a hint. Should be safe to ignore the check in our case.

src/github.com/getlantern/flashlight/app/integration_test.go
```diff
@@ -81,10 +81,10 @@ func TestProxying(t *testing.T) {
 	// Makes a test request
 	testRequest(t, httpAddr, httpsAddr)
 
-	// Switch to obfs4, wait for a new config and test request again
-	atomic.StoreUint32(&useOBFS4, 1)
-	time.Sleep(2 * time.Second)
-	testRequest(t, httpAddr, httpsAddr)
+	// // Switch to obfs4, wait for a new config and test request again
```
fffw: Why comment tests here?

src/github.com/getlantern/ops/ops.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (o *op) FailIf(err error) error {
+	if err != nil {
+		o.failure.Store(err)
+	}
+	return err
+}
+
+func (o *op) Errorf(msg string, args ...interface{}) error {
+	return o.FailIf(fmt.Errorf(msg, args...))
```
fffw: Switch to `errors.New`?

src/github.com/getlantern/ops/ops.go
```diff
@@ -0,0 +1,150 @@
...
+	Set(key string, value interface{}) Op
+
+	// SetDynamic puts a key->value pair into the current Op's context, where the
+	// value is generated by a function that gets evaluated at every Read.
+	SetDynamic(key string, valueFN func() interface{}) Op
+
+	// FailIf marks this Op as failed if the given err is not nil. If FailIf is
+	// called multiple times, the latest error will be reported as the failure.
+	// Returns the original error for convenient chaining.
+	FailIf(err error) error
```
fffw: `FailIf` and `Errorf` may create confusing what is the best way to write error handling code, as there are multiple ways for similar purpose (saw some of them in this PR):
* `log.Errorf(...)`
* `log.Error(errors.New(...))`
* `log.Error(op.FailIf(errors.New(...)))`
* `log.Error(op.Errorf(...)))`

[Another example](https://github.com/getlantern/lantern/pull/4363/files#diff-c27ca30fd23b2036205dc308988a8b0dR123):
```
if err != nil {
  return nil, op.FailIf(err)
}
```
It's awkward to call `FailIf` when `err` is not nil for sure.

As simple wrappers, `FailIf` and `Errorf` creates more problems than solved. I vote to remove `op.Errorf` and rename `op.FailIf()` to `op.Fail()` without returning anything. So someone needs to do multiple things should call each individually, see below: 
```
n, err := doSomething()
if err != nil {
  op.Fail(err)
  log.Error(err)
}
return n, err
```

By provide only one way to do one thing, we make the code more understandable and encourage caller to think a bit on questions like "why return the same error after marking the `op` as `Fail` and logging the error?" or "is it the right place to call `op.Fail`?"

src/github.com/getlantern/testify/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Compiled Object files, Static and Dynamic libs (Shared Objects)
```
fffw: We replaced "getlantern/testify" with "stretchr/testify". It should be removed from repo.

src/github.com/mxk/go-flowrate/LICENSE
```diff
@@ -0,0 +1,29 @@
+Copyright (c) 2014 The Go-FlowRate Authors. All rights reserved.
```
fffw: `go-flowrate` is not used anywhere in client

src/github.com/vharitonsky/iniflags/.gitignore
```diff
@@ -0,0 +1 @@
+tags
```
fffw: `iniflags` is only used by `http-proxy-lantern`. I guess it sneaked in by `go get`

fffw:
@oxtoacart I finished this round. I'm looking forward to have it merged soon once the API is finalized!

src/github.com/getlantern/balancer/balancer.go
oxtoacart: Yeah, we've built something cool here!  Your error abstraction worked out very nicely.

src/github.com/getlantern/borda/borda.go
oxtoacart: Collector is really only meant for use within the server process.  I only made it its own thing for testability reasons.

src/github.com/getlantern/borda/borda.go
oxtoacart: Collector is really a server process, analogous to http.Serve().  However, I did notice that we call terminate at times when we shouldn't (i.e. unable to commit a batch).  I'll fix that.

oxtoacart: Actually, now that I look at it, we're okay.  There are only two reasons this will terminate:

1. Unable to create a batch.  This is not a transient error and we can't recover from it.  In practice, it should never happen.

2. Exhausted allowed retries to commit a batch.  This is a serious condition meaning that we're giving up.  We can set the retries appropriately high.

So, I won't change anything here.

src/github.com/getlantern/borda/borda.go
oxtoacart: Unfortunately we have to do this because we can't run calculations on tags.  For example, we can't run a distinct(<tag>) or max(<tag>).  As you said a while back, the idea is that we throw a bunch of data into the first table, which is time-limited anyway, and then we can filter out stuff we don't need using our continuous queries.

src/github.com/getlantern/errors/hide.go
oxtoacart: Good catch, I'm adding a guard condition.

src/github.com/getlantern/flashlight/app/integration_test.go
oxtoacart: Oh, that was left over from debugging something.  I've uncommented it.  Thanks!

src/github.com/getlantern/ops/ops.go
oxtoacart: This is a silly convenience method.  I'm going to take it out.

src/github.com/getlantern/ops/ops.go
oxtoacart: Errorf was a silly convenience method that just caused confusion, I've removed it.  That should eliminate most of the confusion that you mention.

The confusion between `log.Errorf` and `log.Error(errors.New())` is unrelated to ops.  The decision there is whether or not we want a full stacktrace for the error.  I think I'll update log.Errorf so that it automatically creates an errors.Error instead of just an error.

I still think FailIf is worth keeping because it saves a lot of typing.

src/github.com/getlantern/testify/.gitignore
oxtoacart: Unfortunately, measured and http-proxy still have some tests that rely on getlantern/testify.  I'll fix that and then remove this.

src/github.com/mxk/go-flowrate/LICENSE
oxtoacart: It's used by http-proxy-lantern, which is used in our integration test.

src/github.com/vharitonsky/iniflags/.gitignore
oxtoacart: http-proxy-lantern is used in our integration test, so we need this.

oxtoacart: Actually, that's only needed by the main package, which we're not using here, so I can remove it.  Thanks!

fffw:
Looks pretty good to merge. Awesome work @oxtoacart ! We can tidy up the log messages a bit afterwards.

Personally I still don't quite like the clever trick to chain `log` and `ops` methods together, but let's see how we feel about it in long run.

oxtoacart:
@fffw Thanks for your contributions to getting this monitoring in place, and thanks for reviewing this monster PR!

Issue #4357, opened 2016-05-24
panics for invalid config


fffw:
[User reported](https://groups.google.com/forum/#!topic/lantern-users-zh/JwQuQQJS9D0).

In [config.go#L108](https://github.com/getlantern/lantern/blob/release-2.2.0/src/github.com/getlantern/flashlight/config/config.go#L108) we didn't check if `cfg.Client` is empty.

myleshorton:
This actually isn't possible on devel so should be fixed in 2.3.0.

PR #4364, opened 2016-05-24
add staging flag for hitting staging servers

myleshorton:


src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -59,9 +59,9 @@ type Fetcher interface {
 }
 
 // StartPolling starts the process of polling for new configuration files.
-func StartPolling() {
+func (cfg *Config) StartPolling() {
 	// Force detour to whitelist chained domain
-	u, err := url.Parse(chainedCloudConfigURL)
+	u, err := url.Parse(cfg.CloudConfig)
```
myleshorton: We should really be using the configured server here rather than the default -- made this a method of the struct accordingly.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -78,6 +78,12 @@ func validateConfig(_cfg yamlconf.Config) error {
 	if !ok {
 		return fmt.Errorf("Config is not a flashlight config!")
 	}
+	if cfg.Client == nil {
+		return fmt.Errorf("No client config")
+	}
+	if cfg.Client.ChainedServers == nil {
+		return fmt.Errorf("No chained servers")
+	}
```
myleshorton: This actually just knocks out https://github.com/getlantern/lantern/issues/4357 while we're in here.

src/github.com/getlantern/flashlight/config/fetcher.go
```diff
@@ -16,16 +16,17 @@ import (
...
-	ifNoneMatch           = "X-Lantern-If-None-Match"
-	userIDHeader          = "X-Lantern-User-Id"
-	tokenHeader           = "X-Lantern-Pro-Token"
-	chainedCloudConfigURL = "http://config.getiantem.org/cloud.yaml.gz"
+	etag         = "X-Lantern-Etag"
+	ifNoneMatch  = "X-Lantern-If-None-Match"
+	userIDHeader = "X-Lantern-User-Id"
+	tokenHeader  = "X-Lantern-Pro-Token"
+
+	defaultChainedCloudConfigURL = "http://config.getiantem.org/cloud.yaml.gz"
```
myleshorton: Just a slightly clearer, if longer, name.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -134,7 +134,7 @@ func Run(httpProxyAddr string,
 			// proxying capabilities of Lantern, so it needs everything to be properly
 			// set up with at least an initial bootstrap config (on first run) to
 			// complete successfully.
-			config.StartPolling()
+			cfg.StartPolling()
```
myleshorton: Start polling on the actual config instance.

myleshorton:
hey @oxtoacart you mind taking a look at this one?

src/github.com/getlantern/flashlight/config/config.go
oxtoacart: Shouldn't this have been taken care of by ApplyDefaults()?

myleshorton: Yeah just removed it @oxtoacart -- I think that client-reported error isn't possible with what's on devel actually!

src/github.com/getlantern/flashlight/config/fetcher.go
```diff
@@ -49,8 +57,43 @@ type UserConfig interface {
...
 // NewFetcher creates a new configuration fetcher with the specified
 // interface for obtaining the user ID and token if those are populated.
-func NewFetcher(conf UserConfig, rt http.RoundTripper) Fetcher {
-	return &fetcher{lastCloudConfigETag: map[string]string{}, user: conf, rt: rt}
+func NewFetcher(conf UserConfig, rt http.RoundTripper, flags map[string]interface{}) Fetcher {
+	var stage bool
+	if s, ok := flags["staging"].(bool); ok {
+		stage = s
+	} else {
+		stage = false
```
oxtoacart: This is redundant, as stage is already false.

myleshorton: Good catch should be all set!

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -40,31 +40,25 @@ var (
...
+	MemProfile      string
+	UpdateServerURL string
+	Client          *client.ClientConfig
+	ProxiedSites    *proxiedsites.Config // List of proxied site domains that get routed through Lantern rather than accessed directly
+	TrustedCAs      []*fronted.CA
 }
 
 // StartPolling starts the process of polling for new configuration files.
-func StartPolling() {
+func (cfg *Config) StartPolling() {
```
oxtoacart: Why did you change polling to refer to a specific Config instance?  Polling is a process that continues forever, even though the underlying config instance changes over time.  Also, there's nothing in here that specifically refers to the cfg instance, so there's no benefit to it being an instance method.  Plus it refers to the global yamlconf.Manager.

src/github.com/getlantern/flashlight/flashlight.go
oxtoacart: See comment above - I would revert this to polling on config package.

oxtoacart:
@myleshorton This now works when I test.  Travis test failed for what seems to be an unrelated issue, I've restarted it.

The basic logic looks solid to me, there's just a couple of things:

1. Polling on the Config instance doesn't make any sense
2. There's a redundant else branch assigning false to an already false boolean

Other than those and assuming the test passes this time, this LGTM.

src/github.com/getlantern/flashlight/config/config.go
myleshorton: That from when I started using the chained URL in the config, so it was necessary @oxtoacart. I'll change it back now that there's no chained URL in the config though.

myleshorton:
I changed StartPolling back to not be a method of Config @oxtoacart, and I already fixed the extra else.

PR #4378, opened 2016-05-26
Bypassing disallowed ports in SOCKS proxy

oxtoacart:
This enables mobile to work with proxies that only allow specific ports.

### DIFF
--- a/src/github.com/getlantern/flashlight/client/client.go
+++ b/src/github.com/getlantern/flashlight/client/client.go
@@ -5,16 +5,17 @@ import (
 	"net"
 	"net/http"
 	"reflect"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	"github.com/armon/go-socks5"
-	"github.com/getlantern/balancer"
 	"github.com/getlantern/detour"
 	"github.com/getlantern/eventual"
 	"github.com/getlantern/golog"
+	"github.com/getlantern/protected"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -219,6 +217,47 @@ func (client *Client) proxiedDialer(orig func(network, addr string) (net.Conn, e
 	}
 }
 
+func (client *Client) dialCONNECT(addr string, port int) (net.Conn, error) {
```
oxtoacart: Centralized this logic in one function for use both in HTTP proxy and SOCKS proxy.

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -219,6 +217,47 @@ func (client *Client) proxiedDialer(orig func(network, addr string) (net.Conn, e
...
+			// We signal it to send the CONNECT here using the network transport argument
+			// that is effectively always "tcp" in the end, but we look for this
+			// special "transport" in the dialer and send a CONNECT request in that
+			// case.
+			return client.getBalancer().Dial("connect", addr)
+		})
+		return d("tcp", addr)
+	}
+	log.Tracef("Port not allowed, bypassing proxy and sending CONNECT request directly to %v", addr)
+	return protected.Dial("tcp", addr, 1*time.Minute)
```
oxtoacart: We have to use protected.Dial so that on mobile this bypasses the VPN.

src/github.com/getlantern/protected/protected.go
```diff
@@ -127,6 +132,11 @@ func Resolve(addr string) (*net.TCPAddr, error) {
 //   specified system device (this is primarily
 //   used for Android VpnService routing functionality)
 func Dial(network, addr string, timeout time.Duration) (net.Conn, error) {
+	protect, _ := getCurrent()
+	if protect == nil {
```
oxtoacart: This allows us to use protected.Dial even when we're not on mobile, by simply falling back to net.DialTimeout

src/github.com/getlantern/flashlight/client/client.go
fffw: It's a bit unusual to have common code depending on a package used only on mobile. I don't have better option, though.

fffw:
LGTM

Issue #4282, opened 2016-05-13
Loading screen of Death on windows 


Droyk:
Good evening guys 

I am having a problem https://streamable.com/g2cx

Loading screen of death it never goes away tried everything but still doesn't go away 

using the lantern latest version 2.2.3
I am on windows 10 pro x64
Chrome 50.0.2661.102 m (64-bit)

CHN008:
Maybe you need set it by the "Settings" File,Or Run it  "As Administrator".

![lt-settings](https://cloud.githubusercontent.com/assets/16646697/15252520/da5bedc0-1960-11e6-945a-5fa3449fa628.jpg)

Mine Runs Well,But I Cant't Open The Setting Page.

My Mainly used tool is:"xx-net".

https://github.com/XX-net/XX-Net

myleshorton:
Was it working fine for you before @Droyk?

xiam:
Hello @Droyk, do you have any message on the web console? could you share a fullscreen video from installation until replicating the problem? Thanks.

fthvgb1:
I have a same problem on windows 10,
<img src="http://i4.buimg.com/e9365640b34b3b48.png">
<img src="http://i4.buimg.com/fd13942ca641fe2a.png">
,but it can work normally on ubuntu

Droyk:
@CHN008 I don't have any setting files on the folder
![screenshot_20160514122625](https://cloud.githubusercontent.com/assets/15848350/15266864/8269b790-19cf-11e6-9e11-a44b636b4d49.png)
@myleshorton No actually I am a new user of lantern
@xiam Well there are 2 error messages on console 
![g0gzett](https://cloud.githubusercontent.com/assets/15848350/15266838/af30f5f0-19ce-11e6-8051-3aa72452aa34.png)

xiam:
Thanks @Droyk, that really helps, could you send us another picture of your system proxy settings?

CHN008:
@Droyk I Found It Will Appears Like That On Your Computer When I Am Using Chrome And Firefox,But It Will Be Better OnSouGou Browser.
I Can't Open The Setting Page,But It Helps Surfing Google.

Maybe You Can Run "Any desk" on your computer and we can help you,via the Remote assistance.

Droyk:
[...]

PR #4379, opened 2016-05-26
not serving proxied sites to ui

fffw:
We no longer show or modify proxied sites in UI. Remove relevant code.

It may also helpful for https://github.com/getlantern/lantern/issues/4282. Though we don't exactly know the cause, but inspired by [this SO link](http://stackoverflow.com/questions/31265789/websocket-invalid-frame-header), a large WebSocket frame may trigger "Invalid frame header" error if browser/server has some bug handle it.

@xiam Mind have a look?

### DIFF
--- a/src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go
+++ b/src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go
@@ -1,26 +1,16 @@
 package proxiedsites
 
 import (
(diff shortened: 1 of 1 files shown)


fffw:
`proxiedsites` was the first frame UI received from the WebSocket, that's why I suspect it to be the cause.

xiam:
Not sure if this helps, but we definitely don't need it anymore, thanks @fffw

PR #4408, opened 2016-05-30
Revert "Bypassing disallowed ports in SOCKS proxy"

fffw:
Reverts getlantern/lantern#4378


fffw:
Revert because it fails to build on Windows. `syscall.Handle` is not `int` on Windows.

fffw:
I end up reset https://github.com/getlantern/lantern/pull/4378 directly

PR #4409, opened 2016-05-30
Bypassing disallowed ports in SOCKS proxy

fffw:
Same as https://github.com/getlantern/lantern/pull/4378. Needs more commits to fix the issue building on Windows.

PR #4425, opened 2016-05-31
flashlight client call protect package only on Android

fffw:
@oxtoacart It's an attempt to avoid the SOCKS5 change from failing on Windows.


oxtoacart:
Code LGTM.  How was this failing on Windows?

fffw:
The errors are [here](https://github.com/getlantern/lantern/pull/4409#issuecomment-222494106)

PR #4494, opened 2016-06-02
Dynamic bandwidth tracking

oxtoacart:
In order to test this, you can locally run a proxy that spits out fake bandwidth tracking stats by building from https://github.com/getlantern/http-proxy-lantern/tree/throttling-ox

### DIFF
--- a/lantern-ui/app/js/services.js
+++ b/lantern-ui/app/js/services.js
@@ -90,6 +90,9 @@ angular.module('app.services', [])
           window.location = settings.redirectTo;
         }
       },
+      'bandwidth': function(bandwidth) {
+        console.log('Got bandwidth data: ', bandwidth);
+      },
       'localDiscovery': function(data) {
         model.localLanterns = data;
       },
--- a/src/github.com/getlantern/bandwidth/bandwidth.go
+++ b/src/github.com/getlantern/bandwidth/bandwidth.go
@@ -36,7 +36,13 @@ func GetQuota() *Quota {
 }
 
 // Track updates the bandwith quota information based on the XBQ header in
-// the given response.
+// the given response. The header is expected to follow this format:
+//
+// <used>/<allowed>/<asof>
+//
+// <used> is the string representation of a 64-bit unsigned integer
+// <allowed> is the string representation of a 64-bit unsigned integer
+// <asof> is the 64-bit signed integer representing nanoseconds since epoch
 func Track(resp *http.Response) {
 	xbq := resp.Header.Get("XBQ")
 	if xbq == "" {
--- a/src/github.com/getlantern/bandwidth/bandwidth.go
+++ b/src/github.com/getlantern/bandwidth/bandwidth.go
@@ -17,7 +17,7 @@ var (
 	mutex sync.RWMutex
 
 	// Updates is a channel on which one can receive updates to the Quota
-	Updates = make(chan *Quota)
+	Updates = make(chan *Quota, 100)
 )
 
 // Quota encapsulates information about the user's bandwidth quota.
--- a/src/github.com/getlantern/bandwidth/bandwidth.go
+++ b/src/github.com/getlantern/bandwidth/bandwidth.go
@@ -1,0 +1,80 @@
+package bandwidth
+
+import (
+	"net/http"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/getlantern/golog"
+)
+
+var (
+	log = golog.LoggerFor("bandwidth")
+
+	quota *Quota
+	mutex sync.RWMutex
+
+	// Updates is a channel on which one can receive updates to the Quota
+	Updates = make(chan *Quota)
+)
+
+// Quota encapsulates information about the user's bandwidth quota.
+type Quota struct {
+	MiBAllowed uint64    `json:"mibAllowed"`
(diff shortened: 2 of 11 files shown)


lantern-ui/app/js/services.js
```diff
@@ -90,6 +90,9 @@ angular.module('app.services', [])
           window.location = settings.redirectTo;
         }
       },
+      'bandwidth': function(bandwidth) {
+        console.log('Got bandwidth data: ', bandwidth);
```
oxtoacart: @myleshorton Bandwidth data is plumbed through to web browser.  Can you take it from here and bind into the UI as appropriate?

myleshorton: Definitely @oxtoacart thanks!

src/github.com/getlantern/flashlight/ui/service.go
```diff
@@ -46,7 +45,7 @@ func (s *Service) write() {
 	}
 }
 
-func Register(t string, newMessage func() interface{}, helloFn helloFnType) (*Service, error) {
+func Register(t string, helloFn helloFnType) (*Service, error) {
```
oxtoacart: Opportunistic cleanup.

myleshorton: Nice!

oxtoacart:
The test failure doesn't seem to be isolated to this branch.

oxtoacart:
Yeah, lantern_test is failing on devel and master too.

myleshorton:
wow really @oxtoacart? I wonder how that happened. Interesting. Does that test live proxying, as in hit real servers?

oxtoacart:
@myleshorton Yes, it's hitting real proxies and google.com.

myleshorton:
oh interesting -- I wonder if those proxies are failing...

src/github.com/getlantern/bandwidth/bandwidth.go
```diff
@@ -0,0 +1,80 @@
...
+func GetQuota() *Quota {
+	mutex.RLock()
+	result := quota
+	mutex.RUnlock()
+	return result
+}
+
+// Track updates the bandwith quota information based on the XBQ header in
+// the given response.
+func Track(resp *http.Response) {
```
myleshorton: Can you throw the format we're parsing in a comment here @oxtoacart?

myleshorton: A test would also be ideal but i realize it's late in the game

oxtoacart:
It looks like it's trying to hit a proxy that doesn't exist.

src/github.com/getlantern/bandwidth/bandwidth.go
```diff
@@ -0,0 +1,80 @@
...
+		return
+	}
+	asofInt, err := strconv.ParseInt(parts[2], 10, 64)
+	if err != nil {
+		log.Debugf("Malformed XBQ header %v, can't parse as of time: %v", err)
+		return
+	}
+	asof := time.Unix(0, asofInt)
+	mutex.Lock()
+	if quota == nil || quota.AsOf.Before(asof) {
```
myleshorton: Since we throw this away when the channel is full, @oxtoacart, I wonder if it might be cleaner to just create a new Quota here and not worry about the locking and reusing an existing one.

myleshorton: I realize we use it in GetQuota, but I think we can also just skip the hello function, which is the only place that's used @oxtoacart

src/github.com/getlantern/bandwidth/bandwidth.go
```diff
@@ -0,0 +1,80 @@
...
+	"time"
+
+	"github.com/getlantern/golog"
+)
+
+var (
+	log = golog.LoggerFor("bandwidth")
+
+	quota *Quota
+	mutex sync.RWMutex
```
myleshorton: Seems like we should be able to skip both of the above to me. If we do really want the hello function, couldn't we just fetch the data from the channel there as well?

myleshorton:
This is overall looking great @oxtoacart, just a few comments.

src/github.com/getlantern/bandwidth/bandwidth.go
oxtoacart: I added a comment about the format.

src/github.com/getlantern/bandwidth/bandwidth.go
oxtoacart: I think the hello function is important because that's used every time someone opens the UI. Without that, we have to wait for an actual request to be made to display the usage.  The operation inside this lock is really fast, so I wouldn't worry about performance.

src/github.com/getlantern/bandwidth/bandwidth.go
oxtoacart: We can't fetch the data from the channel because it's not a buffered channel. Even if we made it a buffered channel, it would still be capped, so at some point the channel would be capturing stale data.

oxtoacart:
@myleshorton Thanks for the review.  I've addressed your comments.  I'd like to leave the GetQuota capability in place as it seems useful for giving up-to-date information to new UI clients.

myleshorton:
OK looks good @oxtoacart!

PR #4502, opened 2016-06-03
Whitelisting staging config URL

oxtoacart:
I noticed that we didn't update the config server whitelisting code when we added the staging flag.  This fixes that.


myleshorton:
LGTM!

PR #4524, opened 2016-06-05
When no check function provided for dialer, return true

oxtoacart:
I noticed a lot of log output from the balancer test and saw that the default check function returns false.  That creates unnecessary work since it triggers a recheck, and that recheck can never succeed because the default check function will always return false.


fffw:
Good catch @oxtoacart . LGTM

PR #4525, opened 2016-06-05
Added TLS ClientSessionCache in a few places to reduce TLS handshaking

oxtoacart:
With all these http.Transports that we're creating, we missed a few opportunities to cache TLS sessions, which results in unnecessary handshaking/load to servers.

This probably won't have a huge impact since it's mostly used for our internal polling stuff, but it can't hurt.


myleshorton:
LGTM @oxtoacart. I guess we lost that in direct.go with the proxied refactoring is that right?

PR #4593, opened 2016-06-08
Fix leaking connections

oxtoacart:


src/github.com/getlantern/flashlight/client/chained.go
```diff
@@ -14,10 +14,12 @@ import (
...
-// Close connections idle for a period to avoid dangling connections.
-// 1 hour is long enough to avoid interrupt normal connections but short enough
-// to eliminate "too many open files" error.
-var idleTimeout = 1 * time.Hour
+// Close connections idle for a period to avoid dangling connections. 45 seconds
+// is long enough to avoid interrupt normal connections but shorter than the
+// idle timeout on the server to avoid running into closed connection problems.
+// 45 seconds is also longer than the MaxIdleTime on our http.Transport, so it
+// doesn't interfere with that.
+var idleTimeout = 45 * time.Second
```
oxtoacart: Now that we have an idletiming that returns io.EOF for pending reads and writes instead of "Use of closed network connection", we can make this much shorter.  We make it shorter than the server's idle timeout so that we don't risk sending data on a connection that will subsequently get idled by the server.

I increased the server's idle timeout to 70 seconds in [this commit](https://github.com/getlantern/http-proxy-lantern/commit/0ea7f958212068e2b579860e4e40be9a26a10a40).

src/github.com/getlantern/flashlight/client/chained.go
```diff
@@ -123,10 +125,7 @@ func (s *chainedServer) dialer(deviceID string) (*balancer.Dialer, error) {
 				return nil, op.FailIf(err)
 			}
 			conn = idletiming.Conn(conn, idleTimeout, func() {
-				log.Debugf("Proxy connection to %s via %s idle for %v, closing", addr, conn.RemoteAddr(), idleTimeout)
-				if err := conn.Close(); err != nil {
-					log.Debugf("Unable to close connection: %v", err)
-				}
+				log.Debugf("Proxy connection to %s via %s idle for %v, closed", addr, conn.RemoteAddr(), idleTimeout)
```
oxtoacart: idletiming handles closing for us now.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -70,12 +70,12 @@ func (client *Client) intercept(resp http.ResponseWriter, req *http.Request, op
 	closeConns := func() {
 		if clientConn != nil {
 			if closeErr := clientConn.Close(); closeErr != nil {
-				log.Debugf("Error closing the out connection: %s", closeErr)
+				log.Tracef("Error closing the out connection: %s", closeErr)
```
oxtoacart: These errors were getting annoying, so I turned down the log level.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -114,13 +114,11 @@ func (client *Client) intercept(resp http.ResponseWriter, req *http.Request, op
...
 func pipeData(clientConn net.Conn, connOut net.Conn, op *ops.Op, closeFunc func()) {
-	writeErrCh := make(chan error)
+	writeErrCh := make(chan error, 1)
 	// Start piping from client to proxy
 	op.Go(func() {
 		_, writeErr := io.Copy(connOut, clientConn)
-		if writeErr != nil {
-			writeErrCh <- writeErr
-		}
+		writeErrCh <- writeErr
```
oxtoacart: This was the main source of the leak.  In the case of successful copies, we never pushed anything onto writeErrCh, which would prevent the read down below, which prevented the call to close the connections.

src/github.com/getlantern/flashlight/client/reverseproxy.go
```diff
@@ -18,7 +18,9 @@ import (
 func (client *Client) newReverseProxy(bal *balancer.Balancer) *httputil.ReverseProxy {
 	transport := &http.Transport{
 		TLSHandshakeTimeout: 40 * time.Second,
+		MaxIdleTime:         30 * time.Second,
```
oxtoacart: This requires our custom fork of go.  It prevents us from trying to reuse connections that timed out due to idleness.

src/github.com/getlantern/flashlight/proxied/proxied.go
```diff
@@ -387,6 +387,10 @@ func chained(rootCA string, persistent bool) (http.RoundTripper, error) {
 			ClientSessionCache: clientSessionCache,
 		},
 	}
+	if persistent {
```
oxtoacart: This also prevents us from trying to reuse connections that were timed out due to being idle.

src/github.com/getlantern/idletiming/idletiming_conn.go
```diff
@@ -3,7 +3,10 @@
 package idletiming
```
oxtoacart: The changes to idletiming are being reviewed [here](https://github.com/getlantern/idletiming/pull/3).

src/github.com/getlantern/http-proxy/forward/forward.go
```diff
@@ -47,11 +47,7 @@ func New(opts *Options) filters.Filter {
 				return nil, err
 			}
 
-			idleConn := idletiming.Conn(conn, opts.IdleTimeout, func() {
-				if conn != nil {
-					conn.Close()
-				}
-			})
+			idleConn := idletiming.Conn(conn, opts.IdleTimeout, nil)
```
oxtoacart: The changes to http-proxy are being reviewed [here](https://github.com/getlantern/http-proxy/pull/84).

myleshorton:
This generally looks great @oxtoacart. Does it rely on the order of the merges? I.e. can I merge this independently from the others?

I'm also curious what led down this path. Did you see leaking connections on the client side? Thanks!

oxtoacart:
Yes, i was testing and noticed that i was quickly getting into hundreds of
file descriptors

The whole "use of closed network connection" thing had been nagging at me
for a while too, so I took this as an opportunity to dig into it.


-- 
Cheers,
Ox

PGP Key - https://pgp.mit.edu/pks/lookup?op=get&search=ox@getlantern.org

- Marge Piercy

myleshorton:
very interesting @oxtoacart. so just so I understand, what exactly was causing those "use of closed network connection" errors -- that's been nagging me too on an almost daily basis!

oxtoacart:
@myleshorton The below examples demonstrate what was happening with idletiming.  Basically, a net.Conn is supposed to return io.EOF to pending reads and writes if/when the remote end disconnects.  However, if we close the connection locally (as idletiming does), then pending ops get the "use of closed network connection" error.  The fix was to give pending ops an io.EOF.

Bad
```go
package main

import (
	"io"
	"log"
	"net"
	"os"
	"sync"
	"time"
)

func main() {
	l, err := net.Listen("tcp", "localhost:")
	if err != nil {
		log.Fatal(err)
	}

	go func() {
		for {
			conn, err := l.Accept()
			if err != nil {
				log.Fatal(err)
			}
			conn.Write([]byte("Hello there"))
			time.Sleep(2 * time.Second)
			conn.Close()
		}
	}()

	conn, err := net.Dial("tcp", l.Addr().String())
	if err != nil {
		log.Fatal(err)
	}

	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		b := make([]byte, 10)
		for {
			n, err := conn.Read(b)
			if err != nil {
				if err != io.EOF {
					log.Printf("Got unexpected error %v while reading", err)
				} else {
					os.Stdout.Write(b[:n])
				}
				break
			}
			os.Stdout.Write(b[:n])
		}
		wg.Done()
	}()

	time.Sleep(1 * time.Second)

	// We're closing the connection prematurely, just like idletiming used to
	conn.Close()

	wg.Wait()
}
```

Good
```go
package main

import (
	"io"
	"log"
	"net"
	"os"
	"time"
)

func main() {
	l, err := net.Listen("tcp", "localhost:")
	if err != nil {
		log.Fatal(err)
	}

	go func() {
		for {
			conn, err := l.Accept()
			if err != nil {
				log.Fatal(err)
			}
			conn.Write([]byte("Hello there"))
			time.Sleep(2 * time.Second)
			conn.Close()
		}
	}()

	conn, err := net.Dial("tcp", l.Addr().String())
	if err != nil {
		log.Fatal(err)
	}
	b := make([]byte, 10)
	for {
		n, err := conn.Read(b)
		if err != nil {
			if err != io.EOF {
				log.Printf("Got unexpected error %v while reading", err)
			} else {
				os.Stdout.Write(b[:n])
			}
			break
		}
		os.Stdout.Write(b[:n])
	}
	conn.Close()
}
```

myleshorton:
Thanks @oxtoacart and sorry the order of the merges doesn't matter right?

oxtoacart:
@myleshorton Correct, the merge order doesn't matter because this branch of the lantern repo already includes the unmerged changes to idletiming and http-proxy.

myleshorton:
OK gotcha.

On Wed, Jun 8, 2016 at 10:48 AM, oxtoacart <notifications@github.com> wrote:

> @myleshorton <https://github.com/myleshorton> Correct, the merge order
> doesn't matter because this branch of the lantern repo already includes the
> unmerged changes to idletiming and http-proxy.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/getlantern/lantern/pull/4593#issuecomment-224672599>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABF0nsJZ6E1f4aBaE_WWJbWLsjsrE3z8ks5qJwBzgaJpZM4Iw7i7>
> .
>



-- 
--
President
https://www.getlantern.org

src/github.com/getlantern/flashlight/client/handler.go
myleshorton: Great catch @oxtoacart makes total sense.

oxtoacart: This falls into the category of "you break it, you fix it", as I'm the one who broke it :)

src/github.com/getlantern/flashlight/proxied/proxied.go
myleshorton: It seems like this might make sense to add whether or not the connection is specified as persistent @oxtoacart. I believe `DisableKeepAlives` essentially specifies whether or not to include the `Connection: close` header, but it's also always possible some server or client along the line doesn't work like it's supposed to.

oxtoacart: Sorry, I'm not sure I follow.  Whatever code is using this transport is free to set the Connection: Close header on new requests as it wishes, and that header will propagate through to the server.  If the server closes the connection, Transport is smart enough not to reuse it.  What's the piece that's missing?

src/github.com/getlantern/flashlight/client/reverseproxy.go
myleshorton: It strikes me that we should probably be using TLS session caching here too @oxtoacart no? Not related to this PR of course, but just snooping.

oxtoacart: Doesn't matter, because we never send HTTPS traffic through the reverse proxy.

src/github.com/getlantern/flashlight/proxied/proxied.go
myleshorton: I'm just saying that `persistent` flag to the `chained` method just specifies whether or not to set `DisableKeepAlives` in the `http.Transport`. The effectively governs whether or not the transport will add the `Connection: close` header. 

All I'm saying is I'm not sure the inclusion or lack of `Connection: close` (the `if persistent`) makes sense as the criteria for deciding whether or not to set a max idle time. It seems like we might just want to always set it.

src/github.com/getlantern/flashlight/client/reverseproxy.go
myleshorton: Oh right -- I keep swapping those in my head!

src/github.com/getlantern/flashlight/proxied/proxied.go
oxtoacart: Oh, I see what you're saying.  DisableKeepAlives doesn't just add a Connection: close header, it actually closes the connection itself even if the server didn't.  See line 513 and 524 of [transport.go](https://golang.org/src/net/http/transport.go).

myleshorton: Oh cool OK makes sense @oxtoacart! That actually answers some of my other questions about other places that use DisableKeepAlives -- all good.

oxtoacart: Consequently, if we're disabling Keep Alives, there's never any idle keep alive connections to remove.

oxtoacart: Oh, the other piece to understand is that MaxIdleTime here is not an idle timeout for our idletiming connection.  It's telling the Transport how long to hang on to pooled keepalive connections.

src/github.com/getlantern/idletiming/idletiming_conn.go
```diff
@@ -91,6 +100,13 @@ func (c *IdleTimingConn) TimesOutAt() time.Time {
 
 // Read implements the method from io.Reader
 func (c *IdleTimingConn) Read(b []byte) (int, error) {
+	c.closeMutex.RLock()
+	defer c.closeMutex.RUnlock()
```
myleshorton: Just commenting here, but do we really want to hold on to `closeMutex` for the entire read? That looks like it will make simultaneous reads and writes impossible. Maybe a comment on `closeMutex` specifying what exactly it's protecting (presumably `closed` but if that's the case I would probably put the lock in `checkClosed`)

src/github.com/getlantern/flashlight/proxied/proxied.go
myleshorton: Right good point @oxtoacart -- I had mixed those up.

src/github.com/getlantern/idletiming/idletiming_conn.go
oxtoacart: Everything except the close method is obtaining a R(ead)Lock.  Those are not exclusive.

myleshorton: Oh sorry @oxtoacart RLock allows an arbitrary number of readers -- looks ok.

myleshorton: It is very tempting to move the locking into checkClosed for a lot of those methods though @oxtoacart...

src/github.com/getlantern/idletiming/idletiming_conn.go
```diff
@@ -184,14 +216,27 @@ func (c *IdleTimingConn) Close() error {
...
 func (c *IdleTimingConn) LocalAddr() net.Addr {
+	c.closeMutex.RLock()
+	defer c.closeMutex.RUnlock()
+
 	return c.conn.LocalAddr()
 }
 
 func (c *IdleTimingConn) RemoteAddr() net.Addr {
+	c.closeMutex.RLock()
+	defer c.closeMutex.RUnlock()
```
myleshorton: Why do we need to lock here? We're not reading or writing from closed in this call right?

src/github.com/getlantern/idletiming/idletiming_listener.go
```diff
@@ -33,9 +32,13 @@ type idleTimingListener struct {
...
 	c, err = l.orig.Accept()
 	if err == nil {
-		c = Conn(c, l.idleTimeout, func() {
-			l.onIdle(c)
-		})
+		var onIdle func()
+		if l.onIdle != nil {
+			onIdle = func() {
+				l.onIdle(c)
+			}
```
myleshorton: I would probably add a safeguard here in case `l.onIdle == nil` @oxtoacart, as in:

```
else {
   onIdle = func() {}
}
```

Otherwise if anything inside `Conn` fails to check for nil it'll panic!

src/github.com/getlantern/idletiming/idletiming_conn.go
oxtoacart: The purpose of the lock is to prevent us from closing the underlying connection while an io op is in progress.  That requires us to hold the lock for the duration of the method.

src/github.com/getlantern/idletiming/idletiming_conn.go
oxtoacart: Might not need to if RemoteAddr() is unaffected by close calls, was just being conservative.

src/github.com/getlantern/idletiming/idletiming_conn.go
oxtoacart: The key thing is that the resource we're protecting is not the closed status, it's the underlying connection.

src/github.com/getlantern/idletiming/idletiming_listener.go
oxtoacart: No need, Conn already allows onIdle to be nil.  It's part of the contract for Conn.

myleshorton: idletiming_conn does check for nil in this case, but to me I generally prefer empty functions to relying on nil checks, as it'll still work if something changes.

src/github.com/getlantern/idletiming/idletiming_conn.go
myleshorton: That makes sense @oxtoacart. Can you add a comment to that effect to `closeMutex` for future readers of the code?

src/github.com/getlantern/idletiming/idletiming_conn.go
```diff
@@ -202,30 +247,50 @@ func (c *IdleTimingConn) SetDeadline(t time.Time) error {
...
 	}
+
 	atomic.StoreInt64(&c.writeDeadline, t.UnixNano())
 	return nil
 }
 
 func (c *IdleTimingConn) markActive(n int) bool {
 	if n > 0 {
-		c.activeCh <- true
+		select {
```
oxtoacart: @myleshorton I just noticed this, so snuck in a fix.  We don't want to risk markActive blocking if the run loop isn't reading from the activeCh because it's blocking on something else (like Close()).

src/github.com/getlantern/flashlight/client/handler.go
myleshorton: Really? I was fully ready to take the hit!

src/github.com/getlantern/idletiming/idletiming_conn.go
myleshorton: Yup that works @oxtoacart

myleshorton:
OK LGTM @oxtoacart!

PR #4613, opened 2016-06-10
Refactored protected in attempt to address errors on checking masquerades

oxtoacart:
So, I was never able to reliably reproduce the error, but I did notice a few things that could use improvement.  I've fixed them, and for me things still work.  @atavism would you mind trying this and seeing how it does for you?

cc: @myleshorton


src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
```diff
@@ -64,8 +64,8 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
 
             final List<FeedItem> items = new ArrayList<FeedItem>();
 
-            Lantern.FeedByName(name, new Lantern.FeedRetriever.Stub() {
-                public void AddFeed(String title, String desc, 
+            Lantern.FeedByName(name, new Lantern.FeedRetriever() {
```
oxtoacart: Fallout from upgrading to latest gomobile.

src/github.com/getlantern/netx/netx.go
```diff
@@ -0,0 +1,56 @@
+// Package netx provides additional libraries that extend some of the behaviors
+// in the net standard package.
+package netx
```
oxtoacart: New package that consolidates overridable dialers and resolvers.

src/github.com/getlantern/protected/protected.go
```diff
@@ -126,60 +130,89 @@ func Resolve(addr string) (*net.TCPAddr, error) {
...
 
 	// finally, convert the socket fd to a net.Conn
 	err = conn.convert()
 	if err != nil {
-		log.Errorf("Error converting protected connection: %v", err)
-		return nil, err
+		return nil, errors.New("Error converting protected connection: %v", err)
 	}
-
-	conn.Conn.SetDeadline(time.Now().Add(timeout))
```
oxtoacart: This logic seemed wrong.  SetDeadline sets an absolute deadline on all operations on this connection, whereas the timeout is just intended to govern the dial.  So, I changed this method to use withtimeout to enforce the dial timeout just on the dial phase.

src/github.com/getlantern/protected/protected.go
```diff
@@ -126,60 +130,89 @@ func Resolve(addr string) (*net.TCPAddr, error) {
...
 	}
 
+	_ip, timedOut, err := withtimeout.Do(remainingTimeout(), func() (interface{}, error) {
+		// do DNS query
+		ip := net.ParseIP(host)
+		if ip != nil {
+			return ip, nil
+		}
+		// Try to resolve it
+		addr, err := p.Resolve(network, addr)
```
oxtoacart: Since our dial function can be used in a bunch of places, I added the ability to resolve hostnames if dial was called with something other than an IP.

src/github.com/getlantern/tlsdialer/tlsdialer.go
```diff
@@ -51,21 +44,6 @@ type ConnWithTimings struct {
 	VerifiedChains [][]*x509.Certificate
 }
 
-// OverrideResolve allows overriding the DNS resolution function
-func OverrideResolve(override func(addr string) (*net.TCPAddr, error)) {
```
oxtoacart: This functionality has been superseded by netx.  Note how the old tlsdialer code didn't have a mutex on the global resolve and dialOverride fields.  This could theoretically have led to a race condition that showed the error @atavism  was seeing.  netx uses atomic.Value for these sorts of things, so that shouldn't be a problem.

src/github.com/getlantern/withtimeout/withtimeout.go
```diff
@@ -20,10 +22,10 @@ func (timeoutError) Error() string { return timeoutErrorString }
 func Do(timeout time.Duration, fn func() (interface{}, error)) (result interface{}, timedOut bool, err error) {
 	resultCh := make(chan *resultWithError, 1)
 
-	go func() {
+	ops.Go(func() {
```
oxtoacart: Helps our logging/error reporting.

src/golang.org/x/mobile/bind/gen.go
```diff
@@ -67,6 +67,8 @@ type generator struct {
 	interfaces []interfaceInfo
 	structs    []structInfo
 	otherNames []*types.TypeName
+	// allIntf contains interfaces from all bound packages.
```
oxtoacart: I upgraded to the latest gomobile.

src/github.com/getlantern/lantern/lantern.go
```diff
@@ -37,16 +37,15 @@ type SocketProtector interface {
 // because it keeps Lantern's own connections from being captured by the VPN and
 // resulting in an infinite loop.
 func ProtectConnections(dnsServer string, protector SocketProtector) {
-	protected.Configure(protector.Protect, dnsServer)
-	tlsdialer.OverrideResolve(protected.Resolve)
-	tlsdialer.OverrideDial(protected.Dial)
+	p := protected.New(protector.Protect, dnsServer)
```
oxtoacart: @myleshorton Because protected is only referenced from lantern now, and because lantern is only referenced from the mobile library, we no longer need to do platform detection to exclude protected from non-Android builds.

src/github.com/getlantern/protected/protected.go
```diff
@@ -35,29 +40,28 @@ type ProtectedConn struct {
...
+}
 
-func Configure(protect Protect, dnsServer string) {
-	mutex.Lock()
-	currentProtect = protect
-	if dnsServer != "" {
-		currentDnsServer = dnsServer
-	} else {
-		dnsServer = defaultDnsServer
+func New(protect Protect, dnsServer string) *Protector {
```
oxtoacart: protected itself is now stateless. It now returns an immutable Protector.

atavism:
Thanks, @oxtoacart! Lots of great improvements here. One issue I'm already seeing already though (even without turning full-device VPN mode on):

```
E  ERROR flashlight.client: handler.go:130   at github.com/getlantern/flashlight/client.pipeData (handler.go:130)
E  ERROR flashlight.client: handler.go:130   at github.com/getlantern/flashlight/client.(*Client).intercept (handler.go:110)
E  ERROR flashlight.client: handler.go:130   at github.com/getlantern/flashlight/client.(*Client).ServeHTTP (handler.go:34)
E  ERROR flashlight.client: handler.go:130   at net/http.serverHandler.ServeHTTP (server.go:2087)
E  ERROR flashlight.client: handler.go:130   at net/http.(*conn).serve (server.go:1474)
E  ERROR flashlight.client: handler.go:130   at runtime.goexit (asm_arm.s:990)
E  ERROR flashlight.client: handler.go:130 Caused by: EOF
```

Then when I actually turn full-device VPN mode on:

```
E  Message: ERROR lantern-android.protected: protected.go:170 Could not connect to socket: no route to host [error=Could not connect to socket: no route to host error_op= error_type=errors.Error]
E  ERROR lantern-android.protected: protected.go:170   at github.com/getlantern/protected.Dial (protected.go:170)
E  ERROR lantern-android.protected: protected.go:170   at github.com/getlantern/tlsdialer.DialForTimings (tlsdialer.go:157)
E  ERROR lantern-android.protected: protected.go:170   at github.com/getlantern/fronted.(*direct).dialServerWith (direct.go:258)
 E  ERROR lantern-android.protected: protected.go:170   at github.com/getlantern/fronted.(*direct).dialWith (direct.go:208)
E  ERROR lantern-android.protected: protected.go:170   at github.com/getlantern/fronted.(*direct).vetOne (direct.go:120)
E  ERROR lantern-android.protected: protected.go:170   at runtime.goexit (asm_arm.s:990)
E  ERROR lantern-android.protected: protected.go:170 Caused by: no route to host
```

So I'm still not sure what is going on. A lot of this started happening when we upgraded gomobile. I'm going to investigate downgrading to see if that helps.

oxtoacart:
@atavism The EOF thing is like related to a change that I recently made to idletiming.  We shouldn't be reporting this as an error as it's normal and even intended by the [HTTP spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html#sec8.1.4).

The "no route to host" error looks weird because those line numbers don't match what's on the `issue4554` branch.  Are you sure you've got the recent code?

src/github.com/getlantern/protected/protected.go
atavism: Nice! Do we maybe want to log the various errors that could happen here?

atavism:
@oxtoacart You're right. Looks like I entered some weird git state (most likely from switching between ``lantern`` and our private repo while working on the same branch). No longer seeing any of the ``no route to host`` errors with your changes!

src/github.com/getlantern/lantern-mobile/app/src/main/java/org/getlantern/lantern/fragment/FeedFragment.java
atavism: Thanks for fixing this! I noticed this also.

src/github.com/getlantern/protected/protected.go
oxtoacart: The errors propagate and I think they'll get logged at a higher level.  We could add some trace logging here if you think it'll be helpful.

oxtoacart:
@atavism Thanks for the review!

Issue #4638, opened 2016-06-11
where is the "MaxIdleTime"?


beautygirl001:
when i use "make windows" to make lantern for windows today, the terminal tell me the error "MaxIdleTime is not defined" ("EnforceMaxIdleTime").The dir is "lantern/src/github.com/getlantern/flashlight/client/reverseproxy.go"
I find nothing in the "/usr/local/go/src/net/http/transport.go" in go1.7 or go1.6.
How to solve this problem ?please help me !

wfjsw:
[Custom fork of Go](https://github.com/getlantern/go/tree/lantern) is currently required. You may build with the modified version of Go or use Docker ( `make docker-windows` ).

Issue #4605, opened 2016-06-09
ubuntu16.04无法运行lantern


yfszzx:
ubuntu16.04 32位，安装lantern-installer-beta-32-bit.deb没有报错，但运行lantern显示如下错误:
/usr/lib/lantern/lantern-binary: OK
DEBUG flashlight.ui: ui.go:50 Creating tarfs filesystem that prefers local resources at /lantern/src/github.com/getlantern/lantern-ui/app
DEBUG flashlight.ui: ui.go:50 Creating tarfs filesystem that prefers local resources at /lantern/src/github.com/getlantern/lantern-ui/app
Failed to connect to Mir: Failed to connect to server socket: No such file or directory
Unable to init server: Could not connect: Connection refused

(lantern:7977): Gtk-WARNING **: cannot open display:
求指点

wfjsw:
Your PC don't have a GUI screen? 
Compile Lantern from source by `HEADLESS make linux-386`. Also https://github.com/getlantern/lantern/wiki/Installation:-server

yfszzx:
I have shutted  down  GUI screen. I will try .Thank you

yfszzx:
I compiled Lantern from source by HEADLESS=true make linux-386,but I don't know what's wrong whick my PC :
** VERSION was not set, using default version. This is OK while in development.
Build tags: headless
Extra ldflags:
# github.com/getlantern/flashlight/client
src/github.com/getlantern/flashlight/client/reverseproxy.go:21: unknown http.Transport field 'MaxIdleTime' in struct literal
# github.com/getlantern/flashlight/proxied
src/github.com/getlantern/flashlight/proxied/proxied.go:391: tr.MaxIdleTime undefined (type *http.Transport has no field or method MaxIdleTime)
src/github.com/getlantern/flashlight/proxied/proxied.go:392: tr.EnforceMaxIdleTime undefined (type *http.Transport has no field or method EnforceMaxIdleTime)
Makefile:238: recipe for target 'linux-386' failed
make: *** [linux-386] Error 2

iamlook:
@yfszzx met the same problem as you too. Did you find the solution?

iamlook:
@yfszzx have to use lantern's go-lang fork: https://github.com/getlantern/go , net/http is different from master branch.

wfjsw:
That's dirty...
Time to fire up Docker.

wfjsw:
[...]

PR #4650, opened 2016-06-12
Using an explicit patch for Go

xiam:
If you're not using the Lantern's Go fork and you try to build it without docker it will fail:

* https://github.com/getlantern/lantern/issues/4638
* https://github.com/getlantern/lantern/issues/4605#issuecomment-224985945
* https://github.com/getlantern/lantern/issues/4692

For users there's no obvious indication of what happened. Some users could find out that our README changed a bit and that now we require a fork of Go but some others won't.

What I propose here is that `setenv` checks for the proper environment values and suggests what to do in case that the environment is not adequate (e.g.: when the user does not have our Go fork).

```
make linux
Lantern requires a special fork of Go, use 'make lantern-go'
make: *** [linux-386] Error 1
```

```
make lantern-go
# wait several minutes
```

```
make linux
# success!
```

The included patch was generated with https://github.com/getlantern/go/pull/2


xiam:
Needs more work, travis is failing.

xiam:
Travis is still failing but that's for a different reason than the code I added here, we have the same failure on devel.

Could you take a look @oxtoacart or @myleshorton?

myleshorton:
Yup I've been noticing this too @xiam -- I'll dig deeper into it.

oxtoacart:
Yeah, I'm seeing a failure too, lantern_test.go right?

oxtoacart:
It's weird, according to travis, this commit was good - 9ab47f4105517d8c9b0bbfdfb06b448b0b0f37d6

I tried checking that out locally and running, and the test still fails, so I suspect that this is something that changed on the proxy side.

xiam:
Thanks @oxtoacart, @myleshorton. PR is ready now.

xiam:
We don't need this anymore.

Kabibi:
@xiam What should I do next after downloading the patch? Or how to use?

xiam:
Hey @Kabibi, calling `make` would have done the trick, but this is not required anymore.

Kabibi:
I have tried "make lantern" and still have **"tr.MaxIdleTime undefined"** and **"tr.EnforceMaxIdleTime undefined"** errors. Now I have no idea how to make lantern successfully on my Opensuse. What do you mean by "this is not required anymore?" Can you give me more tips? Thank U so much! @xiam

Issue #4686, opened 2016-06-14
too many open files errors 


myleshorton:
This is happening again. Basically it seems to arise if I run Lantern for long enough. Here's the first log that appears:

```
Jun 14 17:00:55.205 - 107m12s DEBUG detour: detour.go:134 Dial initially to img.mediaplex.com:443 failed, try detour: dial tcp 64.156.167.78:443: socket: too many open files [detour=true op=proxied_dialer origin_host=img.mediaplex
.com origin_port=443 root_op=proxy user_agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.84 Safari/537.36]
Jun 14 17:00:55.205 - 107m12s DEBUG balancer: balancer.go:106 Dialing connect://img.mediaplex.com:443 with (trusted) chained proxy at 104.131.82.167:443 [] [detour=true op=proxied_dialer origin_host=img.mediaplex.com origin_port=443 root_op=proxy user_agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.84 Safari/537.36]
Jun 14 17:00:55.206 - 107m12s DEBUG balancer: dialer.go:159 Will recheck (trusted) chained proxy at 104.131.82.167:443 [] 8.062879289s later because it failed for 1 times [detour=true op=proxied_dialer origin_host=img.mediaplex.com origin_port=443 root_op=proxy user_agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.84 Safari/537.36]
Jun 14 17:00:55.206 - 107m12s DEBUG detour: detour.go:134 Dial initially to secure.img-cdn.mediaplex.com:443 failed, try detour: dial tcp 23.222.173.158:443: socket: too many open files [detour=true op=proxied_dialer origin_host=secure.img-cdn.mediaplex.com origin_port=443 root_op=proxy user_agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.84 Safari/537.36]
[...]

Issue #4705, opened 2016-06-15
Stopped dialer is still working


fffw:
Should be replaced by new ones. It should be the reason why https://github.com/getlantern/lantern/pull/4634 failed CI.

```
Jun 15 10:41:04.413 - 0m4s DEBUG flashlight.client: balancer.go:40 Adding chained server: <masked>:443
Jun 15 10:41:04.413 - 0m4s DEBUG flashlight: flashlight.go:112 Applied updated configuration
Jun 15 10:41:04.413 - 0m4s DEBUG flashlight.client: balancer.go:59 Closed old balancer
Jun 15 10:41:04.413 - 0m4s TRACE balancer: dialer.go:77 Dialer (trusted) chained proxy at 45.55.184.76:443 [] stopped
Jun 15 10:41:30.594 - 0m30s DEBUG flashlight.client: chained.go:175 Error testing dialer 45.55.184.76:443 to http://www.google.com:80/index.html: Unable to dial server (trusted) chained proxy at 45.55.184.76:443 []: dial tcp 45.55.184.76:443: i/o timeout
...
Jun 15 10:42:00.598 - 1m0s TRACE balancer: dialer.go:153 Dialer (trusted) chained proxy at 45.55.184.76:443 [] consecutive failures: 4 -> 5 [detour=false op=proxied_dialer root_op=proxied_dialer]
```

fffw:
The sequence is:
1. Dialing through old balancer with no accessible proxy server(say if the previously assigned proxies are retired).
2. New configuration is fetched.
3. Old balancer is closed and new one starts working.
4. Old balancer's `Dial` is still working, desperately dialing same proxy `dialAttempts` times, then fail.

We need to avoid 4. So we should never close balancer. Instead, reinitialize it with new set of dialers, and having `Dial` pick dialers currently dialers instead of cached copy.

myleshorton:
This is interesting @fffw. Do you think this is causing issues from the user's perspective?

fffw:
Yeah, I think so @myleshorton . The initial pages won't open even with new proxy servers swapped in, until user refreshes the page manually. Chrome mitigates the pain by refreshing automatically when user clicks a previously failed tab.

oxtoacart:
@fffw The reason CI is failing is because of this - https://github.com/getlantern/lantern_aws/issues/209

fffw:
Yes, but if client can use the newly assigned proxy to redial, CI can pass even with broken fallbacks

[...]

PR #4707, opened 2016-06-15
Using interruptible bidirectional copying to keep idle connections fr…

oxtoacart:
…om piling up, closes getlantern/lantern#4686


src/github.com/getlantern/netx/copy.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// BidiCopy copies between in and out in both directions, returning the errors
+// from copying to out and copying to in.
+func BidiCopy(out net.Conn, in net.Conn) (outErr error, inErr error) {
+	pool := getPool()
+	bufin := pool.Get()
+	bufout := pool.Get()
+	defer pool.Put(bufin)
+	defer pool.Put(bufout)
```
oxtoacart: Note - using a buffer pool is not necessary to fix this bug, but it's a useful optimization anyway.

src/github.com/getlantern/netx/copy.go
```diff
@@ -0,0 +1,72 @@
...
+	"time"
+)
+
+var (
+	copyTimeout = 1 * time.Second
+)
+
+// BidiCopy copies between in and out in both directions, returning the errors
+// from copying to out and copying to in.
+func BidiCopy(out net.Conn, in net.Conn) (outErr error, inErr error) {
```
oxtoacart: Once this PR is complete, I'll use this same function on the proxy, which basically does the same thing as the client [here](https://github.com/getlantern/http-proxy/blob/master/httpconnect/httpconnect.go#L129).

fffw:
https://github.com/getlantern/lantern/issues/4705 becomes the blocker of all the CI. I had some progress on that one.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -115,16 +121,11 @@ func (client *Client) intercept(resp http.ResponseWriter, req *http.Request, op
...
-	writeErrCh := make(chan error, 1)
-	// Start piping from client to proxy
-	op.Go(func() {
-		_, writeErr := io.Copy(connOut, clientConn)
-		writeErrCh <- writeErr
-	})
-
-	// Then start copying from proxy to client.
-	_, readErr := io.Copy(clientConn, connOut)
-	writeErr := <-writeErrCh
```
myleshorton: Where in the previous code are connections leaking @oxtoacart?

oxtoacart: There is no leak.  It's just that we don't close the connections until both the reading and the writing are done.  That meant that we were relying on the idle timeout to close the connections, which would stop the reader and writer.

myleshorton: Wait sorry but what's bad about not closing the connection until both reading and writing are done? Why would we want to close them before they're done or before the idle timeout @oxtoacart?

oxtoacart: Here's the scenario:

1. Client connects
2. Proxy connects to server
3. Proxy starts piping both ways
4. Client disconnects
5. Reading from the client will stop at this point due to an EOF, but writing to the client doesn't stop, so we keep both connections open

The simplistic way to do this is to just close both connections as soon as either io.Copy is finished, which is what we used to do, but that sometimes results in us closing a connection while we're still doing an io op on it.  The new code in netx sets timeouts on the io ops and interrupts in between ops.

myleshorton: Gotcha @oxtoacart. OK the code looks fine to me. Merging!

myleshorton:
Although it looks like one of the netx tests is failing @oxtoacart...

oxtoacart:
@myleshorton I fixed the test for Linux.

src/github.com/getlantern/netx/copy.go
```diff
@@ -0,0 +1,69 @@
...
+		if atomic.LoadUint32(stop) == 1 {
+			break
+		}
+		deadline := time.Now().Add(copyTimeout)
+		src.SetReadDeadline(deadline)
+		dst.SetWriteDeadline(deadline)
+		nr, er := src.Read(buf)
+		if nr > 0 {
+			nw, ew := dst.Write(buf[0:nr])
+			if ew != nil && !isTimeout(err) {
```
fffw: Shouldn't this be `!isTimeout(ew)`?

oxtoacart: No.  Timeouts are expected because we're intentionally setting a short deadline to give us a chance to unblock and check the stop flag.

src/github.com/getlantern/netx/copy.go
```diff
@@ -0,0 +1,69 @@
...
+				err = errors.New("Error writing: %v", ew)
```
oxtoacart: I just realized this is bad, as it masks the true error.  Removing.

fffw:
I was thinking of an probably simpler alternative: When piping on one direction returns, set the read and write dead line to a short period, so the pipe in opposite direction can be closed in a short time.

src/github.com/getlantern/netx/copy.go
fffw: I mean you are checking `err` which is always `nil` here. Should be `ew` instead.

fffw:
Oh my alternative has a flaw. It will close connections prematurely in below rarely happen case:
1. client sends request and close.
2. server continuously send large chunk of data back, then close.

Same flaw exists for current code.

src/github.com/getlantern/netx/copy.go
```diff
@@ -0,0 +1,67 @@
...
+// doCopy is based on io.copyBuffer
+func doCopy(dst net.Conn, src net.Conn, buf []byte, errCh chan error, stop *uint32) {
+	var err error
+	defer func() {
+		atomic.StoreUint32(stop, 1)
+		errCh <- err
+	}()
+
+	for {
+		if atomic.LoadUint32(stop) == 1 {
```
fffw: If we continue to loop here if no error, whether it's timeout or not, has occurred in previous iteration, the flaw I mentioned is fixed.

src/github.com/getlantern/netx/copy.go
oxtoacart: Ah, yes, great catch!

oxtoacart:
> Oh my alternative has a flaw. It will close connections prematurely in below rarely happen case:
>1. client sends request and close.
>2. server continuously send large chunk of data back, then close.

>Same flaw exists for current code.

Great catch!

myleshorton:
Hey @oxtoacart and @fffw I can't tell if all of the concerns are now addressed -- mine are.

src/github.com/getlantern/netx/copy.go
```diff
@@ -32,29 +35,43 @@ func doCopy(dst net.Conn, src net.Conn, buf []byte, errCh chan error, stop *uint
...
 		}
 		if er != nil && !isTimeout(er) {
 			err = er
-			break
+			return
+		}
+	}
+}
+
+func writeTo(dst net.Conn, buf []byte, writeTimeout time.Duration) error {
```
oxtoacart: @fffw I think this addresses your concern about short writes.  I modified the test to make sure we cover this.

Your idea of just setting the deadline after the other pipe has finished is interesting, but I'm not sure if it will work on all platforms.  Specifically, I don't know if calls to SetWriteDeadline() always affect existing Write() calls.  They do on Mac, but that seems like the sort of thing that might be platform specific.

Using an atomic closed indicator here seems to work.  Are you okay with that?

src/github.com/getlantern/netx/copy.go
```diff
@@ -0,0 +1,92 @@
...
+	go doCopy(out, in, bufIn, writeTimeout, outErrCh, &stop)
+	go doCopy(in, out, bufOut, writeTimeout, inErrCh, &stop)
+	return <-outErrCh, <-inErrCh
+}
+
+// doCopy is based on io.copyBuffer
+func doCopy(dst net.Conn, src net.Conn, buf []byte, writeTimeout time.Duration, errCh chan error, stop *uint32) {
+	var err error
+	defer func() {
+		atomic.StoreUint32(stop, 1)
```
oxtoacart: @fffw I tested and you were right, setting the deadline only once we're ready to close the connection performs better, and it seems to work on windows, os x and linux.  So, I did it!

fffw: I've read some article says that Go `net` package wraps OS level non-blocking IO into blocked ones, that explains why setting deadline can affect ongoing operations.

fffw:
Code looks great and works great. Awesome work @oxtoacart

oxtoacart:
@fffw Thanks for the thorough review and great suggestions!

PR #4712, opened 2016-06-16
Fixed baked-in fallbacks

oxtoacart:
@myleshorton @aranhoide @fffw This points at some new baked-in fallbacks.  I put them in Singapore because that works best for our China users.  I resized them to 8GB each just in case (hopefully that's not too much?).


oxtoacart:
Okay, this fixed the test, I'm just going to merge it.

PR #4802, opened 2016-06-25
Attempt to fix leaking conns

znqlsx:


myleshorton:
This has already been added in different form.

Issue #4803, opened 2016-06-26
Panics when logging error


fffw:
Happens occasionally. Binary is built from commit getlantern/lantern-pro@88e4c9694f4bc5924e8b63ac8937aa8eb3013a41, but should be something with lantern itself.

```
ERROR balancer: balancer.go:110 Caused by: dial tcp <masked>:443: connect: network is unreachable

panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x48 pc=0x428398c]

goroutine 57 [running]:
panic(0x45c2f40, 0xc82000a100)
	/Users/fffw/workspace/go-darwin-amd64-bootstrap/src/runtime/panic.go:481 +0x3e6
github.com/getlantern/errors.(*structured).MultiLinePrinter.func1(0xc827233f10, 0x4631f0a)
	/Users/fffw/workspace/getlantern/lantern-pro/src/github.com/getlantern/errors/errors.go:228 +0x37c
github.com/getlantern/golog.(*logger).print(0xc82011ca80, 0x5754728, 0xc82027a5c0, 0xc827233f10, 0x5, 0x46767c8, 0x5, 0x4631fa0, 0xc82a149f20)
	/Users/fffw/workspace/getlantern/lantern-pro/src/github.com/getlantern/golog/golog.go:197 +0x52c
github.com/getlantern/golog.(*logger).errorSkipFrames(0xc82011ca80, 0x4631fa0, 0xc82a149f20, 0x1, 0x0, 0x0)
	/Users/fffw/workspace/getlantern/lantern-pro/src/github.com/getlantern/golog/golog.go:262 +0x1c8
github.com/getlantern/golog.(*logger).Errorf(0xc82011ca80, 0x4716e20, 0x1f, 0xc827f62790, 0x1, 0x1, 0x0, 0x0)
	/Users/fffw/workspace/getlantern/lantern-pro/src/github.com/getlantern/golog/golog.go:267 +0xbc
github.com/getlantern/flashlight/config.(*fetcher).pollForConfig(0xc8202bc780, 0x65d6640, 0xc8298101b0, 0x47b0530, 0xda0951330, 0x0, 0x0)
	/Users/fffw/workspace/getlantern/lantern-pro/src/github.com/getlantern/flashlight/config/fetcher.go:121 +0x1ff
[...]

PR #4863, opened 2016-07-02
Simplify golog and errors package by calling PrintStack directly

fffw:
It intends to fix https://github.com/getlantern/lantern/issues/4803 indirectly by replacing `MultiLine` with a more direct one pass `PrintStack`.

There's one slight difference of `golog` output: The file/line number prepended to each line of stacktrace is removed. Actually, it's what I wanted to remove!

I do want to solve the interleaving of log lines in this PR.

### DIFF
--- a/src/github.com/getlantern/errors/errors.go
+++ b/src/github.com/getlantern/errors/errors.go
@@ -230,14 +230,14 @@ func (e *structured) PrintStack(w io.Writer) {
 	for {
 		for stackPosition := 0; stackPosition < len(err.callStack); stackPosition++ {
 			call := err.callStack[stackPosition]
-			fmt.Fprintf(w, "  at %+n (%s:%d)\n", call, call, call)
+			fmt.Fprintf(w, "    at %+n (%s:%d)\n", call, call, call)
 		}
 		cause, ok := err.cause.(*structured)
 		if !ok || cause == nil {
 			return
 		}
 		err = cause
-		fmt.Fprintf(w, "Caused by: %v\n", err)
+		fmt.Fprintf(w, "  Caused by: %v\n", err)
 	}
 }
 
--- a/src/github.com/getlantern/errors/errors.go
+++ b/src/github.com/getlantern/errors/errors.go
@@ -108,7 +108,7 @@ type Error interface {
 	// with the cause of this Error (if any), then the stacktrace when the
 	// cause is created, and so on.
 	// The output is an analogy of Java's stacktrace.
-	PrintStack(io.Writer)
+	PrintStack(io.Writer, string)
 
 	// Op attaches a hint of the operation triggers this Error. Many error types
 	// returned by net and os package have Op pre-filled.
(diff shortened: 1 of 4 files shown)


fffw:
The interleaving occurs between timestamp and error message, which is irrelevant to golog and errors package. It happens rarely, so I chose to not touch it atm.

The PR looks like a sledgehammer to crack a nut, but more straightforward code would be better to spot such bugs, and the log output looks less clutter now. @oxtoacart how do you think?

src/github.com/getlantern/errors/errors.go
```diff
@@ -207,53 +209,19 @@ func (e *structured) Error() string {
...
-			}
-			if err.cause == nil {
-				return false
-			}
-			err = err.cause.(*structured)
-			return true
-		}
-		if stackPosition < len(err.callStack) {
-			buf.WriteString("at ")
+	for {
```
fffw: This is the real change. `errors` has its own format to print stack, so it's not that beneficial to use a `MultiLine` interface.

fffw: I also added intention so the log looks a bit clearer

myleshorton:
I'm reassigning this one to you @oxtoacart since it's more in your wheelhouse if that's cool.

oxtoacart:
Honestly I kind of like @myleshorton's simple solution for the panic itself, and I personally like having the line file and line number on every line because that means I can grep for it.  In my experience with Log4J in Java land, stacktraces usually didn't include the line prefix and it always annoyed me!

fffw:
@oxtoacart does it sound ok if I add back the file and line number by passing a prefix string to `PrintStack`? The `MultiLinePrinter` implementation is still over-complicated to me.

src/github.com/getlantern/errors/errors_test.go
```diff
@@ -62,32 +62,27 @@ func TestNewWithCause(t *testing.T) {
...
-Caused by: orld
-Caused by: ld
-  at github.com/getlantern/errors.buildSubSubCause (errors_test.go:999)
-  at github.com/getlantern/errors.buildSubCause (errors_test.go:999)
-  at github.com/getlantern/errors.buildCause (errors_test.go:999)
-  at github.com/getlantern/errors.TestNewWithCause (errors_test.go:999)
-  at testing.tRunner (testing.go:999)
-  at runtime.goexit (asm_amd999.s:999)
-Caused by: d
+    at github.com/getlantern/errors.TestNewWithCause (errors_test.go:999)
```
oxtoacart: I prefer the shorter indentation as before.  Was the increase intentional?

oxtoacart:
@fffw If you add the prefix back and change the indentation to the original 2 spaces, I'm okay with changing the implementation.  I agree that this new style is cleaner.

fffw:
Cool @oxtoacart will do

fffw:
@oxtoacart I brought back the original indentation and line prefix on stack trace

oxtoacart:
LGTM!

Issue #4902, opened 2016-07-04
[unable to reproduce] Accessing PayPal failed with "a network change was detected" in Chrome


oxtoacart:
I haven't done much research into it yet, but I suspect this might be because I'm now getting multiple proxies and my IP is changing.  I looked into the log, and sure enough my first request to PayPal went through a different proxy than my 2nd.

Note - I was able to recover from this error by refreshing, but its still worth looking into.

oxtoacart:
Not super useful, but here's on thread to pull on - http://superuser.com/questions/747735/regularly-getting-err-network-changed-errors-in-chrome

myleshorton:
Interesting @oxtoacart thanks. So it seems like maybe once we hit X site with X proxy, have to stick with it for awhile or something. There's probably a simpler solution though.

oxtoacart:
I see this pretty reliably on www.nytimes.com.

```
Failed to load resource: net::ERR_NETWORK_CHANGED
```

oxtoacart:
It looks like ERR_NETWORK_CHANGED happens if Chrome detects that the IP address has changed.

https://cs.chromium.org/chromium/src/net/quic/quic_stream_factory.cc?q=ERR_NETWORK_CHANGED&sq=package:chromium&dr=C&l=1331

oxtoacart:
Maybe Chrome is using an internet server to determine its address, and we're proxying traffic to that internet server.

oxtoacart:
It looks like the implementation of this is platform specific, so it's possible that this is isolated to OS X.

p3rcya:
Some foreign trade users complained on this too

oxtoacart:
BTW, I'm seeing this all the time.  Can't imagine that it helps performance/UX.

fffw:
I didn't experience this issue at all because Lantern was always using the same proxy (the one with smallest `EMADialTime`). The `QualityFirst` strategy does work well if one proxy is quicker than others.

OTOH, the `Sticky` strategy keeps using same proxy unless it fails to dial for once. it should solve this issue but if one slow proxy is selected at first, Lantern will stick with that one.

A quick fix would be an improved `Stickey` strategy: When comparing two dialer, always choose the one with more `ConsecSuccesses` (mostly used), unless its `EMADialTime` is more than, say, twofold of another one. Actally that's the goal of `Weighted` strategy, I just lost myself when implementing it.

[...]

PR #5084, opened 2016-07-25
balancer don't retry same dialer if it was failed

### LINKED ISSUE #5075: balancer: don't retry same dialer
Currently, it retries 3 times, no matter if it dials via same dialer or different ones.

A blocked or inaccessible proxy always fail. With current 10 second fixed timeout, it requires 30s to time out, until then `proxied` can switch to fronted.

An unreliable proxy fail sometimes, which makes more trouble to the `proxied` package because it will switch to "chained only mode" even if it take 25s to make a call, while it only take 1s through fronted.

It's also not good from user's perspective. It's always better to get feedback if a page can't be open, rather than desperately waiting.

fffw:
resolve #5075

### DIFF
--- a/src/github.com/getlantern/balancer/balancer.go
+++ b/src/github.com/getlantern/balancer/balancer.go
@@ -102,11 +102,17 @@ func (b *Balancer) Dial(network, addr string) (net.Conn, error) {
 		trustedOnly = true
 	}
 
+	var lastDialer *dialer
 	for i := 0; i < dialAttempts; i++ {
 		d, pickErr := b.pickDialer(trustedOnly)
 		if pickErr != nil {
 			return nil, pickErr
 		}
+		if d == lastDialer {
+			log.Debugf("Skip dialing %s://%s with same dailer %s", network, addr, d.Label)
+			continue
+		}
+		lastDialer = d
 		log.Tracef("Dialing %s://%s with %s", network, addr, d.Label)
 		conn, err := d.dial(network, addr)
 		if err != nil {
--- a/src/github.com/getlantern/flashlight/client/client.go
+++ b/src/github.com/getlantern/flashlight/client/client.go
@@ -215,7 +215,9 @@ func (client *Client) proxiedDialer(orig func(network, addr string) (net.Conn, e
 			log.Tracef("Rewriting %v to %v", addr, rewritten)
 			return net.Dial(network, rewritten)
 		}
+		start := time.Now()
 		conn, err := proxied(network, addr)
+		log.Debugf("Dialing proxy takes %v for %s", time.Since(start), addr)
 		return conn, op.FailIf(err)
 	}
 }
--- a/src/github.com/getlantern/balancer/balancer_test.go
+++ b/src/github.com/getlantern/balancer/balancer_test.go
@@ -57,6 +57,27 @@ func TestSingleDialer(t *testing.T) {
 	}
 }
 
+func TestRetryOnBadDialer(t *testing.T) {
+	addr, l := echoServer()
+	defer func() { _ = l.Close() }()
+
+	d1Attempts := int32(0)
+	dialer1 := newCondDialer(1, func() bool { atomic.AddInt32(&d1Attempts, 1); return true })
+	d2Attempts := int32(0)
+	dialer2 := newCondDialer(2, func() bool { atomic.AddInt32(&d2Attempts, 1); return true })
+
+	b := New(Sticky, dialer1)
+	_, err := b.Dial("tcp", addr)
+	if assert.Error(t, err, "Dialing bad dialer should fail") {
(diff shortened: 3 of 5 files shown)


src/github.com/getlantern/balancer/balancer.go
```diff
@@ -102,11 +102,17 @@ func (b *Balancer) Dial(network, addr string) (net.Conn, error) {
 		trustedOnly = true
 	}
 
+	var lastDialer *dialer
 	for i := 0; i < dialAttempts; i++ {
 		d, pickErr := b.pickDialer(trustedOnly)
 		if pickErr != nil {
 			return nil, pickErr
 		}
+		if d == lastDialer {
```
fffw: It works only if the client has a single dialer. It's sufficient for current free clients

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -215,7 +215,9 @@ func (client *Client) proxiedDialer(orig func(network, addr string) (net.Conn, e
 			log.Tracef("Rewriting %v to %v", addr, rewritten)
 			return net.Dial(network, rewritten)
 		}
+		start := time.Now()
 		conn, err := proxied(network, addr)
+		log.Debugf("Dialing proxy takes %v for %s", time.Since(start), addr)
```
fffw: Not the initial intention for this PR, but logging dial time tells us how the client behaves from user submitted logs.

fffw:
It's a simple change, but should improve UI responsiveness a lot if the free proxy is inaccessible.

@myleshorton would you mind to take a look?

myleshorton:
I'm on it @fffw. It seems like it's an effective change, but I'm just wondering a little bit if it's the right place for it because it theory the 3 tries to me are there to make absolutely sure that X proxy is inaccessible. We could do something like switch to another proxy as soon as we see a single failure, but that seems like a slightly different problem. I'll look at it harder though.

src/github.com/getlantern/balancer/balancer.go
myleshorton: Nice. I misread this the first time. LGTM @fffw.

myleshorton: The only thing is that the dialer picking is weighted right? So it's quite likely it will pick the same dialer over and over again, likely using up all `dialAttempts` quite quickly. Is that correct? 

It's almost like as soon as we see a single failure we want to switch modes and dial several simultaneously or something like that.

myleshorton: Like whenever there's a failure, we kind of reset and dial all simultaneously. Whichever connects first is the new default dialer.

I know that's a totally different approach, but TBH I'm not a huge fan of the weighted balancer stuff because it means we'll always sometimes try to connect to a potentially really slow proxy. If that's a key CSS file or something, it will completely slow down the page loading (could actually be a reason why users say pro is slow).

I basically don't think the concept of a "balancer" is really what we want.

myleshorton: That said, I think this is clearly an improvement, and I'll go ahead and merge it. We can open other tickets if necessary, but what do you think about the above assessment? Also curious about your thoughts on this @oxtoacart. I'm basically worried about users in China, for example, getting unlucky when the balancer chooses to load some key css file from Amsterdam. That would definitely result in the page appearing slow, and much slower than in free!

myleshorton:
LGTM, although I think a larger change might be worth considering as mentioned in the comments @fffw and @oxtoacart.

src/github.com/getlantern/balancer/balancer.go
fffw: If one dialer fails or becomes slower than others, the weight will be changed so other dialers take precedence. Yeah, it requires 10s (dial timeout) to notice if a proxy stops response. With the added dial time logging, we can have a better understanding how it behaves from user logs.

Parallelly dialing multiple proxies does work better, but it wastes resourse on both proxy and client, and increases the total file descriptors. A better apprach would be:

1. keep persistent connection to each proxy (disconnect it randomly to be less noticable).
1. keep calculating the throughput and RTT.
1. when new requests come in, forward the request via the fastest proxy, write to a LRU association table.
1. always forward follow-up requests of that domain to same proxy, unless the metrics of that proxy becomes unacceptable.

It requires significant change on both client and proxy, but should be worth doing if the rationale is solid.

fffw:
Thanks @myleshorton ! It's definitely a critical part to improve.

src/github.com/getlantern/balancer/balancer.go
oxtoacart: > 1. keep persistent connection to each proxy (disconnect it randomly to be less noticable).

If we can avoid getting fingerprinted here, this is also a potentially nice performance boost because it eliminates unnecessary TCP connection setup.

> always forward follow-up requests of that domain to same proxy, unless the metrics of that proxy becomes unacceptable.

I think this would solve #4902 as well.

PR #5090, opened 2016-07-26
Added loggly-sample-percentage command-line flag which defaults to sa…

oxtoacart:
…mpling 0.5% of device ids


src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -410,3 +401,27 @@ func enableBorda(bordaReportInterval time.Duration, bordaSamplePercentage float6
...
 }
+
+func includeInSample(deviceID string, samplePercentage float64) bool {
+	if samplePercentage == 0 {
+		return false
+	}
+
+	// Sample a subset of device IDs.
+	// DeviceID is expected to be a Base64 encoded 48-bit (6 byte) MAC address
+	deviceIDBytes, base64Err := base64.StdEncoding.DecodeString(deviceID)
```
myleshorton: @oxtoacart it strikes me that at least for sampling purposes we don't care about the correlation of this device ID with the device ID elsewhere in the app -- i.e. we could just use `uuid.NodeID()` directly here and skip decoding. The only reason I mention it is things might get weird if for some reason the encoding changed without realizing there's an assumption about the encoding here.

Additionally, though, I do wonder about storing the device ID from a data retention perspective. Are we really using it at all? If not, I'd be tempted to also remove that from the data we store.

myleshorton: Would be pretty easy just to write a quick unit test for this too @oxtoacart

myleshorton: Actually on the unit testing note @oxtoacart I think `logging.go` and `errors.go` are the #1 and #2 files in terms of lines *not covered* by tests. Looking at https://coveralls.io/builds/7207505/source?filename=src%2Fgithub.com%2Fgetlantern%2Fflashlight%2Flogging%2Flogging.go, it seems like things like `EnableFileLogging` and `initContext` would be pretty quick wins in terms of just adding potentially pretty trivial tests.

I know the tls proxy stuff and also just getting this in today (actually I think getting a version out today is probably even more important than tlsproxy ATM) are super high priority, but just noticed that while you're in the area we might be able to take another bite out of the untested lines in coverage.

oxtoacart: I'll add some unit test coverage in #5099 since I had to touch the tests there anyway.

Regarding DeviceID - I use it (as encoded) because it's available on both the client and server and allows us to correlate.  For borda, this affects sampling because we want sampling on the client and server to be consistent so that for those device IDs included in the sample, we have end-to-end coverage.  I like keeping it consistent with Loggly to because this ensures that we have full coverage across Loggly and Borda for whichever device IDs are in the sample.

src/github.com/getlantern/flashlight/logging/logging_test.go
```diff
@@ -112,3 +114,15 @@ func TestLoggly(t *testing.T) {
 		assert.Equal(t, 100, len(result["message"].(string)))
 	}
 }
+
+func TestIncludeInSample(t *testing.T) {
```
oxtoacart: @myleshorton I added a unit test for includeInSample here.  Turns out it found a bug :)  I needed to use LittleEndian instead of BigEndian.

I'll add additional unit tests in the other PR.

src/github.com/getlantern/flashlight/logging/logging.go
myleshorton: OK cool -- that's fine with me.

src/github.com/getlantern/flashlight/logging/logging_test.go
myleshorton: Nice catch @oxtoacart!

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -420,7 +420,6 @@ func includeInSample(deviceID string, samplePercentage float64) bool {
 	}
 	// Pad and decode to int
 	paddedDeviceIDBytes := append(deviceIDBytes, 0, 0)
-	// Use LittleEndian because Mac address has most significant bytes on left
-	deviceIDInt := binary.LittleEndian.Uint64(paddedDeviceIDBytes)
+	deviceIDInt := binary.BigEndian.Uint64(paddedDeviceIDBytes)
```
oxtoacart: @myleshorton I found that in practice, my "fix" actually resulted in my own PC always getting sample no matter how small of a sample percentage I set.  The old "broken" algorithm didn't give us quite the right sampling percentage, but we know it's working in production because we get data for only a subset of clients, so I'm going back to that.  I added a TODO to clean this up later, but it's not entirely clear to me right now how.

src/github.com/getlantern/flashlight/logging/logging_test.go
oxtoacart: See my more recent comment.  This "catch" actually broke things in practice when using real MAC addresses.  I went back to the old algorithm because we know that it actually does only sample a subset of clients, even if it's not exactly the right percentage.

myleshorton: OK sounds good.

myleshorton:
Merging

PR #5097, opened 2016-07-28
Move all generated go files to separate folder

fffw:
It's a bit inconvenient to me to check which is source file and which is generated in `config` package, so I made this change and a few others along the way.


src/github.com/getlantern/flashlight/genconfig/genconfig.go
```diff
@@ -111,15 +111,15 @@ func main() {
 		}
 	}
 	if *proxiedSitesOutFile != "" {
-		proxiedSitesTmpl := loadTemplate(*proxiedSitesOutFile)
+		proxiedSitesTmpl := loadTemplate("proxiedsites.go.tmpl")
```
fffw: It's quite interesting to me how `genconfig.go` could run correctly without using the correct template. However, the generated `proxiedsites.go` and `fallbacks.go` is used nowhere.

src/github.com/getlantern/flashlight/genconfig/genproxies.bash
```diff
@@ -5,22 +5,35 @@ function die() {
   exit 1
 }
 
+if [ -z "$LANTERN_AWS_PATH" ]; then
```
fffw: In case  `lantern_aws` is not in the default path

src/github.com/getlantern/flashlight/genconfig/masquerades.go.tmpl
```diff
@@ -1,16 +1,16 @@
-package fronted
```
fffw: Again, the generated `masquerades.go` is used nowhere

fffw:
@myleshorton I made this simple change to move all generated go files to a different directory. I successfully run `genconfig.bash` and `make darwin`, the generated binary runs correctly.

What puzzled me is that 3 of the generated go file is used nowhere, and previous `genconfig.go` seems not working.

Feel free to push more commits directly if you think this PR is kinda useful!

myleshorton:
cool thanks @fffw -- I'll have a look!

myleshorton:
@fffw I'm holding off on merging this and my own PR because I think a round of QA would be useful for this one, so I don't want to just push it out.

fffw:
Sure @myleshorton. Definitely true for such type of changes

myleshorton:
LGTM @fffw!

src/github.com/getlantern/flashlight/genconfig/genconfig.go
myleshorton: Actually @aranhoide this one is interesting -- I wonder if that could be causing any issues on the server side generation? I don't think any go files are used there, but maybe worth a look.

aranhoide: @myleshorton : definitely no go files are used there.

PR #5396, opened 2016-11-04
Cfscanner

myleshorton:
Hey @aranhoide you want to take a quick look at this?


aranhoide:
LGTM; IIUC this only removes stuff that became obsolete when we repurposed the Cloudflare IP range scanner for CloudFront.  The fact that this compiles at all means nothing necessary was removed.  There's one failing test, but it seems in an unrelated area.

PR #8408, opened 2026-01-08
Improve Android CI build time

atavism:
Resolves https://github.com/getlantern/engineering/issues/2920

Tested here: https://github.com/getlantern/lantern/actions/runs/20823647563


jay-418 (approved):
Looks good to me!

I also just learned that [baseline costs](https://docs.github.com/en/billing/concepts/product-billing/github-actions#baseline-minute-costs) for macOS are +6x more expensive than Windows, which is itself 2-5x more expensive than Linux.

And so 90% of our costs so far are just the macOS runner. Hopefully that 15 > 26 version boost has some positive effects. 🤞 

Image

PR #8411, opened 2026-01-12
Fix Windows desktop checkout

atavism:
support WebView popups + return checkout redirect URL


jigar-f (approved):
Changes are looking great, just test it with auth and private server flow once.

PR #8414, opened 2026-01-13
Add app lifecycle hook and improve QR scanner

atavism:


lib/core/hooks/use_app_lifecycle_listener.dart
jigar-f, line 1: @atavism Thanks for making these chanegs, But I think hook already has built-in life App Lifecycle lisnter.

lib/features/qr_scanner/qr_code_scanner.dart
```diff
@@ -1,51 +1,98 @@
...
+    );
+
+    final isHandling = useRef(false);
+
+    useEffect(() {
+      return controller.dispose;
+    }, [controller]);
+
+    // Lifecycle: stop camera in background, restart on resume
+    useAppLifecycleListener((state) {
```
jigar-f: Maybe we can migrate to using bult in, Let me kow if that sounds good

```dart
useOnAppLifecycleStateChange((previous, current) {
      final state = current;
      final isCurrentRoute = (ModalRoute.of(context)?.isCurrent ?? true);

      if (state == AppLifecycleState.paused ||
          state == AppLifecycleState.inactive) {
        controller.stop();
        return;
      }

      if (state == AppLifecycleState.resumed && isCurrentRoute) {
        if (!isHandling.value) controller.start();
      }
    },);
```

atavism: Oh of course, I didn't know this already existed. Thanks!

atavism: Updated

PR #8430, opened 2026-01-22
Linux service integration

atavism:
For https://github.com/getlantern/engineering/issues/2836

- Adds Makefile targets to build/stage lanternsvc + systemd unit into the packaging root
- Adds Linux lanternsvc daemon entrypoint and wires FFI VPN controls to it over IPC


lantern-core/ffi/ffi_linux.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	mode, err := ipc.GetClashMode(ctx)
+	if err != nil || mode == "" {
+		mode = "global"
+	}
+
+	if err := ipc.SelectOutbound(ctx, mode, tag); err != nil {
+		return SendError(fmt.Errorf("select outbound failed: %w", err))
+	}
```
garmr-ulfr: While you can use the `ipc` functions directly, for connecting/selecting servers (_e.g. `connectToServer`) at least, I would recommend using `vpn.ConnectToServer`/`vpn.QuickConnect`. They ensure the correct clash mode is used — of which only `auto`, `lantern`, and `user` are valid.

atavism: Thanks @garmr-ulfr I will fix that

garmr-ulfr:
@atavism heads-up, I am in the middle of making some changes to for authenticating the client connected to the socket and respect the provided socket path if it ends with the file name. ATM radiance will tack on `radiance.sock` to whatever is provided, assuming it's a directory.

atavism:
> With the exception of using a daemon, lantern will work on linux using the same VPN calls as other platforms

@garmr-ulfr So we don't need the FFI/IPC client changes here then? Can you remove those and please test the packaging updates?

PR #8439, opened 2026-02-02
Fix auto-updater feed URL

atavism:
Resolves https://github.com/getlantern/engineering/issues/2966

- Introduces Updater to centralize auto-updater setup
- Set feed URL based on build type (beta vs production)


lib/core/updater/updater.dart
```diff
@@ -0,0 +1,55 @@
+import 'dart:async';
+import 'dart:io';
+import 'package:auto_updater/auto_updater.dart';
+import 'package:flutter/foundation.dart';
+import 'package:lantern/core/common/app_build_info.dart';
+import 'package:lantern/core/common/common.dart';
+import 'package:lantern/core/models/feature_flags.dart';
+
+class Updater {
```
jigar-f, line 10: Can we inject this using an depandancy injection so we use it like 
`sl<Updater>();`

lib/features/setting/setting.dart
```diff
@@ -303,13 +306,14 @@ class _SettingState extends ConsumerState<Setting> {
 
   Future<void> checkForUpdates() async {
     try {
-      autoUpdater.checkForUpdates();
-    } catch (e) {
-      appLogger.error('Error checking for updates: $e');
+      await Updater.checkNow();
```
jigar-f: We could do something like 
`sl<Updater>();`

lib/main.dart
```diff
@@ -43,13 +41,14 @@ Future<void> main() async {
...
-
+    appLogger.debug('Loading translations...');
+    await Future.microtask(Localization.loadTranslations);
   } catch (e, st) {
     appLogger.error("Error during app initialization", e, st);
   }
   final flags = await _loadFeatureFlags();
   final sentryEnabled = flags.getBool(FeatureFlag.sentry) && kReleaseMode;
-  await _configureAutoUpdate(flags: flags);
+  await Updater.init(flags: flags);
```
jigar-f: same here

lib/core/updater/updater.dart
atavism: Ah yeah that sounds great

PR #8443, opened 2026-02-04
updated Windows code-signing using Signpath

jay-418:
## background

After being gaslit by Microsoft a couple of times, we switched to the more expensive Signpath for signing, and their subcontracted GlobalSign EV certificate.
- https://github.com/getlantern/engineering/issues/2940

## important info

- Signpath account is in 1password, username `admin@lantern.io`, approved identity: @jay-418 
- `build-windows.yml` has an input, `use-self-signed-cert`, which can switch between the self-signed and the signed EV certificate which required an extensively verified identity

## deliverables

- [x] setup CI for [test project](https://app.signpath.io/Web/192cf31f-4d27-425a-b334-412f9267f939/SigningRequests)
- [x] submit [application](https://docs.google.com/spreadsheets/d/1yFkkgdTHOTpH3wNrRaaxn40MWLKFvhwpWZQmoY6_q74/edit?usp=sharing) to https://signpath.org/apply
- [-] ~meet terms terms https://signpath.org/terms~ (not for an org of our maturity)
- [x] complete EV certificate and attach to `prod-policy`, verify it works before merge

### DIFF
--- a/.github/workflows/transifex-push.yml
+++ b/.github/workflows/transifex-push.yml
@@ -3,8 +3,6 @@ name: "Push to transifex"
 on:
   push:
     branches: [main]
-
-  # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
 jobs:
@@ -14,6 +12,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+
+      - name: Validate PO file
+        run: |
+          sudo apt-get install -y gettext
+          msgfmt --check assets/locales/en.po
 
       - name: "Push main translations file"
         uses: transifex/cli-action@v2
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -60,6 +60,25 @@
                 <category android:name="android.intent.category.BROWSABLE" />
                 <data android:scheme="https" android:host="lantern.io" android:pathPrefix="/auth" />
             </intent-filter>
+            <!-- Custom scheme fallback for browsers that don't support Android App Links (e.g. Firefox) -->
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="lantern" android:host="auth" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="lantern" android:host="report-issue" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="lantern" android:host="private-server" />
+            </intent-filter>
             <intent-filter android:autoVerify="true">
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
--- a/assets/locales/en.po
+++ b/assets/locales/en.po
@@ -1,3 +1,8 @@
+msgid ""
+msgstr ""
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+
 msgid "app_name"
 msgstr "Lantern"
 
@@ -1184,10 +1189,10 @@ msgid "lantern_pro_license"
 msgstr "Lantern Pro License"
 
 msgid "lantern_pro_license_required"
-msgstr “Lantern Pro license is required”
+msgstr "Lantern Pro license is required"
 
 msgid "lantern_pro_license_invalid"
-msgstr "Invalid Lantern Pro license”
+msgstr "Invalid Lantern Pro license"
 
 msgid "lantern_pro_license_invalid_length"
 msgstr "Lantern Pro license must be 25 characters"
@@ -1359,10 +1364,6 @@ msgstr "Data Usage Alerts"
 
 msgid "mb_free_data_remaining"
 msgstr "%d MB free data remaining"
-
-msgid "daily_data_cap_reached"
-msgstr "Daily data cap reached"
-
 
 msgid "daily_data_cap_reached_notification_message"
 msgstr "Your data will reset at %s. Upgrade to Pro for unlimited high-speed data."
--- a/ios/Runner/Info.plist
+++ b/ios/Runner/Info.plist
@@ -60,5 +60,18 @@
 	<string>We use notifications to keep you informed.</string>
 	<key>FlutterDeepLinkingEnabled</key>
 	<false/>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLName</key>
+			<string>org.getlantern.lantern</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>lantern</string>
+			</array>
+		</dict>
+	</array>
 </dict>
 </plist>
--- a/lib/features/account/account.dart
+++ b/lib/features/account/account.dart
@@ -61,10 +61,10 @@ class Account extends HookConsumerWidget {
           if (isExpired) ...{
             InfoRow(
               minTileHeight: 40,
-              backgroundColor: AppColors.red1,
-              borderColor: AppColors.red2,
+              backgroundColor: buildContext.statusErrorBg,
+              borderColor: buildContext.statusErrorBorder,
               textStyle: theme.labelLarge!.copyWith(
-                color: buildContext.statusErrorBg,
+                color: buildContext.statusErrorText,
               ),
               text: 'pro_subscription_expired_message'.i18n,
             ),
--- a/lib/lantern_app.dart
+++ b/lib/lantern_app.dart
@@ -101,7 +101,8 @@ class _LanternAppState extends ConsumerState<LanternApp>
     // Normalize: custom scheme lantern://open/path → treat as /path
     final path = uri.path;
 
-    if (path.startsWith('/report-issue')) {
+    if (path.startsWith('/report-issue') ||
+        (uri.scheme == 'lantern' && uri.host == 'report-issue')) {
       final pathUrl = uri.toString();
       final queryParams = uri.queryParameters;
       final segment = pathUrl.split('#');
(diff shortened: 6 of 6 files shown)


windows/packaging/exe/inno_setup.iss
```diff
@@ -285,8 +285,6 @@ Name: "{#ProgramDataDir}"; Permissions: users-modify
 
 [Files]
 Source: "{{SOURCE_DIR}}\\*"; DestDir: "{app}"; Flags: ignoreversion recursesubdirs createallsubdirs
-Source: "{{SOURCE_DIR}}\\wintun.dll"; DestDir: "{app}"; Flags: ignoreversion
-Source: "{{SOURCE_DIR}}\\lanternsvc.exe"; DestDir: "{app}"; Flags: ignoreversion
```
jay-418, line 289: These seem to be redundant with the recursive copy in the prior line, and also caused a build failure.

.github/workflows/build-windows.yml
```diff
@@ -106,35 +109,158 @@ jobs:
...
+        id: sign
+        shell: pwsh
+        run: |
+          # Use test signing policy if self-signed cert is enabled, otherwise use production policy
+          $signingPolicy = if ("${{ inputs.use-self-signed-cert }}" -eq "true") {
+            "${{ vars.SIGNPATH_SIGNING_POLICY_SLUG_TEST }}"
+          } else {
+            "${{ vars.SIGNPATH_SIGNING_POLICY_SLUG }}"
+          }
+
```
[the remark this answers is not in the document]
jay-418: refactored and extracted signing into a powersheel script so we can sign both individual binaries and also the whole package.

scripts/ci/sign-windows.ps1
```diff
@@ -0,0 +1,158 @@
...
+        "Description" = if ($Description) { $Description } else { "Signing $fileName" }
+    }
+
+if ($response.StatusCode -ne 201) {
+    Write-Error "Failed to submit signing request: HTTP $($response.StatusCode)"
+    Write-Host "Response body: $($response.Content)"
+    exit 1
+}
+
+$signRequestUrl = $response.Headers.Location[0]
```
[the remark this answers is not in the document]
jay-418: I think this is wrong, and it's pulling the first and only item from the list.

scripts/ci/sign-windows.ps1
```diff
@@ -0,0 +1,176 @@
+<#
+.SYNOPSIS
+    Signs a Windows binary using SignPath.
+
+.DESCRIPTION
+    Submits a file to SignPath for code signing, waits for completion,
+    and downloads the signed artifact back to the original path.
+
+.PARAMETER FilePath
+    Path to the file to sign. The signed file will overwrite this path.
```
Crosse, line 10: For no reason whatsoever—because I doubt you decide to dive head-first into embracing PowerShell as your default—I'll note that another way to document parameters by adding a comment on the parameter itself, like [this](https://github.com/Crosse/Crosse.PowerShell.Media/blob/63e8408bb3e8d5c117d99e2548c64f6c791ec509/Out-M4V.psm1#L20-L43). I tend to prefer that idiom instead for obvious locality reasons, but either way is valid.

scripts/ci/sign-windows.ps1
```diff
@@ -0,0 +1,176 @@
...
+    ./sign-windows.ps1 -FilePath "build/lantern.exe" -SigningPolicy "release-signing" `
+        -OrganizationId $env:SIGNPATH_ORG_ID -ProjectSlug "lantern" -ApiToken $env:SIGNPATH_API_TOKEN
+#>
+
+param(
+    [Parameter(Mandatory = $true)]
+    [string]$FilePath,
+
+    [Parameter(Mandatory = $true)]
+    [string]$SigningPolicy,
```
Crosse, line 47: Reading the parameter description led me to believe that there were a defined set of allowed values for `-SigningPolicy`, but I don't see any other validation in the code. For what it's worth: If there _are_ only a handful of valid values, you can use `[ValidateSet()]` here, [like so](https://github.com/Crosse/Crosse.PowerShell.Media/blob/63e8408bb3e8d5c117d99e2548c64f6c791ec509/Out-M4V.psm1#L51-L55).

scripts/ci/sign-windows.ps1
```diff
@@ -0,0 +1,176 @@
...
+    [Parameter(Mandatory = $false)]
+    [int]$PollIntervalSeconds = 10
+)
+
+$ErrorActionPreference = "Stop"
+
+# Validate file exists
+if (-not (Test-Path $FilePath)) {
+    Write-Error "File not found: $FilePath"
+    exit 1
```
Crosse: Since you set `$ErrorActionPreference` to `Stop` above, the `exit 1` is superfluous; the script will exit right after `Write-Error`.

```pwsh
# [PS] > gc ./test.ps1
$ErrorActionPreference = "Continue"
Write-Error "This error won't stop the script"

$ErrorActionPreference = "Stop"
Write-Error "The script stops here"

Write-Error "You should never see this"

# [PS] > ./test.ps1
Write-Error: This error won't stop the script
Write-Error: The script stops here
# [PS] >
```

scripts/ci/sign-windows.ps1
```diff
@@ -0,0 +1,176 @@
...
+        "ProjectSlug" = $ProjectSlug
+        "SigningPolicySlug" = $SigningPolicy
+        "Artifact" = Get-Item $FilePath
+        "Description" = if ($Description) { $Description } else { "Signing $fileName" }
+    }
+
+if ($response.StatusCode -ne 201) {
+    Write-Error "Failed to submit signing request: HTTP $($response.StatusCode)"
+    Write-Host "Response body: $($response.Content)"
+    exit 1
```
Crosse: Lines 94 and 95 will never run since `$ErrorActionPreference` is set to `Stop`.

scripts/ci/sign-windows.ps1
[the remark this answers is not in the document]
Crosse: It's not quite _wrong_, because I doubt Copilot researched enough to know that the `Header` property is always a map of string keys to arrays. You're doing it right. There _are_ ways to turn `Object`s into arrays of `Object`s if you're ever in a situation where it's ambiguous:

```powershell
$arr = @("I'm", "an", "array")
$str = "I'm a string"

$arr.Length
# ==> 3

@($arr).Length
# ==> 3

$str.Length
# ==> 12
@($str).Length
# ==> 1
```

Doing that can also shut some linters up, too…

scripts/ci/sign-windows.ps1
```diff
@@ -0,0 +1,176 @@
...
+Write-Host "Waiting for signing to complete..."
+
+$attempt = 0
+while ($attempt -lt $MaxAttempts) {
+    Start-Sleep -Seconds $PollIntervalSeconds
+    $attempt++
+
+    $status = Invoke-RestMethod -Method GET `
+        -Uri $signRequestUrl `
+        -Headers @{ "Authorization" = "Bearer $ApiToken" }
```
Crosse: Since `$ErrorActionPreference = "Stop"`, any HTTP status code >= 400 will cause the script to exit unless you add the `-SkipHttpErrorCheck` param here.

scripts/ci/sign-windows.ps1
```diff
@@ -0,0 +1,176 @@
...
+
+    Write-Host "Status: $($status.Status) (attempt $attempt/$MaxAttempts)"
+
+    if ($status.Status -eq "Completed") {
```
Crosse: I would very seriously consider inverting this logic to reduce the indentation level:

```powershell
# line 105:
:certStatusCheck while ($attempt -lt $MaxAttempts) {
    [...]
    # line 115:
    switch ($status.Status) {
        { $_ -in @("Failed", "Denied") } {
            Write-Error "Signing failed with status: $($status.Status)"
            # Write-Error causes script to stop execution
        }
        "Completed" {
            break certStatusCheck
        }
        { $attempt -ge $MaxAttempts } {
            Write-Error "Timeout waiting for signing to complete after $MaxAttempts attempts"
        }
    }
}

# The 'break' in the switch continues execution here
# [lines 115–168 go here, but with less indentation…]
```

(Or something like that, at least)

scripts/ci/sign-windows.ps1
```diff
@@ -0,0 +1,176 @@
...
+        Write-Host "Signing completed successfully!"
+
+        # Download signed artifact to a temp file first
+        $tempFile = "$FilePath.signed"
```
Crosse: nit: uses braces around variable names if concatenating them with extra text.

```suggestion
        $tempFile = "${FilePath}.signed"
```

scripts/ci/sign-windows.ps1
```diff
@@ -0,0 +1,176 @@
...
+        Invoke-WebRequest -Method GET `
+            -Uri "$signRequestUrl/SignedArtifact" `
+            -Headers @{ "Authorization" = "Bearer $ApiToken" } `
+            -OutFile $tempFile
```
Crosse:
```suggestion
            -OutFile $tempFile `
            -SkipHttpErrorCheck
```

…and then handle the error manually so that PowerShell doesn't vomit useless response content as the error "message".

scripts/ci/sign-windows.ps1
```diff
@@ -0,0 +1,176 @@
...
+            Write-Warning "Signature status is $($sig.Status) for self-signed/test policy '$SigningPolicy' - this may be expected for self-signed certificates"
+        }
+
+        Write-Host "Signing complete: $fileName"
+        $global:LASTEXITCODE = 0
+        return
+
+    } elseif ($status.Status -eq "Failed" -or $status.Status -eq "Denied") {
+        Write-Error "Signing failed with status: $($status.Status)"
+        exit 1
```
Crosse:
```suggestion
```

scripts/ci/sign-windows.ps1
```diff
@@ -0,0 +1,176 @@
...
+    }
+}
+
+Write-Error "Timeout waiting for signing to complete after $MaxAttempts attempts"
+exit 1
```
Crosse: This is unneeded if the while-loop is restructured for the happy path.

windows/CMakeLists.txt
```diff
@@ -89,6 +80,12 @@ install(FILES "${FLUTTER_ICU_DATA_FILE}" DESTINATION "${INSTALL_BUNDLE_DATA_DIR}
 install(FILES "${FLUTTER_LIBRARY}" DESTINATION "${INSTALL_BUNDLE_LIB_DIR}"
   COMPONENT Runtime)
 
+# Copy liblantern.dll
+set(LANTERN_DIR "..")
```
Crosse: My CMake-fu is a little rusty at this point, but it may be a good idea to use a known root to avoid potential pitfalls in the future, like so:
 
```suggestion
set(LANTERN_DIR "${CMAKE_CURRENT_SOURCE_DIR}/..")
```

scripts/ci/sign-windows.ps1
```diff
@@ -0,0 +1,176 @@
...
+        Write-Host "=== Signature Verification ==="
+        Write-Host "Status: $($sig.Status)"
+        Write-Host "Signer: $($sig.SignerCertificate.Subject)"
+        Write-Host "Thumbprint: $($sig.SignerCertificate.Thumbprint)"
+        Write-Host "=============================="
+
+        if ($sig.Status -ne "Valid") {
+            # Determine whether this is a self-signed/test flow based on the signing policy.
+            # Example policies mentioned in the script header: 'release-signing' and 'test-signing'.
+            $isSelfSignedFlow = $SigningPolicy -like "*test*"
```
atavism: Maybe we should pass an explicit is_self_signed flag into this script instead of inferring from $SigningPolicy -like "*test*"?

atavism (approved):

PR #8447, opened 2026-02-05
Migrate client state off ObjectBox to Radiance

atavism:
Resolves https://github.com/getlantern/engineering/issues/2944

This PR replaces Flutter local DB/local storage reads for user, plans, split-tunneling, private servers, and selected server location with Riverpod providers backed by lantern-core/Radiance APIs (with Go-side on-disk caching)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ LANTERN_LIB_NAME := liblantern
 LANTERN_CORE := lantern-core
 RADIANCE_REPO := github.com/getlantern/radiance
 FFI_DIR := $(LANTERN_CORE)/ffi
-EXTRA_LDFLAGS ?= -X '$(RADIANCE_REPO)/common.constants.AppVersion=$(VERSION)'
+EXTRA_LDFLAGS ?= -X '$(RADIANCE_REPO)/common.Version=$(VERSION)'
 
 DARWIN_APP_NAME := $(CAPITALIZED_APP).app
 DARWIN_LIB := $(LANTERN_LIB_NAME).dylib


jigar-f:
@atavism I think this pretty big change, did we remove obejctbox completly? I thought we were just removing private server stuff. 

- With this change, now every time we need adda new local state, we need to add an ffi method and a platform channel against it, before we just did with object box, only one method.
- With this, now it goes to flutter->ffi/platfomchannel-> go aginest flutter->done

Correct if I'm wrong, I have not looked in detail.

jigar-f:
Agreed, Radiance should definitely be the source for VPN-related functionality. That said, I don’t think Radiance needs to care whether a user has a plan or not, or whether the user is shown an intro, etc. That logic can probably live elsewhere — though I don’t have a strong opinion against it.

I think we should take the time to test everything thoroughly and make sure all flows are working as expected before merging.
Did you get a chance to test all flows, especially auth and purchase?

jigar-f:
@atavism, sorry, I know thisis on me. I will start reviewing this soon, caught up on someother stuff.

lib/core/services/app_purchase.dart
```diff
@@ -278,47 +276,21 @@ class AppPurchase {
...
-    if (user?.legacyUserData != null) {
-      final legacyData = user!.legacyUserData;
-      final subscriptionStatus = legacyData.subscriptionData.status;
-      if (subscriptionStatus == 'active') {
-        return true;
-      }
-      return false;
-    }
-
+    // No durable purchase cache in app state yet.
```
jigar-f: Why? We need to fetch user data here and add a check

jigar-f:
@atavism Created PR for follow-up changes https://github.com/getlantern/lantern/pull/8509 take look when you get chance

PR #8459, opened 2026-02-10
Adds server location submenu to the tray with flags

myleshorton:
Adds a server location submenu to the system tray to allow users to quickly select the location.

The `tray_manager` flutter project is supposedly switching to a native core, so I have not contributed this upstream yet. That said, I'll put together a PR.

### DIFF
--- a/lib/core/common/app_image_paths.dart
+++ b/lib/core/common/app_image_paths.dart
@@ -109,8 +109,6 @@ class AppImagePaths {
   static const privateServerIntro = 'assets/images/private_server_intro.svg';
   static const smartRouteMode = 'assets/images/smart_route_mode.svg';
 
-  static String flagPath(String countryCode) =>
-      'assets/images/flags/${countryCode.toLowerCase()}.png';
 
   /// Validates and returns a safe flag path for the given country code.
   /// Returns null if the country code is invalid or the flag asset doesn't exist.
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -94,7 +94,8 @@ dependencies:
   app_links: ^6.4.1
   # Desktop
   window_manager: ^0.5.1
-  tray_manager: ^0.5.2
+  tray_manager:
+    path: packages/tray_manager
```
jigar-f: @myleshorton I think we should create a new repo for packages and put this chanegs there, and we can use that repo as refernce here, I think that would be a much cleaner solution

myleshorton: Makes sense. All set!

lib/features/system_tray/provider/system_tray_notifier.dart
```diff
@@ -167,6 +166,9 @@ class SystemTrayNotifier extends _$SystemTrayNotifier with TrayListener {
                 return MenuItem(
                   key: 'location_${location.tag}',
                   label: displayName,
+                  icon: location.countryCode.isNotEmpty
+                      ? AppImagePaths.flagPath(location.countryCode)
+                      : null,
                   onClick: (_) => _onLocationSelected(location),
```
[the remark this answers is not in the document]
myleshorton: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/getlantern/lantern/pull/8459#discussion_r2787977280)

PR #8464, opened 2026-02-10
CI: gate nightly/release publishing on successful builds

atavism:
This PR fixes a CI issue where nightly and release workflows could still publish artifacts even when build jobs fail (due to `if: always()` on publish and app store uploads not being gated on overall success)

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -384,15 +384,15 @@ jobs:
       ]
     if: |
       always() &&
-      !cancelled() &&
       github.event.inputs.dry_run != 'true' &&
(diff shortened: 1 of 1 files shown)


jay-418 (commented):
Hey @atavism, I'm sorry I didn't mention this sooner, but I saw the same problem, and decided that having separate release.yaml and nightly.yaml was making for some duplicated and convoluted logic, and I've rewritten this to just use one release file (and extracted some inline code out into scripts.
- https://github.com/getlantern/lantern/pull/8455

I'm happy to do whatever you prefer here:
- close this and only merge mine, or
- merge yours first, then'll I'll rebase mine on the changes and deal with the conflict

PR #8472, opened 2026-02-17
Cleaned up stopping VPN on Windows

myleshorton:
From Claude Code -- will need testing.

```
  1. CmdStopTunnel dispatch (line 368-376) — Made synchronous

  Previously fired vpn_tunnel.StopVPN() in a goroutine and immediately returned "stopped": true. Now it runs synchronously so the Dart client's disconnect() call actually waits for the VPN to
  stop before returning. The Go-side vpn.Disconnect() already has a 10-second internal timeout, so this won't hang.

  2. shutdown() method (lines 157-167) — New

  Explicitly calls vpn_tunnel.StopVPN() and vpn_tunnel.CloseIPC() during service shutdown. Previously neither was called — the service just cancelled the context and closed the pipe listener.

  3. Start() exit path (line 146) — Calls shutdown() on context cancellation

  When the Windows SCM sends Stop/Shutdown, the context is cancelled, the pipe listener closes, and ln.Accept() fails. Now s.shutdown() is called before returning, ensuring the VPN tunnel and
  IPC server are properly cleaned up even if the Dart client didn't send a CmdStopTunnel first (e.g., if the app crashed or the system is shutting down).
```

### DIFF
--- a/lantern-core/wintunmgr/service_windows.go
+++ b/lantern-core/wintunmgr/service_windows.go
@@ -143,6 +143,7 @@ func (s *Service) Start(ctx context.Context) error {
 		conn, err := ln.Accept()
 		if err != nil {
 			if ctx.Err() != nil {
+				s.shutdown()
 				return nil
 			}
 			continue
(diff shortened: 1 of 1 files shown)


lantern-core/wintunmgr/service_windows.go
```diff
@@ -353,15 +366,13 @@ func (s *Service) dispatch(ctx context.Context, r *Request) *Response {
...
-		go func() {
-			events.Emit(ipc.StatusUpdateEvent{Status: ipc.Disconnecting})
-			if err := vpn_tunnel.StopVPN(); err != nil {
-				slog.Error("Error stopping service", "error", err)
-				events.Emit(ipc.StatusUpdateEvent{Status: ipc.ErrorStatus, Error: err})
-			} else {
-				events.Emit(ipc.StatusUpdateEvent{Status: ipc.Disconnected})
-			}
-		}()
+		events.Emit(ipc.StatusUpdateEvent{Status: ipc.Disconnecting})
```
jigar-f, line 369: @myleshorton, this needs to be inside goroutines to get the status back to Flutter.

PR #8473, opened 2026-02-17
Clean up issues stopping the VPN on iOS and MacOS

myleshorton:
This pull request updates the tunnel shutdown logic in the `ExtensionProvider` classes for both iOS and macOS. The main change ensures that the IPC connection is properly closed after attempting to stop the VPN tunnel, and improves error handling for this operation.

Tunnel shutdown improvements:

* In `ios/Tunnel/SingBox/ExtensionProvider.swift`, the tunnel shutdown sequence now calls `MobileCloseIPC` after `MobileStopVPN`, with error logging for both operations. Previously, the code would return early if stopping the tunnel failed, potentially skipping IPC closure.
* In `macos/PacketTunnel/SingBox/ExtensionProvider.swift`, the early return after a VPN stop error was removed, allowing `MobileCloseIPC` to always be called, ensuring consistent resource cleanup.

### DIFF
--- a/ios/Tunnel/SingBox/ExtensionProvider.swift
+++ b/ios/Tunnel/SingBox/ExtensionProvider.swift
@@ -86,11 +86,15 @@ class ExtensionProvider: NEPacketTunnelProvider {
 
   override open func stopTunnel(with reason: NEProviderStopReason) async {
     let startTime = Date()
-
     appLogger.log("(lantern-tunnel) stopping, reason: \(reason)")
     stopService()
+    var error: NSError?
+    MobileCloseIPC(&error)
+    if error != nil {
+      appLogger.log("error closing IPC \(error?.localizedDescription ?? "")")
+    }
     let elapsed = Date().timeIntervalSince(startTime)
-    NSLog("stopTunnel completed in \(elapsed) seconds")
+    appLogger.log("(lantern-tunnel) stopTunnel completed in \(elapsed) seconds")
   }
 
   func opts() -> UtilsOpts {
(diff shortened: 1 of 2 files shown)


ios/Tunnel/SingBox/ExtensionProvider.swift
```diff
@@ -109,7 +109,10 @@ class ExtensionProvider: NEPacketTunnelProvider {
     MobileStopVPN(&error)
     if error != nil {
       appLogger.log("error while stopping tunnel \(error?.localizedDescription ?? "")")
-      return
+    }
+    MobileCloseIPC(&error)
```
jigar-f: If we are closing IPC here, do we need to do the same for Android?

jigar-f: Also, we could just add this  StopVPN in `mobile.go`, and we don't have to update the native code at all.

myleshorton: OK yeah let me look at that!

garmr-ulfr: I forgot to close the IPC server on iOS..we just need to move this to https://github.com/getlantern/lantern/blob/e3fa0cc0fdddd529e114926ea67a8e0729bcd53b/ios/Tunnel/SingBox/ExtensionProvider.swift#L91. The server should not be closed in `stopService`. See https://wdynhnkxvsdx.slack.com/archives/C01K9DJ1ES2/p1771348129194979?thread_ts=1771333497.085179&cid=C01K9DJ1ES2

PR #8477, opened 2026-02-17
set IPC settings path on Linux during initialization

garmr-ulfr:
This pull request introduces a Linux-specific initialization step to the `LanternCore` component, ensuring that the IPC settings path is correctly configured on Linux systems.

Platform-specific initialization:

* In the `initialize` method of `lantern-core/core.go`, added logic to set the IPC settings path using `ipc.SetSettingsPath` when running on Linux, with error handling and logging.

### DIFF
--- a/lantern-core/core.go
+++ b/lantern-core/core.go
@@ -191,6 +191,13 @@ func (lc *LanternCore) initialize(opts *utils.Opts, eventEmitter utils.FlutterEv
 		return fmt.Errorf("unable to create split tunnel handler: %v", sthErr)
 	}
 
+	if runtime.GOOS == "linux" {
+		if err := ipc.SetSettingsPath(context.Background(), settings.GetString(settings.DataPathKey)); err != nil {
+			slog.Error("Failed to set IPC settings path", "error", err)
+			return fmt.Errorf("failed to set IPC settings path: %w", err)
+		}
+	}
+
 	lc.serverManager = lc.rad.ServerManager()
 	lc.apiClient = lc.rad.APIHandler()
 	lc.eventEmitter = eventEmitter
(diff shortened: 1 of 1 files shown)


lantern-core/core.go
```diff
@@ -207,6 +209,13 @@ func (lc *LanternCore) initialize(opts *utils.Opts, eventEmitter utils.FlutterEv
 		userData, _ := core.FetchUserData()
 		slog.Debug("Fetched user data", "data", string(userData))
 	}
+
+	if runtime.GOOS == "linux" {
+		if err := ipc.SetSettingsPath(context.Background(), settings.GetString(settings.DataPathKey)); err != nil {
+			slog.Error("Failed to set IPC settings path", "error", err)
+			return fmt.Errorf("failed to set IPC settings path: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
garmr-ulfr: Moved to earlier in initialize function.

lantern-core/core.go
```diff
@@ -189,6 +191,13 @@ func (lc *LanternCore) initialize(opts *utils.Opts, eventEmitter utils.FlutterEv
 		return fmt.Errorf("unable to create split tunnel handler: %v", sthErr)
 	}
 
+	if runtime.GOOS == "linux" {
+		if err := ipc.SetSettingsPath(context.Background(), settings.GetString(settings.DataPathKey)); err != nil {
+			slog.Error("Failed to set IPC settings path", "error", err)
+			return fmt.Errorf("failed to set IPC settings path: %w", err)
```
jigar-f, line 197: Just confirming, we are returning an error when setting the path fails, not a warning.

garmr-ulfr: I wasn't sure which would be best here, so I opted for returning an error. If this should just log a warning and continue, let me know and I can change it.

jigar-f: @garmr-ulfr question, so if this fails to set pat,h does Linux still work? Or is this related to tunnel?

garmr-ulfr: If it fails to set the path, it won't work. The daemon won't know where to find the data files, _i.e._ config.

PR #8505, opened 2026-03-03
Enable iOS diagnostic logs

atavism:
Resolves https://github.com/getlantern/engineering/issues/3041


ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -1032,12 +1032,10 @@
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = "$(SRCROOT)/Runner/Runner.entitlements";
 				CODE_SIGN_IDENTITY = "Apple Development";
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
-				CODE_SIGN_STYLE = Manual;
+				CODE_SIGN_STYLE = Automatic;
```
jigar-f: Not sure why we changed to Automatic, I think we need Manual code sign here?

ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -1069,7 +1067,6 @@
 				PRODUCT_BUNDLE_IDENTIFIER = org.getlantern.lantern;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE_SPECIFIER = "";
-				"PROVISIONING_PROFILE_SPECIFIER[sdk=iphoneos*]" = "IOS Development";
```
jigar-f, line 1072: Same here, I think we should revert this back orignal.

ios/Runner/Handlers/LogsEventHandler.swift
```diff
@@ -21,20 +29,110 @@ class LogsEventHandler: NSObject, FlutterPlugin, FlutterStreamHandler {
...
+    tearDownStream()
     events = nil
     return nil
   }
+
+  private func startStreaming() {
+    tearDownStream()
+    batchDiffer.reset()
+
+    let client = CommandClient(.log, logMaxLines: Self.logMaxLines)
```
jigar-f: Using CommandClient, I am seeing only one log, which is tunnel started successfully. I don't think we are doing smae other platforms. I think we should do the same as we are doing in macOS and Android. Just read the file. Let me know what you think?

ios/RunnerTests/RunnerTests.swift
```diff
@@ -1,12 +1,40 @@
-import Flutter
-import UIKit
+@testable import Runner
 import XCTest
 
-class RunnerTests: XCTestCase {
+final class RunnerTests: XCTestCase {
```
jigar-f, line 5: Thanks for adding test

ios/Tunnel/SingBox/ExtensionProvider.swift
```diff
@@ -23,7 +23,10 @@ import NetworkExtension
 #endif
 
 class ExtensionProvider: NEPacketTunnelProvider {
+  private static let commandServerMaxLines: Int32 = 1200
+
   private var platformInterface: ExtensionPlatformInterface!
+  private var commandServer: LibboxCommandServer?
```
jigar-f: I think we should be reading `lantern.logs` here, not just app logs. that how we are doing it in other platforms.

lib/features/logs/logs.dart
```diff
@@ -11,43 +12,33 @@ import 'package:lantern/features/logs/log_line.dart';
...
-      WidgetsBinding.instance.addPostFrameCallback((_) {
-        if (!scrollController.hasClients) return;
-        scrollController.jumpTo(scrollController.position.maxScrollExtent);
-      });
-    }
 
     Future<void> shareLogFile() async {
       try {
+        if (Platform.isIOS) {
+          final visibleLogs = latestLogsForDisplay(
```
jigar-f: I think on share we should be sharing the whole file, not just the logs we can see.

test/features/logs/log_stream_screen_test.dart
```diff
@@ -0,0 +1,47 @@
...
+
+import 'package:flutter/material.dart';
+import 'package:flutter_screenutil/flutter_screenutil.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:lantern/features/logs/logs.dart';
+import 'package:lantern/features/logs/provider/diagnostic_log_provider.dart';
+
+void main() {
+  testWidgets('diagnostic logs screen renders streaming updates',
```
jigar-f, line 23: Thanks for adding test

PR #8523, opened 2026-03-12
 Fix Android startup DI

atavism:


lib/core/services/local_storage.dart
```diff
@@ -86,9 +86,12 @@ class LocalStorageService {
     }
 
     dbLogger.debug("Opening ObjectBox store...");
+    final macosAppGroup = Platform.isMacOS && macosApplicationGroup.isNotEmpty
+        ? macosApplicationGroup
+        : null;
     _store = await openStore(
       directory: dbPath,
-      macosApplicationGroup: macosApplicationGroup,
+      macosApplicationGroup: macosAppGroup,
```
jigar-f, line 94: We can remove this, it is not needed at all.

lib/core/services/injection_container.dart
```diff
@@ -16,6 +17,10 @@ import 'logger_service.dart';
 final GetIt sl = GetIt.instance;
 
 Future<void> injectServices() async {
+  if (!sl.isRegistered<Updater>()) {
```
jigar-f: This is called when the app starts, so this will always be true.

lib/core/services/injection_container.dart
```diff
@@ -38,13 +43,16 @@ Future<void> injectServices() async {
     await ps.init();
     sl.registerSingleton<LanternPlatformService>(ps);
 
-    sl.registerLazySingleton<Updater>(() => Updater());
```
jigar-f, line 41: We can keep this as it. I think

lib/core/services/injection_container.dart
```diff
@@ -38,13 +43,16 @@ Future<void> injectServices() async {
...
     if (PlatformUtils.isFFISupported) {
       sl.registerLazySingleton(() => LanternFFIService());
       await sl<LanternFFIService>().init();
+      sl.registerLazySingleton<LanternCoreService>(
+          () => sl<LanternFFIService>());
     } else {
       sl.registerLazySingleton<LanternFFIService>(
           () => MockLanternFFIService());
+      sl.registerLazySingleton<LanternCoreService>(
```
jigar-f: I don't think we should be using LanternCoreService anywhere directly. It should go through Lantern Service always.

lib/core/services/app_purchase.dart
```diff
@@ -124,6 +122,8 @@ class AppPurchase {
     required PaymentSuccessCallback onSuccess,
     required void Function(String error) onError,
   }) async {
+    init();
```
jigar-f: Why do we need this? Since we already have `_waitForProducts` which handles things for us

PR #8562, opened 2026-03-23
Speed up Android CI

atavism:


Makefile
```diff
@@ -486,12 +496,39 @@ android-aab-release:
...
+android-release-artifacts:
+	@dart_defines=$$( \
+	  printf '%s\n' "BUILD_TYPE=$(BUILD_TYPE)" "$(if $(VERSION),VERSION=$(VERSION),)" \
+	    | sed '/^$$/d' \
+	    | while IFS= read -r define; do \
+	        printf '%s' "$$define" | base64 | tr -d '\n'; \
+	        printf ','; \
+	      done | sed 's/,$$//' \
+	); \
+	bash ./android/gradlew -p android $(ANDROID_RELEASE_GRADLE_ARGS) -Pdart-defines="$$dart_defines" assembleRelease bundleRelease
```
jigar-f: Any reason to use `gradlew` directly? I think we should be using Flutter

PR #8563, opened 2026-03-23
Pass version while building lib

jigar-f:
This pull request refactors how the application version is set and used in the `Makefile`. The main improvement is that the `APP_VERSION` is now defined earlier and used to set the `VERSION` variable, which is then consistently referenced throughout the build process.

**Version handling improvements:**

* Moved the definition of `APP_VERSION` (extracted from `pubspec.yaml`) to the top of the `Makefile` and set the `VERSION` variable to default to `APP_VERSION` if not already set. This ensures that the version used in build flags and Dart defines is always consistent with the version specified in `pubspec.yaml`.
* Removed the duplicate definition of `APP_VERSION` from a later section of the `Makefile` to avoid redundancy and potential confusion.


jigar-f:
@jay-418 Please make sure, I am not breaking any CI changes.

PR #8584, opened 2026-03-26
use test policy for nightly buildsn and expose option for workflow dispatch

jay-418:
relates to getlantern/engineering#3109

This moves from prod policy to test policy, but does *not* actually reduce total org-wide artifact upload quotas.


jay-418:
closing because -- on consideration -- I don't think this option actually fixes our problem of artifact quota usage

PR #8607, opened 2026-04-02
Add auth E2E tests and wire Linux/Windows CI

atavism:
Closes https://github.com/getlantern/engineering/issues/3057

### DIFF
--- a/integration_test/auth/auth_smoke_harness.dart
+++ b/integration_test/auth/auth_smoke_harness.dart
@@ -520,7 +520,7 @@ Future<void> _runDeleteAccountClearsSessionScenario(WidgetTester tester) async {
     initialAppSetting: const AppSetting(
       userLoggedIn: true,
       email: _deleteUserEmail,
-      oAuthToken: 'session-token',
+      oAuthToken: '',
       oAuthLoginProvider: 'email',
     ),
     initialUser: _buildFreeUser(email: _deleteUserEmail),


integration_test/auth/windows_auth_smoke_test.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+
+import 'auth_smoke_harness.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  testWidgets('Windows auth smoke suite', (tester) async {
```
jigar-f: If Windows and Linux use the same files, can't we merge them?  Since each test runs on each platform.

integration_test/auth/auth_smoke_harness.dart
```diff
@@ -0,0 +1,569 @@
...
+    expect(settings.oAuthLoginProvider, isEmpty);
+
+    expect(context.fakeService.deleteAccountCalls, 1);
+  } finally {
+    context.dispose();
+  }
+}
+
+Future<void> runAuthSmokeHarness(WidgetTester tester) async {
+  await _runSignInSuccessScenario(tester);
```
jigar-f: Any reason? Why are we running all test no single test?  DO you think we should create a test for each Scenario? Or is there any drawback for that?

lib/features/account/account.dart
```diff
@@ -393,6 +396,7 @@ class Account extends HookConsumerWidget {
           },
         ),
         AppTextButton(
+          key: AuthFlowTestKeys.accountLogoutConfirmButton,
```
jigar-f: I think we should create AppKey files for keys

PR #8638, opened 2026-04-10
Smart location country fix

jigar-f:
This pull request makes a minor update to the server selection logic by removing the code that reset the server location to "auto" after a successful operation. Now, after the operation succeeds, the app simply navigates back to the root without changing the server location.

- Removed the call to `updateServerLocation` with an "auto" server location in `_ServerSelectionState`, so the server location is no longer reset after a successful operation (`lib/features/vpn/server_selection.dart`).


atavism (approved):
LGTM

PR #8648, opened 2026-04-12
fix: publish nightly releases even when some platforms fail

myleshorton:
## Summary

Nightly builds were always deleting the draft GitHub Release regardless of build success, so nightlies were only available via S3. Now nightlies get published as GitHub Releases when at least one platform succeeds.

### Changes to `release-finalize` in `release.yml`:
- **Publish step**: Remove `BUILD_TYPE != 'nightly'` gate — nightlies now publish when uploads succeed
- **Delete step**: Remove `BUILD_TYPE == 'nightly'` unconditional delete — nightlies only deleted on total failure

The upload jobs already handle partial success (they run if any build succeeded), so a broken Linux build won't block the macOS/Windows/Android/iOS artifacts from being released.

### Context
https://wdynhnkxvsdx.slack.com/archives/C06F5EKHN5S/p1776014784078499

## Test plan
- [ ] Trigger a nightly workflow_dispatch with one platform
- [ ] Verify the GitHub Release is published (not deleted)
- [ ] Verify a total failure still cleans up the draft

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -726,7 +726,6 @@ jobs:
       - name: Publish release on success
         if: |
           env.IS_TEST_RUN != 'true' &&
-          env.BUILD_TYPE != 'nightly' &&
           needs.upload-s3.result == 'success' &&
           needs.upload-release-artifacts.result == 'success' &&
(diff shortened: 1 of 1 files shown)


myleshorton:
Revising approach — the issue is that matrix job failures cause upload-s3 to skip entirely.

PR #8657, opened 2026-04-13
fix: derive Location_.protocol from outbound tag

myleshorton:
## Summary

The \`Location_\` model already had a \`protocol\` field with this comment:

\`\`\`dart
// As have default value, we can derive protocol from tag
String protocol = '';
\`\`\`

...but the deriving never happened. Every server row in the location picker rendered with a blank protocol subtitle.

## Why it matters

Came up during today's Reflex rollout: the bandit put reflex outbounds in Frankfurt / Chicago / Stockholm (it rotates as EXP3 learns), and samizdat outbounds in those same cities. Users saw "Frankfurt" and "Stockholm" in the picker but couldn't tell the reflex row apart from the samizdat row because the protocol subtitle was never populated.

## The fix

Tags from lantern-cloud always have the form \`<protocol>-out-<track>-<uuid>\` (e.g. \`reflex-out-reflex-linode-pro-419eebd5-...\`), so the first segment is the protocol. Parse it at the \`Lantern.fromJson\` call site, where we already cascade-set \`tag = k\`.

Kept client-side rather than adding a \`Protocol\` field to \`github.com/getlantern/common.ServerLocation\`. That would require coordinated changes across the whole fleet; this achieves the same visible outcome with one line.

## Test

- Open location picker after merging.
- Expand a country with multiple cities (e.g. U.S.A.).
- Cities that have reflex routes now show "Reflex" as subtitle; samizdat rows show "Samizdat".


myleshorton:
Redundant — `lantern_ffi_service.dart:1722-1740` `applyProtocols` already derives protocol via `outbound.tag → outbound.type` lookup with tag-split fallback. No fix needed on the client side; when reflex-out-... is in outbounds, the Location_ gets "reflex" as protocol.

PR #8658, opened 2026-04-14
only use permalinks

jay-418:
Because a nightly/latest may be from a non-main branch, use only the permalink to reduce confusion.

Release notes and slack messages will now link only list the version-specific permalink.

/latest links will continue to be primary links for public spaces.


atavism:
LGTM!

PR #8662, opened 2026-04-15
mobile: return string instead of []byte from gomobile-exported funcs

myleshorton:
## Summary

Gomobile-exported functions returning `([]byte, error)` can crash with a Go runtime write-barrier panic during GC. This PR converts 9 such functions to return `(string, error)`.

## Root Cause

The gomobile wrapper runs on the cgo callback goroutine (locked to the C thread). After the Go function returns, the wrapper copies the return value into a C struct on the OS thread stack using `runtime.wbMove`. Because `[]byte` contains a pointer (slice data), `wbMove` calls `bulkBarrierPreWrite` to track the write. With an active GC cycle, `bulkBarrierPreWrite` verifies the destination is GC-tracked — the C thread stack is not — and calls `runtime.throw`.

Full analysis: getlantern/engineering#3175

## Affected Functions

- `UserData`, `FetchUserData`
- `GetAvailableServers`
- `OAuthLoginCallback`
- `AcknowledgeGooglePurchase`, `AcknowledgeApplePurchase`
- `Login`, `Logout`
- `DeleteAccount`

## Breaking Changes for Swift/Obj-C

The gomobile-generated Objective-C signatures change from `NSData*` to `NSString*`. Callers that need raw bytes can convert with `.data(using: .utf8)`.

## Source

Diagnosed from Freshdesk [#172640](https://lantern.freshdesk.com/a/tickets/172640) — internal user Derek reported a hard crash on macOS 26.3.1. The crash signature was `runtime.throw` in `runtime.bulkBarrierPreWrite` from `_cgoexp_*_proxymobile__UserData` with a GC mark termination cycle active concurrently.

Note: `radiance/common/gostack.go` has a `RunOffCgoStack` shim that was intended to fix this class of bug. Its comment claims moving the function body to a real goroutine "avoids this entirely", but the return-value marshaling still happens on the cgo callback goroutine — so the shim only protected the body, not the return path. This PR addresses the remaining gap.

## Test Plan
- [x] `go build ./...` passes in lantern repo
- [x] `go vet ./lantern-core/...` clean
- [ ] Swift/Obj-C callers updated to handle `NSString` return types (follow-up)
- [ ] Verify in v9.0.26 beta that the crash no longer reproduces

### DIFF
--- a/lantern-core/mobile/mobile.go
+++ b/lantern-core/mobile/mobile.go
@@ -252,8 +252,15 @@ func StopAutoLocationListener() error {
 }
 
 // GetAvailableServers returns the available servers in JSON format.
-func GetAvailableServers() ([]byte, error) {
-	return withCoreR(func(c lanterncore.Core) ([]byte, error) { return c.GetAvailableServers(), nil })
+//
+// Returns string (not []byte) so the gomobile wrapper marshals the return value
+// via C.malloc rather than leaving it as a Go slice header. This avoids a
+// runtime.bulkBarrierPreWrite panic on the cgo callback goroutine during GC
+// (see getlantern/engineering#3175).
+func GetAvailableServers() (string, error) {
+	return withCoreR(func(c lanterncore.Core) (string, error) {
+		return string(c.GetAvailableServers()), nil
+	})
 }
 
 func IsVPNConnected() bool {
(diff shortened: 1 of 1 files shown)


jigar-f:
@myleshorton Claude suggested the same changes to me as well, but I think we should wait on this PR. I think the refactor PR might bring some changes, and I think it is better to make the changes on the refactor PR, not on this, we will be just doing unnecessary work. 

Of course. If we are planning to put out a new bill, we should, if not until next week we should wait.

myleshorton:
Yeah I was going to basically apply the same changes to the refactor PR as well

myleshorton:
Superseded by #8663, which targets the refactor branch per discussion in Slack (and also includes the Swift-side updates for macOS/iOS).

PR #8668, opened 2026-04-16
Fix false “Daily data cap reached” on first launch

atavism:
Resolves https://github.com/getlantern/engineering/issues/3125

Prevent showing data-cap reached UI when data cap is enabled but missing a valid allotment (bytesAllotted == 0)

### DIFF
--- a/lib/features/home/data_usage.dart
+++ b/lib/features/home/data_usage.dart
@@ -24,7 +24,13 @@ class DataUsage extends ConsumerWidget {
         final dataCap = dataCapResponse.usage!;
 
         /// Do all math in BYTES
-        final int totalBytes = dataCap.bytesAllotted;
+        final int totalBytes = max(0, dataCap.bytesAllotted);
+        if (totalBytes == 0) {
+          appLogger.warning(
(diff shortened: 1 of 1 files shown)


lib/features/home/data_usage.dart
```diff
@@ -24,7 +24,13 @@ class DataUsage extends ConsumerWidget {
         final dataCap = dataCapResponse.usage!;
 
         /// Do all math in BYTES
-        final int totalBytes = dataCap.bytesAllotted;
+        final int totalBytes = max(0, dataCap.bytesAllotted);
+        if (totalBytes == 0) {
```
jigar-f, line 28: @atavism This will create a new edge case where if the user is new, then the user will not see DataCap until the user uses some data.

myleshorton (approved):

lib/features/home/data_usage.dart
atavism: I think if bytesAllotted is 0, it’s better to hide the card than render a false “cap reached” state. And bytesAllotted is the cap/limit

PR #8669, opened 2026-04-16
lantern-core: wire auto-selected events through IPC

myleshorton:
## Summary

On the refactor branch the **Smart Location** tile on Server Selection never shows the current auto-selected server, and tapping it while connected produces *"Unable to connect to VPN server."* Both symptoms trace back to `listenAutoSelectedEvents` using `events.SubscribeContext` — an in-process fan-out. `vpn.AutoSelectedEvent` is emitted inside the packet-tunnel extension's radiance process, so the host app's subscription never fires.

This PR switches `listenAutoSelectedEvents` in `lantern-core/core.go` to use the existing `lc.client.AutoSelectedEvents` SSE stream (`/server/auto-selected/events`) and wraps it in a reconnect loop. The stream itself was already exposed by `ipc.Client`.

## Why the retry loop matters

On macOS the extension's IPC server only binds its Unix socket (`/var/run/lantern/lanternd.sock`) **after** `PacketTunnelProvider.startTunnel` runs — i.e. after the user initiates a connection. Log evidence from the repro session:

```
20:51:33  LanternStatus: Connected
20:51:34  pkg=radiance (*Server).Start.func1  msg="IPC server started" address=/var/run/lantern/lanternd.sock
20:51:43  (*LocalBackend).Close.func1 msg="Closing Radiance"        ← host's local fallback finally shuts down
```

LanternCore initializes (and previously attempted `listenAutoSelectedEvents`) long before `20:51:34`. A single attempt at startup hits `ErrIPCNotRunning` and exits; auto-selected updates are never delivered to Flutter for the rest of the session.

The new loop reconnects with a 500ms backoff until the stream succeeds or `lc.ctx` is cancelled — the same pattern `ipc.Client.TailLogs` already uses for exactly this reason.

## Scope

- **In scope:** auto-selected events.
- **Known adjacent issue not fixed here:** `listenDataCapEvents` has the same one-shot race and will also lose events until the tunnel starts. Same fix applies; tracked in getlantern/engineering#3182.
- **Out of scope, also tracked in getlantern/engineering#3182:**
  - `config.NewConfigEvent` (still on in-process fan-out; needs a new `/config/events` IPC endpoint on the radiance side before lantern-core can subscribe over IPC)
  - Hardening the Dart path in `lib/features/vpn/server_selection.dart:233` so tapping Smart Location with a null `autoLocation` doesn't send an empty tag to the backend

## Test plan
- [x] `go build ./lantern-core/...`
- [x] `go vet ./lantern-core/...`
- [ ] Smoke test on the macOS build: connect, open Server Selection, confirm the Smart Location tile populates with flag + country + protocol after the tunnel is up, and tapping it doesn't produce the error toast

### DIFF
--- a/lantern-core/core.go
+++ b/lantern-core/core.go
@@ -198,7 +198,7 @@ func (lc *LanternCore) initialize(opts *utils.Opts, eventEmitter utils.FlutterEv
 	lc.eventEmitter = eventEmitter
 
 	lc.listenConfigEvents()
-	lc.listenAutoSelectedEvents()
+	go lc.listenAutoSelectedEvents()
 	go lc.listenDataCapEvents()
 
 	slog.Debug("LanternCore initialized successfully")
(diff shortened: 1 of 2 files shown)


garmr-ulfr:
@myleshorton I already reverted the change Jigar did to this section, which was working.

myleshorton:
Closing, as @garmr-ulfr already merged a version of this to 
garmr/radiance-daemon-refactor

PR #8676, opened 2026-04-17
Fix Windows Makefile quoting for version ldflags

atavism:


atavism:
@myleshorton Feel free to combine with https://github.com/getlantern/lantern/pull/8677!

myleshorton:
Closing in favor of #8677, which now includes the non-overlapping pieces of this PR (commit 3fbe19c6a):

- Windows `APP_VERSION` PowerShell quoting swap (outer-single / inner-double)
- `GO_VERSION` Windows PowerShell quoting — same fix
- Windows service executable startup log line

#8677 also adds a few things beyond this PR that together give a more complete fix:
- `APP_VERSION ?=` so Windows CI's `GITHUB_ENV`-exported `APP_VERSION` actually wins over the shell-out (this is the path that was producing empty `common.Version` and 400ing config requests on nightlies — see Freshdesk #172794)
- `APP_VERSION_PUBSPEC` computed via `$(firstword $(subst +, ,$(APP_VERSION)))` (Make built-ins, no shell), so the ldflag value never depends on a shell-out working
- A parse-time `$(error)` guard that fails loudly if the version can't be resolved, instead of silently producing a broken binary like we just shipped

Thanks for the quoting diagnosis — rolled it in verbatim.

myleshorton:
OK I combined your changes into #8677! Should I just go ahead and merge that?

PR #8681, opened 2026-04-19
Makefile: fix empty common.Version on Windows CI (refactor branch port of #8677)

myleshorton:
## Summary

Ports [#8677](https://github.com/getlantern/lantern/pull/8677) to the \`garmr/radiance-daemon-refactor\` branch. The refactor branch's Makefile was at the pre-fix state — Windows CI builds produced an empty \`common.Version\` ldflag, causing every \`/v1/config-new\` request to 400 with \"missing app version\" (Freshdesk #172794, Windows 9.0.26 nightly).

## What's ported

- \`APP_VERSION ?=\` precedence (env > pubspec.yaml) so CI's \`GITHUB_ENV\` export actually wins
- PowerShell quoting swap for local Windows dev fallback (outer-single / inner-double)
- \`APP_VERSION_PUBSPEC := \$(firstword \$(subst +, ,\$(APP_VERSION)))\` using Make built-ins so the linker input doesn't depend on grep/awk/PowerShell
- Parse-time guard that fails the build if \`APP_VERSION_PUBSPEC\` resolves empty

## What's not ported

- The \`lantern-core/cmd/lanternsvc/main.go\` startup log from #8677. That file doesn't exist on refactor — the Windows service is \`lanternd\` in \`radiance/cmd/lanternd\`. That log line should be added in radiance as a separate follow-up PR.
- The Windows \`GO_VERSION\` PowerShell branch — refactor doesn't have an OS-aware \`GO_VERSION\` block (the unified \`grep '^go ' go.mod | awk ...\` line resolves via Git Bash on windows-latest), so leaving that as-is keeps the diff minimal.

## Test plan

- [x] \`make -n lanternd-windows-amd64\` produces \`-ldflags \"-X '.../common.Version=9.0.21'\"\` (non-empty)
- [x] \`APP_VERSION=\"\" make -n lanternd-windows-amd64\` fails with the new parse-time error
- [x] \`APP_VERSION=\"9.9.9+123\" make -n lanternd-windows-amd64\` produces \`common.Version=9.9.9\` (env override wins; +buildnumber stripped via Make built-ins)
- [ ] Trigger \`Windows Service\` workflow on this branch and confirm the emitted \`lanternd.exe\` embeds a non-empty version
- [ ] Verify resulting Windows nightly successfully fetches config from \`/v1/config-new\` (no more 400 \"missing app version\")


atavism:
The actual Makefile fix is present on main now

PR #8695, opened 2026-04-23
android: detach connect() scope + single-flight gate (port of #8689)

myleshorton:
## Summary

Port of [#8689](https://github.com/getlantern/lantern/pull/8689) from `garmr/radiance-daemon-refactor` to `main`.

[#8688](https://github.com/getlantern/lantern/pull/8688) shipped `withTimeout` wrapping, but wrapped the connect() `async` in a structured `coroutineScope { }`. Structured scopes cancel their children on timeout **and then wait for them to finish**. Since `Mobile.startVPN` / `connectToServer` is a blocking JNI call with no suspension points, cooperative cancellation never triggers — the outer `coroutineScope` blocks on the orphan and the UI hangs anyway, defeating the timeout (which is what Freshdesk #173507 reported originally).

## Change

1. **Detached `CoroutineScope(SupervisorJob() + Dispatchers.IO)`** around the connect async. Its Job is not a child of the enclosing coroutine, so cancelling it signals but doesn't join. The orphan keeps running the JNI call in the background until Go returns (or the process exits), but the caller is unblocked and the UI surfaces `${errorCode}_timeout` through the existing `runCatching.onFailure` path.

2. **`AtomicBoolean` single-flight gate** to prevent orphan accumulation across retries. `compareAndSet` rejects concurrent attempts with `IllegalStateException` while a previous one is still in flight; the flag clears in the async block's `try/finally` when the JNI call actually returns. So once Go unsticks, a future retry is admitted. (Copilot's original suggestion was `Dispatchers.IO.limitedParallelism(1)` but that would queue retries behind the stuck orphan — turning the retry into another 60s hang. Fast rejection is the cleaner trade-off.)

## History on the refactor branch
- [#8689](https://github.com/getlantern/lantern/pull/8689) landed the detach in one commit and the single-flight gate in a review-driven follow-up. This PR squashes them into one commit for main.

## Test plan
- [ ] Normal VPN start — timeout never fires, connect completes, status → Connected
- [ ] Simulated deadlock (e.g. `time.Sleep(90s)` in Go `/service/start`) — UI unfreezes after 60s with `*_timeout` error
- [ ] Retry while orphan still running — fails fast with `IllegalStateException("previous VPN connect attempt still in flight")`, not a second 60s hang
- [ ] Orphan eventually completes → future retry succeeds

### DIFF
--- a/android/app/src/main/kotlin/org/getlantern/lantern/service/LanternVpnService.kt
+++ b/android/app/src/main/kotlin/org/getlantern/lantern/service/LanternVpnService.kt
@@ -4,13 +4,13 @@ import android.content.Intent
 import android.net.VpnService
 import android.os.Build
 import android.os.ParcelFileDescriptor
+import java.util.concurrent.atomic.AtomicBoolean
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.TimeoutCancellationException
 import kotlinx.coroutines.async
 import kotlinx.coroutines.cancel
-import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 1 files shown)


myleshorton:
Closing -- focusing on refactor branch.

PR #8704, opened 2026-04-24
Add screenshot attachments to issue reports

atavism:
Resolves https://github.com/getlantern/engineering/issues/3164


myleshorton:
Heads up — there's a use case on the refactor branch that would benefit from the mechanism in this PR.

Context: in #8706 I added `lanterncore.SetupLogging(logDir, level)` which writes a rotating `lantern-core.log` to the UI-process log dir (captures Go slog output from lanterncore + everything it transitively calls when loaded as a library on desktop FFI, where the default stderr handler is silently dropped — so FFI-side errors currently go nowhere on Windows). Report for Freshdesk #173778 / engineering#3335.

I also had an auto-attach of `*.log` from the UI logDir (`<logDir>/*.log` includes `lantern-core.log` and `flutter.log`) into the issue bundle, because the daemon's archive builder only globs its own logDir. But my implementation added a separate `App.ReportIssueWithAttachments(..., attachments []string)` method, which would collide with `attachmentsJSON` here. I just reverted that part of my PR (f5ee94576) so the auto-attach can come back through your mechanism once this lands and gets merged to refactor.

Proposal: when this PR lands and the refactor branch picks it up, add a small FFI-side helper that appends the UI-logDir `*.log` files to the `attachmentsJSON` before handing it to `c.ReportIssue(...)`. That way flutter.log + lantern-core.log are transparently in every issue bundle on desktop without Dart having to know — just as screenshots go through the normal path.

No action for you here. Just flagging so you're not surprised when refactor starts depending on your signature.

lib/features/report_issue/models/report_issue_attachment.dart
```diff
@@ -0,0 +1,59 @@
+import 'package:flutter/foundation.dart';
+
+@immutable
+class ReportIssueAttachment {
```
jigar-f, line 5: Can we move this class to the models folder?

lantern-core/report_issue_attachments.go
```diff
@@ -0,0 +1,213 @@
+package lanterncore
+
+import (
```
jigar-f: Can we create reportissue folder under utils to keep the tree clean?

lib/features/report_issue/models/report_issue_attachment.dart
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  Map<String, dynamic> toJson() => <String, dynamic>{
+    'name': name,
+    'path': path,
+    'mimeType': mimeType,
+    'sizeBytes': sizeBytes,
+  };
+
+  factory ReportIssueAttachment.fromJson(Map<String, dynamic> json) {
```
jigar-f: I don't think we are using this anywhere. Could we remove it?

lib/features/report_issue/models/report_issue_attachment.dart
```diff
@@ -0,0 +1,59 @@
...
+      path: json['path'] as String? ?? '',
+      mimeType: json['mimeType'] as String? ?? '',
+      sizeBytes: (json['sizeBytes'] as num?)?.toInt() ?? 0,
+      securityScopedBookmark: _bookmarkFromJson(json),
+    );
+  }
+
+  static String? _bookmarkFromJson(Map<String, dynamic> json) {
+    final bookmark = (json['securityScopedBookmark'] as String?)?.trim();
+    if (bookmark == null || bookmark.isEmpty) {
```
jigar-f: Same

lib/features/report_issue/models/report_issue_attachment_rules.dart
```diff
@@ -0,0 +1,137 @@
+import 'package:lantern/core/localization/i18n.dart';
+import 'package:lantern/features/report_issue/models/report_issue_attachment.dart';
+import 'package:path/path.dart' as p;
+
+class ReportIssueAttachmentRules {
```
jigar-f: I think a better name would be ReportIssueAttachmentRulesUtils?

lib/features/report_issue/models/report_issue_attachment_rules.dart
```diff
@@ -0,0 +1,137 @@
...
+
+  static String? extensionFor(String source) {
+    final ext = p.extension(source).toLowerCase();
+    if (ext.isEmpty) {
+      return null;
+    }
+    return ext.substring(1);
+  }
+
+  static String displayName(ReportIssueAttachment attachment) {
```
jigar-f: We could do something like this and move it to **ReportIssueAttachment** and use it `attachment.displayName,`

  ```dart
String get displayName {
    final candidate = name.trim().isEmpty ? path.trim() : name.trim();
    if (candidate.isEmpty) {
      return 'attachment';
    }
    return p.basename(candidate);
  }
```

File: lib/features/report_issue/models/report_issue_attachment_rules.dart [hunk 7]
```diff
@@ -0,0 +1,137 @@
...
+    final candidate = attachment.name.trim().isEmpty
+        ? attachment.path.trim()
+        : attachment.name.trim();
+    if (candidate.isEmpty) {
+      return 'attachment';
+    }
+    return p.basename(candidate);
+  }
+
+  static String formatSize(int bytes) {
```
jigar-f: Same for this

lib/features/report_issue/provider/attachment_access.dart
```diff
@@ -0,0 +1,117 @@
...
+import 'dart:io';
+
+import 'package:desktop_drop/desktop_drop.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:lantern/core/common/common.dart';
+import 'package:lantern/features/report_issue/models/report_issue_attachment.dart';
+import 'package:lantern/features/report_issue/models/report_issue_attachment_rules.dart';
+
+final reportIssueAttachmentAccessProvider =
+    Provider<ReportIssueAttachmentAccess>(
```
jigar-f, line 12: Since we are using this as a provider, I think we should drop the service folder and move this inside provider

```
 report_issue/
  ├── report_issue.dart
  ├── provider/                                          
  │   ├── attachment_access.dart        ← from services/
  │   ├── attachment_budget.dart        ← from services/                       
  │   ├── attachment_picker.dart        ← from services/
  │   ├── submitter.dart                ← from services/                       
  │   └── draft_notifier.dart                            
  └── widgets/                                                                 
      └── attachment_dropzone.dart
```

lib/features/report_issue/provider/submitter.dart
```diff
@@ -0,0 +1,97 @@
...
+import 'package:lantern/core/common/common.dart';
+import 'package:lantern/core/utils/device_utils.dart';
+import 'package:lantern/features/report_issue/models/report_issue_attachment.dart';
+import 'package:lantern/features/report_issue/models/report_issue_attachment_rules.dart';
+import 'package:lantern/features/report_issue/services/report_issue_attachment_access.dart';
+import 'package:lantern/features/report_issue/services/report_issue_attachment_budget.dart';
+import 'package:lantern/lantern/lantern_service.dart';
+import 'package:lantern/lantern/lantern_service_notifier.dart';
+
+final reportIssueSubmitterProvider = Provider<ReportIssueSubmitter>(
```
jigar-f, line 14: Same here

jigar-f (commented):
Few comments

atavism:
Thanks @jigar-f I think I've addressed all of them

jigar-f (approved):
LGTM

PR #8707, opened 2026-04-27
ffi + lantern-core: drop non-Linux preflight; bound IPC calls with per-operation timeouts

myleshorton:
## Summary

Two changes that together fix the recurring "lanternd not reachable" first-click bug without introducing the indefinite-hang failure mode that would result from the first change alone.

1. **Drop `checkDaemonReachable` on Windows / macOS / mobile.** The preflight is a vestigial relic of the Linux+systemd flow — its 300 ms timeout was tuned for *fast probe → systemd-rich-error*, and the systemd half only exists on Linux. On other platforms the preflight is just an artificial guillotine in front of `ConnectVPN`. The cold-start IPC path on Windows (named-pipe dial + winio impersonation + H2c connection preface + cold goroutine scheduling) regularly exceeds 300 ms, so the first VPN toggle after every launch trips the timeout and shows the user a spurious error. Linux is unchanged; if we add Windows (`sc query LanternSvc`) or macOS (`launchctl list`) diagnostics later, restore the preflight and call them from there.

2. **Bound per-operation IPC calls with explicit timeouts.** Commit `bd89bea7e` removed the per-call timeouts that used to live on the FFI layer (e.g. `5*time.Second` on `DisconnectVPN`, `8*time.Second` on `ConnectVPN`) at the same time it introduced the bare 300 ms preflight. After that change, the only IPC call with *any* deadline at all was the preflight; every other operation flowed `lc.ctx` (which is `context.WithCancel(context.Background())`) straight through. Drop the preflight without restoring per-call timeouts and a hung lanternd would freeze the UI indefinitely. Restore them at the `LanternCore` layer with values sized for the inherent work each operation does:

| Operation | Timeout | Rationale |
|---|---|---|
| `ConnectVPN` | 60 s | DNS + TLS + sing-box bring-up; can legitimately take many seconds |
| `SelectServer`, `DisconnectVPN` | 30 s | Live-tunnel mutations |
| `VPNStatus`, `IsVPNRunning` | 10 s | Atomic load on the server side; should be near-instant |

The dialer already has a 10 s connect timeout (`radiance/ipc/conn_windows.go`) which covers the lanternd-crashed case. These guard the lanternd-hung case (pipe open, no replies).

Reproduces: same Windows machine across two consecutive beta builds — Freshdesk [#173696](https://lantern.freshdesk.com/a/tickets/173696) (9.0.29) and [#173932](https://lantern.freshdesk.com/a/tickets/173932) (9.0.30). Full archaeology: getlantern/engineering#3382.

## Files

- **`lantern-core/ffi/ffi_nonlinux.go`** — `checkDaemonReachable` becomes a no-op with an explanatory comment about why and a pointer to where to restore it if Windows / macOS get diagnostic equivalents.
- **`lantern-core/core.go`** — adds three timeout constants + wraps `ConnectVPN` / `SelectServer` / `DisconnectVPN` / `VPNStatus` / `IsVPNRunning` with `context.WithTimeout(lc.ctx, ...)`. `CheckDaemonReachable`'s 300 ms is left untouched — Linux's `ffi_linux.go` still calls it for the `systemctl is-active` fallback that's the whole point of the fast probe.
- **`lantern-core/vpn_tunnel/vpn_tunnel.go`** — `StartVPN` / `StopVPN` / `ConnectToServer` take `ctx context.Context` through their signatures instead of building their own `context.Background()`. Caller stays in charge of deadlines.
- **`lantern-core/mobile/mobile.go`** — `Mobile.{StartVPN, StopVPN, ConnectToServer}` set 60 s / 30 s / 60 s contexts on the three gomobile entry points.

## Test plan

- [x] `gofmt` clean on changed files.
- [ ] CI green on Windows + macOS + mobile builds.
- [ ] Manual: install the resulting build, click VPN toggle for the first time after launch — should connect on the first try instead of erroring with "lanternd not reachable". (Was the user-reported regression in #173696 / #173932.)
- [ ] Sanity: simulate a hung lanternd (e.g., add a `time.Sleep(2*time.Minute)` to one of the IPC handlers); verify the UI surfaces an error within the relevant per-call timeout instead of hanging.

## References

- Freshdesk [#173696](https://lantern.freshdesk.com/a/tickets/173696), [#173932](https://lantern.freshdesk.com/a/tickets/173932)
- Engineering issue with full archaeology: getlantern/engineering#3382
- Origin of the preflight: PR getlantern/lantern#8494 (commit `bf054f4ea`, Linux-only at first)
- Generalization that broke it for non-Linux + removed per-call timeouts: commit `bd89bea7e`
- Refactor that hoisted the timeout into LanternCore: PR getlantern/lantern#8578 (commit `4d4e06d9d`)

### DIFF
--- a/lantern-core/core.go
+++ b/lantern-core/core.go
@@ -281,6 +281,20 @@ func (lc *LanternCore) listenDataCapEvents() {
 //     VPN     //
 /////////////////
 
+// Per-call IPC timeouts. These bound the worst case if lanternd is hung
+// (pipe open, no replies). They're long enough to never fire during normal
+// operation — the connect path involves real DNS / TLS / sing-box bring-up
+// that can take many seconds, while a /vpn/status query should be near-
+// instant — but tight enough that a stuck daemon surfaces as a UI error
+// instead of an indefinite spinner. The dialer already has a 10 s connect
+// timeout (radiance/ipc/conn_windows.go), so these only matter once the
+// pipe is established.
+const (
+	ipcConnectTimeout     = 60 * time.Second
+	ipcStateChangeTimeout = 30 * time.Second
+	ipcStatusTimeout      = 10 * time.Second
+)
+
 // ConnectVPN routes a connect request through vpn_tunnel.ConnectToServer,
 // which picks between /vpn/connect (fresh tunnel) and /server/selected
 // (live-tunnel outbound swap) based on VPNStatus. This is load-bearing for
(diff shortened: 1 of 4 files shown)


garmr-ulfr (approved):
LGTM

PR #8708, opened 2026-04-27
qa: plumb RADIANCE_OUTBOUND_SOCKS_ADDRESS into the Android client

myleshorton:
Adds a debug-only path that lets a developer route every outbound network call radiance makes from the Android client through an upstream SOCKS5 (typically the local \`pinger bridge\`), so the bandit treats the client as a real Russia-residential user end-to-end.

Pairs with:
- radiance:      https://github.com/getlantern/radiance/pull/445
- lantern-cloud: https://github.com/getlantern/lantern-cloud/pull/2649

## Summary

- **\`lantern-core/mobile\`**: new gomobile-exported \`SetQAEnvOverrides(socks, tz string) error\` that does \`os.Setenv\` for \`RADIANCE_OUTBOUND_SOCKS_ADDRESS\` and \`TZ\`. Must be called before \`SetupRadiance\` / \`StartIPCServer\` to take effect (radiance reads them at init time).
- **\`android/.../LanternApp.kt\`**: override \`onCreate\` and call the new setter with values from Android system properties:
  - \`debug.lantern.outbound_socks\` → \`RADIANCE_OUTBOUND_SOCKS_ADDRESS\`
  - \`debug.lantern.tz\`             → \`TZ\`
  - Set with \`adb shell setprop debug.lantern.outbound_socks 10.0.2.2:1080\`. No-op when the props are unset, so production builds aren't affected unless someone deliberately sets them on the device.
- **\`go.mod\`**: bump radiance to the qa/outbound-socks-egress branch tip. Should be re-bumped to the merged commit once [radiance#445](https://github.com/getlantern/radiance/pull/445) lands.

## Why two QA props

\`RADIANCE_OUTBOUND_SOCKS_ADDRESS\` makes radiance route through the bridge. \`TZ\` (\`Europe/Moscow\`) is needed because the API's MaxMind logic *overrides* the GeoIP-derived country with the timezone-derived one when they disagree (treats it as a VPN user) — see \`cmd/api/maxmind.go\` \`LookupCountryASNState\`. Without TZ also being Moscow, the API still says \`country=US\` even though the IP is Russia.

## Verified end-to-end

Setup:
\`\`\`
# 1. lantern-cloud bridge (separate PR)
./cmd/pinger/bridge.sh
# → SOCKS5 listening on 127.0.0.1:1080, egress: packetstream/Russia

# 2. emulator boot, install, configure, launch
~/Library/Android/sdk/emulator/emulator @lantern_test -no-snapshot-load -no-boot-anim
adb shell setprop debug.lantern.outbound_socks 10.0.2.2:1080
adb shell setprop debug.lantern.tz Europe/Moscow
adb install -r build/app/outputs/flutter-apk/app-debug.apk
adb shell am start -n org.getlantern.lantern/.MainActivity
\`\`\`

What logcat shows:
\`\`\`
LanternApp: QA env overrides applied: outbound_socks=10.0.2.2:1080 tz=Europe/Moscow
GoLog: QA: set RADIANCE_OUTBOUND_SOCKS_ADDRESS value=10.0.2.2:1080
GoLog: Skipping publicip.Detect because RADIANCE_OUTBOUND_SOCKS_ADDRESS is set
GoLog: routing all radiance HTTP through upstream SOCKS5 addr=10.0.2.2:1080
GoLog: dropping UDP-only protocols from config request kept=15 dropped=5
GoLog: every sing-box outbound will dial via this SOCKS5 addr=10.0.2.2:1080 rewritten_outbounds=7
GoLog: received config "country":"RU","ip":"85.172.81.50"   # Volgograd, AS12389 Rostelecom
\`\`\`

Browsing \`https://api.ipify.org\` in the emulator's Chrome shows the Lantern outbound's exit IP (Stockholm/Singapore depending on which the bandit's auto-pick converges on), **not the Mac's home IP**.

## Test plan

- [ ] **Default (no QA props)**: Lantern boots normally, no \`QA env\` log lines, kindling uses its stacked transports as usual. Production behavior preserved.
- [ ] **With \`debug.lantern.outbound_socks\` set + bridge running**: API sees \`country=RU\` and a Russia residential IP; \`auto\` URLTest converges on a TCP outbound from the Russia-tier set; browsing through the tunnel egresses from the entry-server location.
- [ ] **With invalid system prop value (e.g. unreachable host)**: Lantern logs the QA hook firing but radiance's first \`/v1/config-new\` retries until cancelled; no crash, no production-data leakage.
- [ ] **\`adb shell setprop debug.lantern.outbound_socks ""\`** then re-launch: app reverts to default kindling/no-detour behavior.

## Known caveats

- The bridge SOCKS5 listener is TCP-only (no UDP ASSOCIATE), which is why the radiance PR also drops UDP-only protocols from the request. Hysteria-class protocols don't work in Russia anyway, so this is fine in scope. If we ever want UDP coverage, we'll need to add UDP ASSOCIATE to the bridge.
- The first ~10s after launch may use stale cached config from the previous boot (which can include UDP outbounds and emit transient \`code=7\` errors). Once the fresh \`/v1/config-new\` response arrives, only TCP outbounds are in rotation. Cleaner solution would be wiping \`~/.lantern/data/config.json\` when the QA env is set; deferred.


android/app/src/main/kotlin/org/getlantern/lantern/LanternApp.kt
```diff
@@ -26,5 +28,49 @@ class LanternApp : Application() {
...
+        }
+    }
+
+    /**
+     * Reads an Android system property via reflection on android.os.SystemProperties.
+     * Returns "" if the property is unset or the call fails (e.g. policy denies it).
+     */
+    private fun systemProp(key: String): String {
+        return try {
+            val cls = Class.forName("android.os.SystemProperties")
```
jigar-f, line 55: @myleshorton Lets make sure this is across android version? This can easily cause a crash. Did you run on an older version?

atavism: Thanks, good callout. The whole QA override read/apply path is guarded by try/catch now

PR #8710, opened 2026-04-28
apps_windows: additional Windows app-discovery sources + heuristic noise filter

myleshorton:
Split out from #8706 — the **app-discovery rework** half. The base-bug fix that resolves the actual *"Failed to fetch installed apps"* user reports is at #8709 and merges independently of this PR.

## What this is, and why it's separate

The original investigation chased an empty Windows split-tunnel apps list. Two distinct issues turned up:

1. **The actual user-facing bug** was `GetEnabledApps` returning `nil` (which Dart's `as List` throws on). One-line fix; lives in #8709.
2. **The Windows app-discovery scanner** was also genuinely under-covered — it only looked at Start Menu shortcuts and the Uninstall registry, missing browsers / IDEs / Squirrel apps that register elsewhere. **This PR** is the rework for that.

The two are unrelated in cause and in code. Reviewing them together (originally as #8706) made it hard to evaluate either piece on its own merits.

## Changes

- **HKLM\\Software\\Microsoft\\Windows\\CurrentVersion\\App Paths scanner.** Apps that register here so they're runnable via Win+R / shellexecute. Catches browsers, IDEs, Office, and most third-party apps that don't go through Squirrel or Start Menu.
- **HKLM and HKCU \\Software\\Microsoft\\Windows\\CurrentVersion\\Run.** Squirrel-managed apps (Slack, Discord, VS Code Insiders) register for auto-start with command lines pointing at `Update.exe --processStart`. Same parser as Start Menu `.lnk` targets.
- **%LOCALAPPDATA%\\\\<App>\\\\Update.exe** belt-and-suspenders for Squirrel installs that exist on disk under the well-known layout but haven't shown up in the registry yet.
- **`isAppPathsNoise` filter.** App Paths is heavily polluted by Microsoft-bundled tooling (IE relics, Office helpers, vestigial Mail + tablet apps) and UWP package plumbing. Drops entries under known system paths, `.NET` helper assemblies, anything under `\\Microsoft Office\\` except primary product exes, helper-named basenames (substring match), and basenames suffixed with generic helper words.
- **`ParentKeyName` filter removed from `isNonUserFacingUninstallEntry`.** Introduced in #8641 and over-aggressive — plenty of legitimate end-user apps set it (Squirrel installs, winget packages, MSI bundle children). `SystemComponent=1` and `NoDisplay=1` are the documented signals; those stay.
- **Start Menu COM-failure logs promoted Debug → Warn** (CoInitializeEx, WScript.Shell CreateObject, QueryInterface). These previously failed silently; now an empty result tells us which call broke.
- **Per-filter scan-summary tallies** at the end of each scanner: scanned / kept / per-drop-reason counts plus a sample of the first 20 kept apps for triage. Sample paths are `filepath.Base`'d to avoid PII (full Windows paths typically embed the username).
- **Hot-path constant data hoisted to package scope.** `isAppPathsNoise`'s `systemPaths` / `primaryOfficeExes` / `helperHints` / suffix lists were reallocated on every call (runs hundreds-to-thousands of times per scan); now package-level vars.
- **+152 lines of test coverage** for the new heuristics.

## Risk to be aware of

- **Heuristic filter false positives** are the main thing to watch. The generic suffix list (`update`, `service`, `agent`, `sync`, `broker`) drops anything ending in those words. A legit user-facing app called *"MyApp Service"* would be silently hidden from the split-tunnel UI. Tests cover the patterns we thought of; users have unbounded variety. If someone files *"my app X is missing from split tunnel"* after this lands, that filter is the first place to look.
- **`ParentKeyName` filter removal** changes the uninstall-registry filter behavior slightly. The intent is to recover legitimate apps that were being incorrectly hidden; the risk is now seeing some installer subcomponents we didn't see before. The remaining `SystemComponent=1` / `NoDisplay=1` filters should still catch most non-user-facing entries.

## Test plan

- [x] `gofmt` clean.
- [x] `GOOS=windows go vet ./lantern-core/apps/...` clean.
- [x] `go test ./lantern-core/apps/...` passes locally.
- [ ] CI green.
- [ ] Manual on Windows: confirm previously-missing apps (Slack, Discord, VS Code Insiders, browsers, IDEs) now appear in the split-tunnel "Show Apps" list. Confirm no obvious legit apps are missing that *were* present before.


atavism:
LGTM

PR #8717, opened 2026-04-30
Prevent duplicate payment redirects from rapid checkout taps

atavism:
Partially resolves https://github.com/getlantern/engineering/issues/3393

- Adds in-flight guard around external payment redirect flows
- Disables checkout while a payment redirect request/webview is active
- Releases the guard when redirect setup fails or the payment webview closes


jigar-f (approved):
LGTM

PR #8736, opened 2026-05-06
 send idempotency keys with redirect checkout

atavism:
For https://github.com/getlantern/engineering/issues/3434 and https://github.com/getlantern/engineering/issues/3292


lantern-core/core.go
```diff
@@ -866,29 +866,47 @@ func (lc *LanternCore) SubscriptionPaymentRedirectURL(redirectBody account.Payme
 	return lc.client.SubscriptionPaymentRedirectURL(lc.ctx, redirectBody)
 }
 
-func (lc *LanternCore) StripeSubscriptionPaymentRedirect(subscriptionType, planID, email string) (string, error) {
+func (lc *LanternCore) StripeSubscriptionPaymentRedirect(subscriptionType, planID, email, idempotencyKey string) (string, error) {
+	idempotencyKey, err := normalizePaymentRedirectIdempotencyKey(idempotencyKey)
```
jigar-f, line 879: @atavism Do you think we should just create the **idempotencyKey** key here, since generation is basically the same, so we don't have to pass it through all the platforms? Is there any issue with that?

atavism: @jigar-f I put the key generation in Flutter to tie the key to the user’s checkout attempt. If core creates a new key every time PaymentRedirect is called, two duplicate calls could still become two different payment requests. Do you think the extra platform plumbing is fine given this context?

jigar-f: Fair, we can keep it as it is. But please test, users sometimes try multiple times, just to make sure.

PR #8741, opened 2026-05-08
Fix iOS report issue screenshot picker

atavism:


lib/features/report_issue/models/report_issue_attachment_rules.dart
```diff
@@ -32,6 +32,12 @@ class ReportIssueAttachmentRulesUtils {
     'image/jpg': 'image/jpeg',
   };
 
+  static const List<String> allowedAppleUniformTypeIdentifiers = <String>[
+    'public.png',
+    'public.jpeg',
```
jigar-f, line 43: Not sure, but do we need to add `.HEIC` support since IOS uses that

jigar-f (approved):
LGMT, Just one comment

lib/features/report_issue/models/report_issue_attachment_rules.dart
atavism: Fixed!

PR #8754, opened 2026-05-13
Add Android sideload auto-updates

atavism:
Resolves https://github.com/getlantern/engineering/issues/2240

Adds an Android sideload updater with update checks, manual/startup UX, APK install handoff, and checksum/package/signature validation


android/app/build.gradle
```diff
@@ -91,13 +96,24 @@ android {
         }
     }
 
+    flavorDimensions += "distribution"
```
jigar-f: Why do we need flavor here? We should try to avoid this as much as possible. I think it will create more work for us to maintain both. If possible, we should try to use our `app_cast.xml` here.

jigar-f:
@atavism Is this ready for review? I see you assigned this jay-418. Did you mean to assign this review to me?

jay-418 (approved):
@jigar-f per your comment (https://github.com/getlantern/lantern/pull/8754#issuecomment-4449454312):

I may have been tagged into this just to ensure compatibility with my in-progress autoupdate-server fixes:
- https://github.com/getlantern/lantern-cloud/pull/2635

For which this looks totally good to me, as this will have the following behavior: the client sends a POST to `/update/lantern` with tags.os=android and empty checksum, and gets back a full APK URL with patch_type="" based on the latest production release with an apk.

My approval here is really just confirmation of the above ☝️, I can't speak to the rest of it with any confidence.

jigar-f:
Thanks, @jay-418. Appreciate all the info.

PR #8760, opened 2026-05-14
Prevent multiple Lantern instances on Windows

atavism:
Adds a native Windows single-instance guard so launching Lantern again restores/focuses the existing window instead of starting another Flutter instance

It also forwards Windows app-link activations to the running instance so deep links continue to work

### DIFF
--- a/windows/runner/CMakeLists.txt
+++ b/windows/runner/CMakeLists.txt
@@ -9,6 +9,7 @@ project(runner LANGUAGES CXX)
 add_executable(${BINARY_NAME} WIN32
   "flutter_window.cpp"
   "main.cpp"
+  "single_instance.cpp"
   "utils.cpp"
   "win32_window.cpp"
   "${FLUTTER_MANAGED_DIR}/generated_plugin_registrant.cc"
--- a/windows/runner/main.cpp
+++ b/windows/runner/main.cpp
@@ -3,7 +3,15 @@
 #include <windows.h>
 
 #include "flutter_window.h"
+#include "single_instance.h"
 #include "utils.h"
+
+namespace {
+
+constexpr const wchar_t kSingleInstanceMutexName[] =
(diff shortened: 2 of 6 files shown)


myleshorton (commented):
Looks good overall — the mutex + named ready-event pattern is the right shape, the `AttachThreadInput` dance is the legit way around Windows' foreground-activation restrictions, and the fallback class name handling the upgrade-while-running case is a nice catch. Two clarifying questions before I'd 👍:

1. **`SendAppLink(existing_window)` on a plain double-click launch** — when there's no deep-link URI on the command line (user just clicked the desktop icon), does the `app_links` plugin's Windows handler tolerate this and no-op, or could it forward a garbage/empty URI to the running instance and trigger something unintended in the Flutter side's deep-link listener? Probably fine but I don't see a guard.

2. **Reachability of the "primary instance also calls `ActivateExistingLanternWindow()`" branch** (`main.cpp:38-40`) — this fires when our new build acquires the mutex but still finds a window from an old build via the fallback class name, so the new launcher hands off to the old window and exits. Question: does the Windows installer kill the running Lantern before swapping the binary? If yes this branch is unreachable in normal upgrade flows (no harm, just dead code). If no, a user upgrading while running keeps using the old build silently forever — the new build always hands off and exits until they fully restart Lantern. Worth knowing which one it is.

atavism:
@myleshorton Yep, good catches. I pushed an update for both:

1. Guarded SendAppLink so we only call it when the secondary launch actually has a URI-shaped arg. Plain double-click now just focuses/restores the existing window.
2. Removed the primary-instance fallback handoff. I checked the Inno installer and it has CloseApplications=yes / RestartApplications=no, so normal upgrades should close Lantern first. If the new build acquires the mutex, it should start itself rather than handing off to an older pre-mutex window.

myleshorton:
Nice! I saw we pull this in!

PR #8797, opened 2026-05-19
Fix Play Billing fallback country handling

atavism:
- Uses core/config country as the source of truth instead of device locale for censored-region routing
- Keeps Play Billing fallback state separate from the reported country code
- Normalizes country codes before checking censored regions

### DIFF
--- a/lib/core/services/app_purchase.dart
+++ b/lib/core/services/app_purchase.dart
@@ -113,10 +113,8 @@ class AppPurchase {
       }
     }
 
-    /// All retries exhausted. On Android, treat this as a strong signal that
-    /// Google Play Billing is unreachable (typical in CN/RU/IR even when the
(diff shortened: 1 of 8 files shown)


lib/core/services/app_purchase.dart
```diff
@@ -63,6 +56,31 @@ class AppPurchase {
...
+    return canUsePlayBilling();
+  }
+
+  Future<bool> _initPlayBillingIfAllowed() async {
+    init();
+    if (_subscription != null) {
+      return true;
+    }
+    if (Platform.isAndroid && !CountryCode.isKnown) {
+      await CountryCode.waitUntilKnown();
```
jigar-f: Then this can be an issue if the config fetch does not get the country or delay as it can delay the UI.

PR #8799, opened 2026-05-20
Guard empty auto-selected server tags

atavism:
Resolves https://github.com/getlantern/engineering/issues/3528

Fixes handling of empty auto-selected server events from Radiance.

- Ignores empty auto-selected tags before requesting server details
- Prevents invalid /servers?tag= lookups and decode errors
- Keeps selected-server UI updates unchanged once Radiance emits a real server tag


jigar-f (approved):
LGTM

PR #8804, opened 2026-05-21
lantern-core: propagate identity to settings on Login/Logout/OAuth/DeleteAccount (mobile + FFI)

myleshorton:
## Summary

After login the client keeps sending `/v1/config-new` with the pre-login anonymous `user_id` + `pro_token`, so the backend treats a Pro user as free and the location list shrinks to the free subset (5 in my repro vs the expected ~25). The same shape exists on every platform.

Fix: after each identity-changing entry point, push the new identity through `c.PatchSettings()` so it lands in the canonical settings store both processes (UI + network extension) read from. Mirrors the pattern `AcknowledgeGooglePurchase` / `restoreSubscription` already use.

Applied to **both** `lantern-core/mobile/mobile.go` (iOS + Android + macOS, gomobile path) and `lantern-core/ffi/ffi.go` (Windows + Linux, cgo shared-library path) so all five platforms are covered.

## Why this is needed

The macOS client (and presumably others) runs radiance in two processes — the UI app and the network extension — each with its own in-process settings cache. The on-disk settings.json is root-owned (the NE writes it, last-flusher-wins).

radiance's `account.Client.setData()` (called from `Login`, `Logout`, `OAuthLoginCallback`, and from inside `DeleteAccount` → `NewUser`) updates settings via direct `settings.Set`. That only writes the caller's in-process cache. After login, the UI process knows the user is Pro, but the canonical settings.json (NE-owned) and the NE-side cache stay on the pre-login identity. The config fetcher (running in the NE) keeps sending the old `user_id` + `pro_token`, the backend looks them up as a free anonymous user, and serves only free-tier servers.

`PatchSettings()` goes through the IPC layer that owns the canonical store, which is why purchase flows have always worked. The login path just never adopted the pattern.

## What's covered

| Entry point | mobile.go (iOS/Android/macOS) | ffi.go (Windows/Linux) |
|---|---|---|
| `Login` | ✅ `applyUserDataToSettings` after success | ✅ `applyUserDataToSettings` after success |
| `OAuthLoginCallback` | ✅ Same | ✅ Same |
| `Logout` | ✅ Same | ✅ Same |
| `DeleteAccount` | ✅ Same | ✅ Same |
| `AcknowledgeGooglePurchase` / `…Apple` | Was already calling `PatchSettings` for UserID+Token | (no FFI equivalent) |
| `RestoreGooglePlayPurchase` / `…Apple` | Was already calling `PatchSettings` for UserID+Token | (no FFI equivalent) |

## What's deliberately NOT in scope

- **`SignUp`** — has the same shape (radiance writes settings via direct `Set`), but `LanternCore.SignUp` discards the `SignupResponse`, so a fix here requires an interface change. Left for a follow-up.
- **`CompleteRecoveryByEmail`** — doesn't touch identity; user must call `Login` afterward, which this PR now covers.
- **Purchase paths** — they already use `PatchSettings` for the IDs that matter most for `/v1/config-new`. They do *not* propagate the post-switch `UserLevel` / `Email` / `JwtToken` explicitly (rely on `FetchUserData()` → `setData` for those, which has the same in-process bug). Closing that residual hole would mean a small refactor (extract a shared `patchIdentity` helper, feed it the post-switch `FetchUserData` result). Worth doing as a follow-up; not in scope for the immediate fix.

## Test plan

- [x] `go build ./lantern-core/...`
- [x] `go vet ./lantern-core/...`
- [x] `go test ./lantern-core/...` — all packages pass
- [ ] Manual repro of the original bug on macOS: log into Pro account, verify location list expands to full ~25 servers (instead of the 5 free-tier ones).
- [ ] Same repro on Windows + Linux to validate the FFI mirror.
- [ ] Verify Logout / DeleteAccount don't leave stale Pro identity in settings.


garmr-ulfr:
> The macOS client (and presumably others) runs radiance in two processes — the UI app and the network extension — each with its own in-process settings cache. The on-disk settings.json is root-owned (the NE writes it, last-flusher-wins).

There's only one settings.json file that's shared between the two. The UI app only reads from it when the network extension isn't running, otherwise it requests the data from the extension.

> Applied to both lantern-core/mobile/mobile.go (iOS + Android + macOS, gomobile path) and lantern-core/ffi/ffi.go (Windows + Linux, cgo shared-library path) so all five platforms are covered.

Only the daemon process has a settings.json file on Windows and Linux. The UI has to request everything from it.

> After login the client keeps sending /v1/config-new with the pre-login anonymous user_id + pro_token

@myleshorton, this would actually be a bug in radiance since it's supposed to maintain the user data.

jigar-f:
@myleshorton @garmr-ulfr I'm not sure this change is supposed to be on the client side. I think it should be in radiance?

garmr-ulfr:
I'm pretty sure this was actually caused by @myleshorton somehow having two accounts. https://wdynhnkxvsdx.slack.com/archives/C02FDCPTCF9/p1779386819564389?thread_ts=1779385958.535399&cid=C02FDCPTCF9

jigar-f:
Got it. Since data is coming via a radiance, there is no need to have it here

jigar-f:
I think we don't need this. Closing this for now. Feel free to re-open.

Issue #8731, opened 2026-05-06
[Bug]: the software window issue


echohiccups:
### What happened?

A bug happened!
After closing the software window (by clicking the X in the top-left corner), clicking the icon does not restore the interface; the only option is to exit and restart the app. However, if the window is minimized, clicking the app again will restore the interface.

### Version

9.1.5 Beta

### Which platform(s) does this affect?

- [ ] Android
- [ ] iOS
- [ ] Desktop (Windows)
- [x] Desktop (MacOS)
- [ ] Desktop (Linux)

PR #8806, opened 2026-05-21
Restore macOS window when app icon is clicked

atavism:
Resolves https://github.com/getlantern/engineering/issues/3537 and https://github.com/getlantern/lantern/issues/8731

- Adds a macOS reopen handler for the hidden main Flutter window
- Restores the window when the app is reopened with no visible windows


jigar-f (approved):
LGTM

PR #8812, opened 2026-05-26
macos: continue sysext downgrade when stale deactivation is already missing

atavism:
Fixes a macOS system extension downgrade edge case where the downgrade flow could fail with `OSSystemExtensionErrorDomain error 4` (`extensionNotFound`) during the deactivation step

## Changes

- Treat `extensionNotFound` during replacement deactivation as recoverable when `activateAfter=true`
- Continue by activating the bundled system extension
- Keep manual deactivation behavior unchanged
- Add focused tests for replacement deactivation recovery and non-recovery cases


macos/Runner/VPN/SystemExtensionManager.swift
jigar-f, line 1: @atavism Can cut nightly from this and test it before merging.

PR #8814, opened 2026-05-28
Fix report issue upload attachment size check

atavism:
- stop rejecting report issue attachments solely because Flutter's cached `sizeBytes` differs from the current file size
- use the current file size when preparing attachments, while still rejecting empty files

Resolves https://github.com/getlantern/engineering/issues/3553


lantern-core/utils/reportissue/attachments.go
```diff
@@ -134,8 +134,8 @@ func prepareAttachment(attachment AttachmentMetadata) (preparedAttachment, error
 	if info.IsDir() {
 		return preparedAttachment{}, fmt.Errorf("attachment %q must be a file", name)
 	}
-	if info.Size() != attachment.SizeBytes {
-		return preparedAttachment{}, fmt.Errorf("attachment %q changed on disk before upload", name)
+	if info.Size() == 0 {
```
jigar-f, line 135: Let the server handle all the size issues and validation.

jigar-f (approved):
LGTM

PR #8816, opened 2026-05-28
Fix change-email form submitting the current address

atavism:
Resolves https://github.com/getlantern/engineering/issues/3556

- Keep the change-email field editable instead of inheriting the registered-user disabled/prefilled state.
- Prevent submitting the current account email as the “new” email.


jigar-f (approved):
LGTM

PR #8817, opened 2026-05-28
Fix iOS purchase receipt loss after backend acknowledgment failure

atavism:
Partial fix for https://github.com/getlantern/engineering/issues/3557

- Keep iOS store transactions pending until Lantern backend receipt acknowledgment succeeds
- Retry failed purchase acknowledgments with short backoff, then continue on the capped delay while the app is running
- Persist selected plan metadata so retried purchase updates still send the intended plan ID
- Trust only fresh backend user state before treating a purchase as already active
- Register LanternPlatformService before AppPurchase init so startup purchase updates can safely call backend APIs


lib/core/services/app_purchase.dart
jigar-f, line 1: @atavism, let's not merge this yet. Let's test this thoroughly because at the moment, our Apple payment is stable.

PR #8833, opened 2026-06-02
Warn before pinning unreachable manual servers

atavism:
Resolves https://github.com/getlantern/engineering/issues/3550

- Surface manual-selection risk when a Lantern server has no successful Smart Location probe
- Warn before pinning traffic to a likely unreachable manual server
- Let users either try the selected server anyway or switch back to Smart Location
- Add model tests for fastest-server selection and manual warning behavior


lib/features/vpn/provider/available_servers_notifier.dart
```diff
@@ -50,6 +52,42 @@ class AvailableServersNotifier extends _$AvailableServersNotifier {
...
+    final lastRefresh = _lastReachabilityRefreshAt;
+    if (lastRefresh != null &&
+        now.difference(lastRefresh) < _serverReachabilityRefreshThrottle) {
+      await forceFetchAvailableServers();
+      return;
+    }
+    _lastReachabilityRefreshAt = now;
+
+    for (var attempt = 0; attempt < _urlTestAlreadyRunningAttempts; attempt++) {
+      final result = await ref.read(lanternServiceProvider).runURLTests();
```
jigar-f: Not sure why we need to run tests from here, I think radaince is already doing that, I think that runs as soon app starts
cc @garmr-ulfr let me know if I'm missing anything here

garmr-ulfr: That's correct. The tests run automatically on app start (if a config exists), each time a config is received, and every 3 minutes while the VPN is connected.

atavism: Yep, agreed The current version does not trigger URL tests from `AvailableServersNotifier` anymore

atavism:
Pulling all this in now

PR #8845, opened 2026-06-08
Fix Android VPN state after APK upgrade

atavism:
Resolves https://github.com/getlantern/engineering/issues/3538

### DIFF
--- a/android/app/src/main/kotlin/org/getlantern/lantern/service/LanternVpnService.kt
+++ b/android/app/src/main/kotlin/org/getlantern/lantern/service/LanternVpnService.kt
@@ -1,6 +1,9 @@
 package org.getlantern.lantern.service
 
+import android.content.Context
 import android.content.Intent
+import android.content.pm.PackageInfo
+import android.content.pm.PackageManager
 import android.net.VpnService
 import android.os.Build
 import android.os.ParcelFileDescriptor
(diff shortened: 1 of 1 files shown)


android/app/src/main/kotlin/org/getlantern/lantern/service/LanternVpnService.kt
```diff
@@ -417,6 +430,58 @@ class LanternVpnService :
...
+    private suspend fun resetVpnAfterAppUpgradeIfNeeded() {
+        if (!consumeUpgradeResetMarker()) return
+
+        AppLogger.i(TAG, "App APK updated; resetting VPN state before first tunnel start")
+        stopVPNTunnel()
+        VpnStatusManager.postVPNStatus(VPNStatus.Disconnected)
+        delay(UPGRADE_RESET_SETTLE_MS)
+    }
+
+    private fun consumeUpgradeResetMarker(): Boolean {
```
jigar-f, line 445: @atavism doesn't OS kill the app when the user installs or updates the apps?

atavism: @jigar-f Yes, the OS kills the app process on update, but it doesn’t clear everything an uninstall would. App data, and some Android-managed VPN state can survive an in-place APK update.
So this is meant as a one-time cleanup on the first VPN start after an APK update, to make the upgrade path behave closer to a fresh install before we establish the new TUN.

jigar-f (approved):
LGTM

PR #8849, opened 2026-06-09
Avoid StoreKit product lookups during iOS startup

atavism:
Keeps StoreKit product queries out of the normal iOS launch path while still initializing purchase, restore, and pending purchase recovery flows when they are actually needed

### DIFF
--- a/lib/core/services/app_purchase.dart
+++ b/lib/core/services/app_purchase.dart
@@ -54,6 +54,9 @@ class AppPurchase {
   // emit an empty stream event the way Play Billing does).
   bool _restoreReceivedAny = false;
 
+  /// Starts listening for store purchase updates without fetching product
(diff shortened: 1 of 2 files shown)


lib/core/services/app_purchase.dart
```diff
@@ -68,11 +71,6 @@ class AppPurchase {
       onDone: _updateStreamOnDone,
       onError: _updateStreamOnError,
     );
-    unawaited(
-      fetchSubscriptions().catchError((Object e, StackTrace st) {
-        appLogger.error('[AppPurchase] init: fetchSubscriptions failed', e, st);
-      }),
-    );
```
atavism, line 75: This is the core change: now we only call `fetchSubscriptions()` when the purchase UI actually needs product details

jigar-f (approved):
LGTM.

PR #8856, opened 2026-06-11
Fix Windows report issue and pro account routing

atavism:
- Return structured FFI errors from desktop report submission so failed Freshdesk submissions are surfaced instead of shown as successfully submitted.
- Reuse the registered Pro account state when opening Account from Home/Settings, so users who already have a password are not incorrectly prompted to set one again.

### DIFF
--- a/lib/features/home/home.dart
+++ b/lib/features/home/home.dart
@@ -4,7 +4,6 @@ import 'package:flutter_hooks/flutter_hooks.dart';
 import 'package:flutter_screenutil/flutter_screenutil.dart';
 import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:lantern/core/common/app_text_styles.dart';
-import 'package:lantern/core/extensions/user_data.dart';
 import 'package:lantern/core/models/feature_flags.dart';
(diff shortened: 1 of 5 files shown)


lib/features/home/home.dart
```diff
@@ -122,18 +121,15 @@ class _HomeState extends ConsumerState<Home> {
...
                 final userSignedIn = ref.read(appSettingProvider).userLoggedIn;
-                final email = localUser!.legacyUserData.email;
-                final isPro = localUser.legacyUserData.isPro;
-                if (isPro && !userSignedIn) {
-                  // this means user has pro account but not signed in
-                  await showProAccountFlowDialog(
-                    context: context,
-                    hasEmail: email.isNotEmpty,
-                  );
+                if (localUser == null) {
```
jigar-f: I do not think we should push sign-in that will cause confusion to users, if the data is null (which will be a rare case), then just show a snackbar error. What do you think?

jigar-f (approved):
Just one comment, otherwise LGTM

lib/features/home/home.dart
atavism: Makes sense!

PR #8858, opened 2026-06-12
Fix Windows CI build on newer MSVC

atavism:
Adds MSVC experimental coroutine compatibility define to the Windows Flutter plugin targets that need it: `flutter_inappwebview_windows_plugin` and `flutter_local_notifications_windows`

Tested here: https://github.com/getlantern/lantern/actions/runs/27430885637


jay-418 (approved):
Seems right for the failure we saw.

PR #8874, opened 2026-06-18
Treat sustained probe failures as unreachable

atavism:
Follow-up to the testing-vs-unreachable split. This makes the current failure threshold win over stale successful latency, so servers with old `lastSuccessDelayMs` but repeated current probe failures are still treated as unreachable and excluded from fastest/location representative selection.

### DIFF
--- a/lib/core/models/available_servers.dart
+++ b/lib/core/models/available_servers.dart
@@ -35,15 +35,18 @@ class AvailableServers {
     return null;
   }
 
-  /// Lantern server with the lowest successful probe delay. Null when no
-  /// Lantern server has a successful probe.
+  /// Lantern server with the lowest usable successful probe delay. Null when no
+  /// Lantern server has a successful probe that is not currently hard-failing.
   Server? get fastestLanternServer {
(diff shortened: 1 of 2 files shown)


myleshorton (approved):
LGTM. Verified locally: `flutter analyze` clean, 14/14 model tests pass.

Closes the real gap in #8873 — `recordProbeFailure` never clears `lastSuccessDelayMs`, so a server that succeeded once but is now hard-failing stayed `hasSuccessfulProbe == true` and could still be auto-selected. Dropping the `!hasSuccessfulProbe` guard and excluding `isProbedUnreachable` from the fastest/representative pools is the right call.

The new overlap (a server can now be both `hasSuccessfulProbe` and `isProbedUnreachable`) is handled safely everywhere: the row trailing checks `isProbedUnreachable` first, the section split keys off `shouldWarnBeforeManualSelection`, and no widget reads `hasSuccessfulProbe` directly. Only trade-off is a healthy server hit by a transient 3-failure burst reads "unreachable" until its next success resets the counter — intended "current evidence wins" semantics.

Suggest merging #8873 first, then this.

PR #8875, opened 2026-06-18
Treat untested manual servers as normal

atavism:
Remove the visible “Testing…” spinner/state for manual server locations that have not received probe results yet

Resolves https://github.com/getlantern/engineering/issues/3637


jigar-f (approved):
LGTM

PR #8888, opened 2026-06-26
Remove unavailable server UI

atavism:
## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified VPN server selection by removing “may be unreachable” warning prompts and the related warning icon behavior.
  * Pro users now see a single unified locations list, and selecting a server proceeds directly after any required permission check.
  * Location settings UI was cleaned up: the subtitle no longer includes reachability warning text, and the location picker no longer shows per-item “may be unreachable” indicators.
* **Chores**
  * Removed obsolete English localization strings related to server reachability warnings.


lib/features/vpn/single_city_server_view.dart
```diff
@@ -1,28 +1,22 @@
...
 import 'package:lantern/core/models/available_servers.dart';
-import 'package:lantern/features/vpn/server_selection.dart';
 
 import '../../core/common/common.dart';
 
 // single_city_server_view.dart
 
 class SingleCityServerView extends StatefulWidget {
   final Server server;
-  final OnServerSelected onServerSelected;
```
jigar-f, line 11: I think we can keep `OnServerSelected` Any reason we are not using that ?

jigar-f (approved):
LGTM, just one comment

lib/features/vpn/single_city_server_view.dart
atavism: Fixed!

atavism:
Thanks for the review @jigar-f!

PR #8889, opened 2026-06-26
fix(android): recover default-interface index via Os.if_nametoindex

garmr-ulfr:
## Problem

On Android, every sing-box outbound (`direct`, the auto-selector, and proxy outbounds) and DNS failed with `no available network interface`, while the UI still reported the VPN as **Connected**. Observed on Android 10 / API 29 (Samsung SM-G960U1); the condition persisted for the whole session with zero successful connections, and the auto-selector churned through servers marking each `hard_demoted` — a consequence, not the cause.

## Root cause

`DefaultNetworkMonitor.checkDefaultInterfaceUpdate` is the only path that delivers the default network interface to sing-box (via `updateDefaultInterface`). It resolved the interface index solely through `NetworkInterface.getByName(name).index`, which relies on the interface-enumeration syscall. That syscall is restricted on some Android versions/ROMs, where `getByName` returns null or throws. On failure the 10-retry loop exhausted and returned **without** ever calling `updateDefaultInterface`, leaving sing-box with no default interface — so every outbound failed to bind.

The sibling `PlatformInterfaceWrapper.getInterfaces()` was already hardened for this exact case with an `Os.if_nametoindex` fallback; `checkDefaultInterfaceUpdate` was not.

## Fix

Fall back to `Os.if_nametoindex(interfaceName)` when `getByName` cannot resolve the index, mirroring the hardened sibling, and retry only when the index is still unresolvable (`<= 0`). Also early-return on a null interface name instead of burning the retry budget on `getByName(null)`.

## Testing

Needs a build + on-device check on an API 29 device to confirm.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network detection reliability, helping the app recognize the active connection more consistently.
  * Made interface resolution more resilient on Android, reducing missed updates and stale network state.
  * Added fallback handling for cases where interface details aren’t available through the usual system path.


garmr-ulfr:
@atavism since PR https://github.com/getlantern/lantern/pull/8890 was based off of, and merged into, this one, should I just go ahead and merge this into main? Both modified the same file, which is now identical between them.

atavism (approved):
@garmr-ulfr Yep, LGTM!

Issue #549, opened 2013-02-07
Switch to org.getlantern


leahxschmidt:
Since we don't actually own lantern.org

PR #8897, opened 2026-07-06
deps: bump radiance — proxyless rule-set detour (#549)

myleshorton:
## What

Bump `getlantern/radiance` `686b6b9` → `4356bc0`.

## Changes pulled (5 commits)

| Commit | Impact |
|---|---|
| `4356bc0` **radiance#549** — client honors `ConfigResponse.NonSelectableOutbounds` | the change we want |
| `166e907` iOS-only mobile memory-limits fix | fix (low risk) |
| `737cdb7` / `1485f2a` / `01d51f2` stale-bot CI workflow | CI-only, no runtime impact |

Transitively bumps `getlantern/common` to the merged `NonSelectableOutbounds` field (`common#27`).

## Why

This is the **client half** of the CN geosite cold-start fix. With #549, radiance merges server-declared infrastructure outbounds (e.g. the proxyless rule-set-fetch detour) into the box config so their references resolve, but excludes them from **both** the auto (URLTest) group and the manual selector — so such an outbound can never be auto-selected or user-picked (which would route traffic directly and expose the user's IP).

It's **inert** until the server (lantern-cloud#2936) enables the proxyless detour — which is itself gated behind a feature flag **and** a `PROXYLESS_MIN_CLIENT_VERSION` floor set to the release carrying this bump. So this needs to ship first, then the server flag is enabled.

Context: getlantern/engineering#3657.

## Testing

`go build ./...` passes; `go mod tidy` clean (only radiance + the transitive `common` bump).

https://claude.ai/code/session_01WGpKfgkbfdJwwsTaouMDoR

## Summary by CodeRabbit

* **Chores**
  * Updated two indirect dependencies to newer revisions for improved stability and compatibility.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ replace github.com/quic-go/qpack => github.com/quic-go/qpack v0.5.1
 require (
 	github.com/alecthomas/assert/v2 v2.3.0
 	github.com/getlantern/lantern-server-provisioner v0.0.0-20251031121934-8ea031fccfa9
-	github.com/getlantern/radiance v0.0.0-20260702205513-686b6b916dcb
+	github.com/getlantern/radiance v0.0.0-20260706174732-4356bc07f9a0
 	github.com/sagernet/sing-box v1.12.22
 	golang.org/x/mobile v0.0.0-20250711185624-d5bb5ecc55c0
 	golang.org/x/sys v0.45.0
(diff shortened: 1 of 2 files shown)


myleshorton:
Heads-up: the gating design changed from a hardcoded client-version floor to **capability negotiation**. lantern-cloud now gates the proxyless detour on the client advertising `non_selectable_outbounds` (getlantern/common#28), which radiance only does in getlantern/radiance#555. So this bump should be **re-pointed to a radiance commit that includes both #549 and #555** before the release — otherwise the client won't advertise the capability and the (capability-gated) server will never send the detour. Will re-point once #555 merges.

PR #8899, opened 2026-07-08
Run macOS connect smoke on self-hosted runner

atavism:
Splits the macOS connect smoke into a dedicated self-hosted CI job. The macOS build still runs on the normal hosted workflow, then the signed app artifact is downloaded and tested on the persistent lantern-macos-smoke runner


jigar-f (approved):
LGTM

atavism:
Thanks, @jigar-f!!

PR #8904, opened 2026-07-09
fix(android): avoid API 26 network callback overload on Android 7

atavism:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Android network callback registration across different OS versions.
  * Fixed behavior on older Android releases so the app uses the supported callback method for each API level.


jigar-f (approved):
LGTM

PR #8909, opened 2026-07-14
Run macOS builds and connect smoke on self-hosted CI

atavism:
Moves macOS builds to the persistent ARM64 runner and runs the VPN connect smoke in the same job, reusing dependencies and build artifacts

## Summary by CodeRabbit

* **Bug Fixes**
  * macOS smoke testing now normalizes installer paths and avoids redundant copy/remove when the source and destination resolve identically.
  * Notarization now fails immediately with clearer logs when the notarization status isn’t **Accepted**.
* **Improvements**
  * macOS connect/disconnect smoke testing is consolidated into the main macOS workflow, with configurable options (including IP checks, full-tunnel smoke, and extension activation timeout).
  * macOS framework builds now write directly to the final output location.
  * macOS signing is centralized for stricter validation and safer component-by-component signing with signature verification.


myleshorton (approved):

atavism:
Thanks @myleshorton!

PR #8925, opened 2026-07-23
Fix referral reward UI edge cases

atavism:
Follow-up fixes for #8916:

- Make the Invite Friends screen scrollable on small screens and with large text.
- Persist converted referrals as seen only after the reward dialog is dismissed.
- Reset reward-dialog lock visibility after errors.
- Fix duplicated localized referral reward copy.
- Prevent delayed clipboard animations from updating disposed widgets.
- Replace the deprecated sharing API.
- Add referral model parsing and bonus-calculation tests.
- Merge the latest `main` and resolve the Radiance dependency conflict.

### DIFF
--- a/lib/features/auth/choose_payment_method.dart
+++ b/lib/features/auth/choose_payment_method.dart
@@ -525,20 +525,12 @@ class PaymentCheckoutMethods extends HookConsumerWidget {
               ),
               DividerSpace(padding: EdgeInsets.symmetric(vertical: 10)),
               if (referral.isReferral) ...[
-                Row(
-                  mainAxisAlignment: MainAxisAlignment.spaceBetween,
-                  children: [
-                    Text(
-                      getReferralMessage().replaceAll('free', '').toTitleCase(),
-                      style: theme.bodyMedium,
-                    ),
-                    Text(
-                      'free'.i18n,
-                      style: theme.bodyMedium!.copyWith(
-                        color: context.textDisabled,
-                      ),
-                    ),
-                  ],
+                Align(
+                  alignment: AlignmentDirectional.centerStart,
+                  child: Text(
+                    getReferralMessage().toTitleCase(),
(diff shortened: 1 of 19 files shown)


atavism:
Just a couple follow-ups to https://github.com/getlantern/lantern/pull/8916 if you find any of this useful @jigar-f

lib/features/auth/choose_payment_method.dart
```diff
@@ -525,20 +525,12 @@ class PaymentCheckoutMethods extends HookConsumerWidget {
...
               DividerSpace(padding: EdgeInsets.symmetric(vertical: 10)),
               if (referral.isReferral) ...[
-                Row(
-                  mainAxisAlignment: MainAxisAlignment.spaceBetween,
-                  children: [
-                    Text(
-                      getReferralMessage().replaceAll('free', '').toTitleCase(),
-                      style: theme.bodyMedium,
-                    ),
-                    Text(
```
jigar-f, line 535: We still need this in the checkout UI, Adding it back

lib/features/setting/referral_reward_dialog.dart
```diff
@@ -30,30 +30,29 @@ Future<void> checkAndShowReferralReward(
...
 
   final newDays = unseen.fold<int>(0, (total, r) => total + r.bonusDaysEarned);
   final newMonths = max(1, newDays ~/ 30);
   final totalMonths = max(newMonths, user.legacyUserData.referralBonusMonths);
 
-  if (!context.mounted) {
+  if (!context.mounted) return;
+  _rewardDialogVisible = true;
+  try {
+    await _showReferralRewardDialog(
```
jigar-f: We don't need to use await here, we can show it and mark it as done, Since we know user has seen this.

jigar-f (approved):
LGTM, I pushed some changes

PR #8936, opened 2026-07-29
fix(ios): send empty device ID from network extension

garmr-ulfr:
## Summary

The iOS main app and network extension run as separate processes. Keychain items are not shared across processes without a shared `keychain-access-groups` entitlement — so the extension's `DeviceIdentifier.getUDID()` produced a per-process device ID that diverged from the main app's, and each sent a different ID in config and telemetry requests.

The extension is not authoritative for the device ID, so it now sends an empty `opts.deviceid`. Radiance resolves the shared ID from the app-group settings the main app persisted, so both processes use the same ID.

## Companion PR

Requires getlantern/radiance#583, which adds the radiance-side resolution (caller → persisted settings). The two must merge together.

closes getlantern/engineering/issues/3701

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device identification handling for the iOS tunnel extension by using shared, app-provisioned settings instead of direct device ID retrieval.
  * Prevented inconsistent device ID access between the main app and its extension.
* **Chores**
  * Updated an internal library dependency to a newer version for continued stability and improvements.

### DIFF
--- a/ios/Tunnel/SingBox/ExtensionProvider.swift
+++ b/ios/Tunnel/SingBox/ExtensionProvider.swift
@@ -109,9 +109,9 @@ class ExtensionProvider: NEPacketTunnelProvider {
     let opts = UtilsOpts()
     opts.dataDir = FilePath.dataDirectory.relativePath
     opts.logDir = FilePath.logsDirectory.relativePath
-    // Intentionally left empty. Keychain items aren't shared across processes,
-    // so a per-process ID here would diverge from the main app's. Radiance
-    // resolves the device ID from the shared app-group settings the app persists.
+    // Intentionally left empty. The app and extension don't share a keychain access
+    // group so `DeviceIdentifier.getUDID` would diverge. Radiance resolves the device 
+    // ID from the shared app-group settings the app persists.
     opts.deviceid = ""
     opts.logLevel = "trace"
     opts.locale = Locale.current.identifier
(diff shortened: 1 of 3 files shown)


ios/Tunnel/SingBox/ExtensionProvider.swift
```diff
@@ -109,7 +109,10 @@ class ExtensionProvider: NEPacketTunnelProvider {
     let opts = UtilsOpts()
     opts.dataDir = FilePath.dataDirectory.relativePath
     opts.logDir = FilePath.logsDirectory.relativePath
-    opts.deviceid = DeviceIdentifier.getUDID()
+    // Intentionally left empty. The app and extension don't share a keychain access
+    // group so `DeviceIdentifier.getUDID` would diverge. Radiance resolves the device 
+    // ID from the shared app-group settings the app persists.
+    opts.deviceid = ""
```
jigar-f, line 114: Perfect. I just wanted to confirm this: we can't share keychain access on the extension. Thanks.

ios/Tunnel/SingBox/ExtensionProvider.swift
```diff
@@ -109,7 +109,10 @@ class ExtensionProvider: NEPacketTunnelProvider {
...
     opts.logLevel = "trace"
```
jigar-f, line 115: @garmr-ulfr I would update the comment to 

// Intentionally left empty. The app and extension don't share a keychain access
// Radiance resolves the device ID from the main app process.

jigar-f (approved):
LGTM

ios/Tunnel/SingBox/ExtensionProvider.swift
garmr-ulfr: Updated.

PR #8938, opened 2026-07-29
fix(ios,macos): tear down tunnel on failed NE restart

garmr-ulfr:
## What

On a failed VPN tunnel *restart*, the network extension's `restartService()` swallowed the `MobileStartVPN` error and returned, leaving the extension alive with a dead tunnel the system still reported as connected. radiance's `Restart` then saw a nil error and treated the restart as successful, so the client neither recovered nor surfaced the failure — the "dead but connected" state.

## Change

- `restartService()` now `throws` on failure, calls `cancelTunnelWithError` (matching the initial-connect paths `startVPN`/`connectToServer`), and propagates the error.
- The platform-interface wrapper rethrows through the throwing `runBlocking` overload, so radiance's `RestartService()` returns the error and the VPN client transitions to `ErrorStatus` instead of "restarted successfully."
- Applied to both the iOS and macOS extensions.

related to getlantern/engineering#3700.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved VPN service restart error handling on iOS and macOS.
  * Restart failures are now correctly reported instead of being silently ignored.
  * Failed restarts now safely terminate the affected tunnel, helping prevent unstable or partially active VPN connections.

### DIFF
--- a/ios/Tunnel/SingBox/ExtensionProvider.swift
+++ b/ios/Tunnel/SingBox/ExtensionProvider.swift
@@ -125,7 +125,7 @@ class ExtensionProvider: NEPacketTunnelProvider {
     postServiceClose()
   }
 
-  func restartService() {
+  func restartService() throws {
     appLogger.log("(lantern-tunnel) restarting service")
     reasserting = true
     defer {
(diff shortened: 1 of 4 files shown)


ios/Tunnel/SingBox/ExtensionProvider.swift
```diff
@@ -125,20 +125,23 @@ class ExtensionProvider: NEPacketTunnelProvider {
...
     var error: NSError?
     MobileStartVPN(&error)
     if let error {
-      appLogger.log("(lantern-tunnel) restart failed: \(error.localizedDescription)")
-      return
+      appLogger.error("(lantern-tunnel) restart failed: \(error.localizedDescription)")
+      // A failed (re)start must tear the tunnel down so on-demand/the app can recover
+      // rather than leaving a dead-but-"connected" tunnel; the throw propagates the
+      // failure to radiance's Restart so it reports ErrorStatus instead of success.
+      cancelTunnelWithError(error)
```
jigar-f, line 143: Not sure how I missed that, but nice catch.

jigar-f (approved):
LGTM

PR #8940, opened 2026-07-31
Fix staging issue on android.

jigar-f:
This pull request improves and standardizes how the staging environment is detected and set in the codebase, reducing the risk of inconsistent environment checks and ensuring the correct environment is configured before backend initialization. The main changes are as follows:

**Environment detection and configuration:**

* Added the `IsStaging()` method to the `Opts` struct in `utils/common.go` to centralize and standardize staging environment checks. This method ensures that all checks for staging use a single source of truth and accept both `"stage"` and `"staging"` as valid values.
* Updated `LanternCore.initialize` in `core.go` to use the new `opts.IsStaging()` method instead of directly comparing the `Env` string, improving maintainability and consistency.
* In `mobile/mobile.go`, before backend initialization in `StartIPCServer`, added a call to `env.SetStagingEnv()` if `opts.IsStaging()` is true. This ensures the environment is set early enough, especially on Android where initialization order previously caused issues.
* Imported the `env` package in `mobile/mobile.go` to support the new environment setup logic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved staging environment detection across core and mobile components.
  * Ensured staging settings are applied before starting the local backend.
  * Added support for both `stage` and `staging` environment labels.

### DIFF
--- a/lantern-core/core.go
+++ b/lantern-core/core.go
@@ -223,7 +223,7 @@ func (lc *LanternCore) initialize(opts *utils.Opts, eventEmitter utils.FlutterEv
 	}
 	slog.Debug("Starting LanternCore initialization")
 
-	if opts.Env == "stage" || opts.Env == "staging" {
+	if opts.IsStaging() {
 		slog.Debug("Setting staging environment")
 		env.SetStagingEnv()
 	}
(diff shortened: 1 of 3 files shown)


jigar-f:
Small changes, pulling it in.

PR #8951, opened 2026-08-05
Prevent dialog callbacks from popping underlying screens

atavism:
Preserves the existing AppDialog.dialog callback contract after the dialog refactor, preventing callbacks from accidentally popping both the dialog and the underlying screen. Adds a widget regression test for the navigation behavior.


jigar-f (approved):
LGTM

PR #8952, opened 2026-08-05
Update deferred Stripe renewal and retry handling

atavism:
- Handle both PaymentIntent and SetupIntent subscription flows
- Prevent PaymentSheet retries from creating duplicate subscriptions
- Adds Stripe tests covering both intent modes, retries, initialization, and error filtering


jigar-f (approved):
LGTM, Thanks for adding test

PR #8956, opened 2026-08-06
capture and validate Stripe Checkout screenshots

atavism:
Captures the Stripe Checkout WebView directly during the macOS payment smoke test, waits until it contains visible rendered content, and saves the resulting PNG with the test artifacts.


jigar-f (approved):
LGTM

PR #8967, opened 2026-08-11
Restore desktop auto-update signing and publishing

atavism:
Sign and verify macOS and Windows updates using dedicated native-runner CI scripts, then publish platform-correct metadata for Lantern Cloud’s update service. This also removes the obsolete legacy appcast publisher, restores feed verification, and preserves valid releases when metadata publishing fails.

## Summary by CodeRabbit

* **New Features**
  * Added signed update support for macOS and Windows installers.
  * Release metadata now includes platform signatures and Sparkle version information.
  * Added desktop auto-update smoke testing and reusable update fixtures for macOS and Windows.
  * Windows builds now include the public key required to verify updates and receive the application build version.
  * Added configurable profile builds and update-test options.

* **Bug Fixes**
  * Improved signature validation, update verification, and shutdown handling before Windows updates.

* **Chores**
  * Removed the legacy appcast generation process and unused dependency.


.github/scripts/sign_windows_update.ps1
```diff
@@ -0,0 +1,107 @@
...
+param(
+  [Parameter(Mandatory = $true)]
+  [string]$InstallerPath,
+
+  [Parameter(Mandatory = $true)]
+  [string]$SignaturePath,
+
+  [string]$PrivateKey = $env:WINSPARKLE_DSA_PRIVATE_KEY,
+
+  [string]$PublicKeyPath = "windows\dsa_pub.pem",
```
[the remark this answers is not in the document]
atavism: This is a valid long-term improvement, but the current auto_updater_windows dependency bundles WinSparkle 0.8.1, which does not support EdDSA. EdDSA requires upgrading the bundled WinSparkle runtime to 0.9.0 or newer first. We should keep DSA in this fix for compatibility with existing installations, then migrate in a follow-up by upgrading WinSparkle and publishing both DSA and EdDSA signatures during the transition.

.github/workflows/build-macos.yml
```diff
@@ -32,6 +34,31 @@ on:
         required: false
         type: string
         default: macos-15
+      flutter_target:
+        description: "Optional Flutter entrypoint used to build the app"
+        required: false
+        type: string
+        default: ""
+      flutter_build_mode:
```
jigar-f, line 42: Just confirming: do we need profile mode for any test-related thing?

integration_test/auto_update/auto_update_robot.dart
```diff
@@ -0,0 +1,99 @@
...
+          'Visible keys: ${app.visibleKeys().join(', ')}',
+    );
+    await tester.ensureVisible(checkForUpdates);
+    await _captureScreenshot('auto-update-before');
+
+    e2eLog('Triggering Check for Updates');
+    await app.tap(checkForUpdates, name: 'Check for Updates', settle: false);
+  }
+
+  Future<void> _captureScreenshot(String name) async {
```
jigar-f: Since we are using this lot, we should create _captureScreenshot inside  `app_robot ` and reuse it across multiple tests.

test_driver/integration_test.dart
```diff
@@ -0,0 +1,3 @@
+import 'package:integration_test/integration_test_driver.dart';
+
+Future<void> main() => integrationDriver();
```
jigar-f, line 5: Not sure about this. Do we need drivers? This is being deprecated

jigar-f (commented):
Few comments,

myleshorton (approved):

PR #8975, opened 2026-08-12
Bump radiance for early start memory monitor changes

garmr-ulfr:
Pulls in the changes to the memory monitor lifecycle on mobile https://github.com/getlantern/radiance/pull/597. The memory monitor now starts almost immediately when starting the tunnel in order to also monitor usage during initialization.

## Summary by CodeRabbit

* **Chores**
  * Updated an internal dependency to a newer version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ replace github.com/quic-go/qpack => github.com/quic-go/qpack v0.5.1
 require (
 	github.com/alecthomas/assert/v2 v2.3.0
 	github.com/getlantern/lantern-server-provisioner v0.0.0-20251031121934-8ea031fccfa9
-	github.com/getlantern/radiance v0.0.0-20260812191437-dd149e2462a6
+	github.com/getlantern/radiance v0.0.0-20260812194443-f0277e466143
 	github.com/sagernet/sing-box v1.12.22
 	golang.org/x/mobile v0.0.0-20250711185624-d5bb5ecc55c0
 	golang.org/x/sys v0.45.0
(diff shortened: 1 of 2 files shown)


myleshorton:
Device-verified this pin (`f0277e466143`) on an iPhone 16 / iOS 26.6 — posting the numbers here since I built and tested the identical change in parallel before noticing this had already merged.

Release build from the pin (no local `replace`), per-second memory sampling from #8971:

| build | radiance | peak | steady | teardown | result |
|---|---|---|---|---|---|
| 9.1.18 (baseline) | — | — | 32–34 MB | — | holds |
| #8974 | `dd149e2` | 35.92 MB | 33.81 MB | 37 ms | holds |
| **this (#8975)** | `f0277e4` | 37.94 MB | **32.77 MB** | 45 ms | **holds** |

54 samples, no jetsam, 17.2 MB headroom at plateau — the lowest steady-state of the three. Connect and user-initiated disconnect both clean:

```
PacketTunnelProvider stopping, reason: 1
(lantern-tunnel) stopping, reason: NEProviderStopReason(rawValue: 1)
(lantern-tunnel) stopTunnel completed in 0.045 seconds
```

For contrast, 9.1.19/9.1.20 reached the cap in under a second:

```
kernel: memorystatus: Tunnel exceeded mem limit: ActiveHard 50 MB (fatal)
ReportSystemMemory: killed by jetsam reason per-process-limit
```

Worth noting the startup-window point in radiance#597 matches what the captures showed: the extension was killed roughly a second after connecting, before the monitor's first tick, which is why memmon logged nothing on the failing builds and initially looked broken.

One thing I could not confirm from the device console: whether memmon now actually ticks *during* bring-up. Its ticks go to `lantern.log` on disk rather than os_log, so it would take a log bundle from the app to see them. The memory outcome above is unaffected either way.

Peak varies run to run (35.9 / 37.9 / 39.9 across builds) with which servers get probed; the plateau is the stable figure.

Issue #597, opened 2013-03-15
pacUrl is set/unset incorrectly


lanterndev:
Currently /connectivity/pacUrl gets unset when /settings/systemProxy is false, and set when it's true. This is the opposite of when the UI needs to display it. Instead, it should be set once, as soon as the backend knows what it should be, and never needs to be unset thereafter.

(Also, it's still getting a random number tacked onto the end (e.g. proxy_on.pac-1374763195) that it doesn't need.)

Issue #596, opened 2013-03-14
Clicking "Set as system proxy" checkbox flashes proxy url


leahxschmidt:
Uncaught Error: 10 $digest() iterations reached. Aborting!
Watchers fired in the last 5 iterations: [["fn: function (){ return element[0].scrollHeight; }; newVal: 65; oldVal: 60","fn: function (){ return element[0].scrollHeight; }; newVal: 119; oldVal: 114"],["fn: function (){ return element[0].scrollHeight; }; newVal: 70; oldVal: 65","fn: function (){ return element[0].scrollHeight; }; newVal: 124; oldVal: 119"],["fn: function (){ return element[0].scrollHeight; }; newVal: 75; oldVal: 70","fn: function (){ return element[0].scrollHeight; }; newVal: 126; oldVal: 124"],["fn: function (){ return element[0].scrollHeight; }; newVal: 76; oldVal: 75","fn: function (){ return element[0].scrollHeight; }; newVal: 130; oldVal: 126"],["fn: function (){ return element[0].scrollHeight; }; newVal: 79; oldVal: 76","fn: function (){ return element[0].scrollHeight; }; newVal: 133; oldVal: 130"]]

PR #8976, opened 2026-08-12
deps: bump radiance for early memory-monitor start

myleshorton:
Picks up getlantern/radiance#597, which starts a single memory monitor at the top of tunnel init rather than after libbox bring-up, so the allocation-heavy setup window is no longer unmonitored. It also pins GOMEMLIMIT before libbox setup on iOS, guards `SetExecutor` against a nil executor, and runs the monitor on `t.ctx` so `close()` stops it deterministically.

That window is where the iOS jetsam kills lived: the extension was killed roughly a second after connecting, **before the monitor's first tick**, so nothing observed the burst — which is why memmon looked dead on the failing builds. #597 makes that window observable. The fix for the burst itself was radiance#596, bumped in #8974.

## Device verification

iPhone 16 / iOS 26.6, release build from this pin (no local `replace`), per-second sampling:

| build | radiance | peak | steady | teardown | result |
|---|---|---|---|---|---|
| 9.1.18 (baseline) | — | — | 32–34 MB | — | holds |
| #8974 | `dd149e2` | 35.92 MB | 33.81 MB | 37 ms | holds |
| **this PR** | `f0277e4` | 37.94 MB | **32.77 MB** | 45 ms | **holds** |

54 samples, no jetsam, 17.2 MB headroom at plateau. Connect and a user-initiated disconnect both clean:

```
PacketTunnelProvider stopping, reason: 1
(lantern-tunnel) stopping, reason: NEProviderStopReason(rawValue: 1)
(lantern-tunnel) stopTunnel completed in 0.045 seconds
```

Peak varies run to run (35.9 / 37.9 / 39.9 across builds) with which servers get probed; the plateau is the stable figure and this is the lowest measured so far.

For contrast, the builds this fixes went to the 50 MB cap in under a second:

```
kernel: memorystatus: Tunnel exceeded mem limit: ActiveHard 50 MB (fatal)
ReportSystemMemory: killed by jetsam reason per-process-limit
```

## Scope

`go.mod` + `go.sum` only, one module line each. `go build ./...` clean. Verified the pinned module carries both fixes: `kindling/smart/client.go` has the #596 guard and `vpn/memmon.go` has #597's `SetExecutor`.

Not directly confirmed: that memmon now ticks *during* bring-up. Its ticks go to `lantern.log` on disk rather than os_log, so the device console cannot show them — it would need a log bundle from the app. The memory outcome above is unaffected either way.

https://claude.ai/code/session_01SkKS8hGeHM5g4BDPvxcamg

## Summary by CodeRabbit

* **Chores**
  * Updated an internal dependency to a newer version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ replace github.com/quic-go/qpack => github.com/quic-go/qpack v0.5.1
 require (
 	github.com/alecthomas/assert/v2 v2.3.0
 	github.com/getlantern/lantern-server-provisioner v0.0.0-20251031121934-8ea031fccfa9
-	github.com/getlantern/radiance v0.0.0-20260812191437-dd149e2462a6
+	github.com/getlantern/radiance v0.0.0-20260812194443-f0277e466143
 	github.com/sagernet/sing-box v1.12.22
 	golang.org/x/mobile v0.0.0-20250711185624-d5bb5ecc55c0
 	golang.org/x/sys v0.45.0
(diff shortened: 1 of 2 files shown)


myleshorton:
Closing as a duplicate — #8975 landed the identical pin (`f0277e466143`) at 20:00 UTC while this was building. `git diff origin/main..HEAD` is empty, so there is nothing here to merge.

Device verification has been posted on #8975 instead, since that is the change that shipped.

PR #8970, opened 2026-08-12
Split tunneling UI changes

jigar-f:
This pull request introduces browser detection and new user warnings for bypassing browsers in split tunneling. The main changes add the ability to identify browsers across Android, Windows, and core data models, and surface appropriate warnings to users when they attempt to bypass browsers or all apps. Additionally, one-time explainer dialogs are implemented for bypassing apps or websites, with persistent tracking to avoid repeat prompts.

**Browser detection and flag propagation:**

[...]

PR #8978, opened 2026-08-12
Fix stale browser flag, empty-input clear, and dropped await in split tunneling

atavism:
Follow-up fixes on top of #8970 found during review:

- `browsers_windows.go`: `markBrowsers` returned early before resetting
  `IsBrowser` whenever the registry scan came back empty (permission
  issues, locked-down account, etc.), so a browser flagged `true` from an
  older cache stayed flagged forever, even after uninstall. Removed the
  early return so the reset always runs.
- `website_domain_input.dart`: `addValidatedWebsites` cleared the text
  field unconditionally, even when nothing was actually added (e.g.
  separator-only input like ","). Now it's a no-op when the added list
  is empty.
- `apps_split_tunneling.dart`: the "Add All Except Browsers" / "Add All
  Anyway" buttons on the select-all warning called
  `notifier.selectApps(...)` without awaiting it, so a failure there
  (e.g. FFI command isolate not ready) became an unhandled Future
  rejection instead of surfacing. Now properly awaited.

### DIFF
--- a/lantern-core/apps/browsers_windows.go
+++ b/lantern-core/apps/browsers_windows.go
@@ -171,9 +171,6 @@ func browserCommandExe(cmd string) string {
 // but not byte-identical path.
 func markBrowsers(list []*AppData) {
 	idx := loadBrowserIndexOnce()
-	if len(idx.paths) == 0 && len(idx.basenames) == 0 {
-		return
-	}
 	for _, app := range list {
 		if app == nil {
 			continue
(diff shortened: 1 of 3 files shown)


jigar-f (approved):
LGTM! One small process request: for minor changes like this, could you leave inline review comments instead of opening a separate PR? I'm happy to make the changes myself — it keeps the history cleaner and the discussion in one place.

PR #9005, opened 2026-08-21
Add user messaging to Lantern

atavism:
For getlantern/engineering#3784

## Summary by CodeRabbit

* **New Features**
  * Added accessible in-app user messages displayed as snackbars.
  * Added expiration handling, secure HTTPS links, and plans navigation.
  * Added message refresh and acknowledgment capabilities.
  * Prevented duplicate displays and interruptions during onboarding, payments, dialogs, and blocked screens.
  * Added automatic dismissal and clear closing controls.
  * App version information is now provided to supporting services.

* **Tests**
  * Added comprehensive coverage for message parsing, actions, routing, display behavior, expiration, and acknowledgment.


lib/lantern_app.dart
```diff
@@ -23,6 +27,10 @@ import 'features/system_tray/system_tray_wrapper.dart';
 
 final globalRouter = sl<AppRouter>();
 final RouteObserver<PageRoute> routeObserver = RouteObserver<PageRoute>();
+final UserMessageRouteObserver userMessageRouteObserver =
+    UserMessageRouteObserver(
+      blockedRouteNames: {Onboarding.name, ChoosePaymentMethod.name},
+    );
```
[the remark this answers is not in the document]
atavism: Verified this against the current import graph. `lantern_app.dart` imports `core/common/common.dart`, which explicitly re-exports `router.gr.dart`, so both generated route names resolve. Adding the direct import produces an `unnecessary_import` analyzer finding, so I left the working import structure unchanged.

lib/features/user_message/user_message_host.dart
```diff
@@ -0,0 +1,281 @@
...
+
+  @override
+  void dispose() {
+    WidgetsBinding.instance.removeObserver(this);
+    widget.routeObserver.changes.removeListener(_scheduleAttempt);
+    _retryTimer?.cancel();
+    final presentation = _presentation;
+    presentation?.expirationTimer?.cancel();
+    super.dispose();
+  }
```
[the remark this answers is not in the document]
atavism: Fixed in 31cbbc075. The host now cancels and detaches its presentation when it leaves the tree, removes the snackbar outside the locked widget lifecycle, and releases an unshown claim through a retained controller reference. An active-state guard also prevents queued callbacks from presenting after deactivation, and a remount regression test covers the stuck-claim case.

lib/features/user_message/user_message_host.dart
```diff
@@ -0,0 +1,281 @@
...
+        if (!presentation.visible) {
+          ref
+              .read(userMessageControllerProvider.notifier)
+              .releaseClaim(message.displayId, _now);
+        }
+        _presentation = null;
+        _scheduleAttempt();
+      }),
+    );
+  }
```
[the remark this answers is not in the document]
atavism: Fixed in 31cbbc075. The `closed` cleanup is registered before the second expiration check, so the presentation and claim are cleared even when the clock crosses expiration immediately after the claim. Added a regression test that forces this exact race and verifies the next eligible message can still display.

test/features/user_message/user_message_host_test.dart
```diff
@@ -0,0 +1,340 @@
...
+        ),
+      );
+      await _pumpToSnackbar(tester);
+
+      final bodyFinder = find.byKey(UserMessageHost.bodyKey);
+      expect(bodyFinder, findsOneWidget);
+      expect(Directionality.of(tester.element(bodyFinder)), TextDirection.rtl);
+      expect(find.bySemanticsLabel(body), findsOneWidget);
+      expect(tester.takeException(), isNull);
+      semantics.dispose();
```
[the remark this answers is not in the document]
atavism: Fixed in 31cbbc075. The semantics handle is now protected by `try/finally`, which guarantees disposal if an assertion throws while also closing it before Flutter performs its end-of-test semantics verification.

lib/features/user_message/user_message_host.dart
```diff
@@ -0,0 +1,302 @@
...
+    presentation.visible = true;
+    presentation.expirationTimer?.cancel();
+    unawaited(_controller.markPresented(presentation.message.displayId));
+  }
+
+  void _expireBeforePresentation(_UserMessagePresentation presentation) {
+    if (!mounted || !identical(_presentation, presentation)) return;
+    if (presentation.visible) return;
+    presentation.messenger.removeCurrentSnackBar();
+  }
```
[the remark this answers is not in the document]
atavism: Fixed in a9e33423d. User messages now render on a host-owned presentation surface instead of entering the application ScaffoldMessenger queue. Expiration, unsafe-state dismissal, host deactivation, CTA, close, and timeout therefore clear only the owned user-message surface. Added regressions proving an existing application snackbar survives both host deactivation and pre-display message expiration.

jigar-f:
@atavism This is on my radar; I am planning to review this tomorrow.

lib/features/user_message/user_message_action_dispatcher.dart
```diff
@@ -0,0 +1,48 @@
...
+  }) : _openHttpsUrl = openHttpsUrl,
+       _openPlans = openPlans;
+
+  factory UserMessageActionDispatcher.application(StackRouter router) {
+    return UserMessageActionDispatcher(
+      openHttpsUrl: (uri) async {
+        await launchUrl(uri, mode: LaunchMode.externalApplication);
+      },
+      openPlans: () async {
+        await router.push(Plans());
```
jigar-f: I think we don't need to pass the router here; you can use `appRouter.push(Plans());`. It will work just as this.

lib/features/user_message/user_message_host.dart
```diff
@@ -0,0 +1,275 @@
...
+      _dismissForUnsafeState();
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    ref.watch(userMessageControllerProvider);
+    _scheduleAttempt();
+    final presentation = _presentation;
+    return Stack(
```
jigar-f, line 101: @atavism Can't we avoid the whole overlay and use Flutter's built‑in `ScaffoldMessenger.showSnackBar`? That would simplify things a lot. We wouldn't have to manage attempts and everything.

lib/features/user_message/user_message_host.dart
```diff
@@ -0,0 +1,275 @@
...
+      final presentation = _presentation;
+      if (presentation != null) {
+        if (!_isSafeToPresent()) _dismissForUnsafeState();
+        return;
+      }
+      _attemptPresentation();
+    });
+  }
+
+  bool _isSafeToPresent() {
```
jigar-f, line 125: Same for this as well. We don't have to manage this either.

lib/features/user_message/user_message_repository.dart
jigar-f, line 1: Since we are not using the repository pattern across the codebase, I think we should keep it consistent and convert this into a provider with all these methods. 

If you want to keep that, fine. Later, we should also start migrating other features to this.

lib/features/user_message/user_message_host.dart
atavism: I kept this separate so campaign expiry or dismissal can’t affect other app snackbars. We still need the background and dialog checks either way.

jigar-f: Got it.

ios/Runner/Handlers/MethodHandler.swift
```diff
@@ -252,6 +252,24 @@ class MethodHandler {
...
+      case "acknowledgeUserMessage":
+        guard let data = self.decodeDict(from: call.arguments, result: result),
+          let displayID: String = self.decodeValue(from: data["displayId"], result: result),
+          let accountID: String = self.decodeValue(from: data["accountId"], result: result)
+        else { return }
+        self.acknowledgeUserMessage(result: result, displayID: displayID, accountID: accountID)
+
+      case "setUserMessageActivity":
+        guard let active: Bool = self.decodeValue(from: call.arguments, result: result)
+        else { return }
```
jigar-f, line 270: nit: I think we should add a log here, just to avoid silent failure.

lib/features/user_message/user_message_action_dispatcher.dart
```diff
@@ -0,0 +1,42 @@
...
+typedef OpenUserMessagePlans = Future<void> Function();
+
+class UserMessageActionDispatcher {
+  const UserMessageActionDispatcher({
+    required OpenUserMessageURL openHttpsUrl,
+    required OpenUserMessagePlans openPlans,
+  }) : _openHttpsUrl = openHttpsUrl,
+       _openPlans = openPlans;
+
+  factory UserMessageActionDispatcher.application() {
```
jigar-f, line 16: Nit: Now we don't need this method; you can move this directly to dispatch since that is context-independent now 

```dart
 Future<void> dispatch(UserMessageAction action) async {
    switch (action.type) {
      case UserMessageActionType.openHttpsUrl:
        final uri = action.url;
        if (uri == null || !UserMessageAction.isAllowedHttpsUrl(uri)) return;
         await launchUrl(uri, mode: LaunchMode.externalApplication);

        return;
      case UserMessageActionType.openPlans:
        await appRouter.push(Plans());
        return;
    }
  }
```

lib/features/user_message/user_message_controller.dart
```diff
@@ -0,0 +1,222 @@
...
+      pending: identical(pending, _unchangedPending)
+          ? this.pending
+          : pending as UserMessage?,
+      presentationClaimed: presentationClaimed ?? this.presentationClaimed,
+      displayedThisSession: displayedThisSession ?? this.displayedThisSession,
+    );
+  }
+}
+
+class UserMessageController extends Notifier<UserMessageState> {
```
jigar-f, line 40: Thanks for converting to provider; now we can use autogen like we do for providers

```dart
@Riverpod(keepAlive: true)
class UserMessageController extends  _$UserMessageController {
```

jigar-f (approved):
A few more comments, otherwise, LGTM.

PR #9018, opened 2026-08-27
Fix auth smoke workflow validation and protect CI credentials

atavism:
Fixes invalid GitHub Actions secret checks that prevented auth smoke workflows from starting. Also redacts password values from auth test failure diagnostics


jigar-f (approved):
LGTM

PR #9023, opened 2026-09-02
report_issue: submit issue-type keys, not translated labels

jay-418:
## context

- closes https://github.com/getlantern/engineering/issues/3839

## summary

The report-issue dropdown submitted the translated display label as its value, but [parseIssueType](https://github.com/getlantern/lantern/blob/2855b0da7/lantern-core/core.go#L947) matches untranslated keys, so every pick except English "Slow" filed as "other" (97% of 9.1.x customer tickets).

- [issueOptions](https://github.com/getlantern/lantern/blob/2855b0da7/lib/features/report_issue/report_issue.dart#L36-L46) now holds raw keys; the [dropdown entry](https://github.com/getlantern/lantern/blob/2855b0da7/lib/features/report_issue/report_issue.dart#L534) translates only the label, so drafts and deep links inherit the key
- [parseIssueType](https://github.com/getlantern/lantern/blob/2855b0da7/lantern-core/core.go#L960) accepts cannot_link_devices (the locale msgid is plural in all 22 .po files) and [warns](https://github.com/getlantern/lantern/blob/2855b0da7/lantern-core/core.go#L968) on unrecognized values so a future mismatch is visible in logs
- new [TestIssueOptionsMatchParser](https://github.com/getlantern/lantern/blob/2855b0da7/lantern-core/core_test.go#L44) reads issueOptions from the Dart source and fails if values are translated, unrecognized by the parser, or missing from en.po
- [go.yml triggers](https://github.com/getlantern/lantern/blob/2855b0da7/.github/workflows/go.yml#L6-L24) now include report_issue.dart and en.po so the contract test runs for Dart-only changes

## verify

- [ ] submit an in-app report choosing "Application crashes" and confirm the ticket arrives in Freshdesk typed "crashes", not "other"

Ticket volume typed "other" drops gradually, not immediately: 9.1.x clients in the field keep sending translated labels until they upgrade.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved report issue handling so selected issue types are submitted correctly.
  - Preserved localized labels in the issue-type menu while ensuring the underlying selection remains consistent.
  - Added support for both singular and plural device-link issue descriptions.
  - Unrecognized issue types now fall back safely instead of causing submission errors.

- **Tests**
  - Added coverage to verify report issue options, translations, and submission behavior remain aligned.


jay-418:
Hey @atavism, @jigar-f,

I picked this up because I assumed it was a problem with our wire protocol, but it looks like the mismatch was on the app side (using translated values instead of English keys to select the wire option).

- Am I on the right track here?
- Does the regression test make any sense, or is there a better way to do this?

jigar-f (changes requested):
@jay-418, sorry for the delay. LGTM

jigar-f (approved):
I meant to approve it.

PR #9028, opened 2026-09-04
Fix billing routing for Play builds

atavism:
Resolves https://github.com/getlantern/engineering/issues/3885

Keep the Play AAB on the store purchase path regardless of country, installer detection, etc. Preserve sideload payment behavior and avoid initializing unavailable Play Billing in censored regions.

## Summary by CodeRabbit

- **New Features**
  - Improved detection of official Play Store builds.
  - Android users can use API-provided pricing when Play Billing is unavailable.
  - Restore purchases is shown only when store billing is supported.
  - Payment options now adapt to country and store availability.

- **Bug Fixes**
  - Play Billing availability now considers platform, store-build status, country detection, and reported region.
  - Purchase setup no longer waits indefinitely for country information.
  - Failed product loading no longer redirects automatically to Stripe signup or changes region status.
  - Improved loading and navigation behavior when opening the plans experience.


myleshorton:
Pushed two follow-up fixes (292412df8) found while reviewing:

**1. Russia had no purchase path at all.** Google's notice on engineering#3885 exempts **India and Russia** from the removal, and Play Billing is unreachable in RU — so routing Play builds there through the store left those users unable to pay, for a policy that does not apply to them. CN and IR stay on the store path as intended.

Added `CountryCode.externalPaymentRegions = ["RU"]` and a `useStorePurchaseFlow()` predicate, which separates three questions that were partly collapsed:

| predicate | question | used for |
|---|---|---|
| `isStoreVersion()` | build identity — distributed via a store? | telemetry, update channel |
| `useStorePurchaseFlow()` | routing — which payment path do we present? | purchase/signup routing |
| `canUseStoreBilling()` | capability — is billing reachable now? | restore link, SKU fetch, client init |

India is deliberately not listed: Play Billing works there, so there is nothing to fall back from.

All payment-routing call sites moved to the new predicate. **Two were load-bearing and would have made the carve-out silently not work:**

- `choose_payment_method.dart:174` — `isAndroidSideload` gated the Stripe branch on build identity. A diverted RU user hits `isStoreVersion() == true`, so the `stripe` case matched no branch and fell through to `break` — tapping a payment method would have done nothing.
- `lantern_platform_service.dart:847` — `plans()` sends the channel to core, which decides **which providers come back at all**. Left as `store`, RU users would have been offered Google Play regardless of the client-side routing.

**2. Unknown country hard-blocked purchases everywhere, not just censored regions.** `canUsePlayBilling()` now requires `CountryCode.isKnown`, but the country only arrives on a config update (`lantern-core/core.go:393`) and nothing seeds it at startup. A cold-start tap that missed the 3s `waitUntilKnown` window failed with "Unable to access in-app purchases" — including in markets where Play Billing works fine. On `main` these users purchased normally, so this was a new regression on the revenue path.

User-initiated purchase and restore now wait 10s (the loading dialog is already up, so the wait is free); the background product prefetch keeps the 3s default.

Tests: 4 cases for `resolveStorePurchaseFlow`, plus an invariant test that every external-payment region is also a censored region — a non-censored entry there would silently divert a working market to Stripe. `flutter test` 202 passing, `flutter analyze` unchanged at 33 pre-existing issues.

https://claude.ai/code/session_018PPFajzCWzcS3vcPDC5kes

myleshorton:
Pulled the Russia carve-out in 7f6a06482, per Adam: Russia produces effectively no revenue, so the exemption bought nothing while leaving an external payment path in a build whose whole purpose is showing Play review there is not one. **CN, RU and IR all stay on the store path**, as originally written.

What remains from my earlier commit is only the country-code wait fix, which is independent of the carve-out:

`canUsePlayBilling()` requires `CountryCode.isKnown`, but the country only arrives on a config update (`lantern-core/core.go:393`) and nothing seeds it at startup. A cold-start tap that missed the 3s `waitUntilKnown` window failed with "Unable to access in-app purchases" — **including in markets where Play Billing works fine**. On `main` those users purchased normally, so it was a new regression on the revenue path. User-initiated purchase and restore now wait 10s (the loading dialog is already up); the background product prefetch keeps the 3s default.

Net diff vs. fa8517767 is now `lib/core/services/app_purchase.dart` alone, +12/-1.

`flutter test` 196 passing, `flutter analyze` unchanged at 33 pre-existing issues.

For the record, since it may come up again: the exemption reading was that Google's notice on engineering#3885 scopes removal to "outside India and Russia." If RU revenue ever becomes non-trivial, the carve-out is recoverable from 292412df8 — but it needs the routing/identity split that commit introduced, because several call sites (notably `plans()`, which tells core which providers to return) key off `isStoreVersion()` and would otherwise still serve Google Play.

https://claude.ai/code/session_018PPFajzCWzcS3vcPDC5kes

PR #9037, opened 2026-09-09
tighten checkout confirmation

atavism:
Wait for account data before starting checkout so existing Pro access doesn’t falsely confirm a renewal.


jigar-f (approved):
LGTM

PR #9040, opened 2026-09-10
Update underlying network selection on Android

atavism:
Fixes cases where Android 9–11 selects the VPN as its own underlying network, and keeps startup and server switches from overwriting the network listener’s selection. Also clarifies network request capabilities and cleanup.


android/app/src/main/kotlin/org/getlantern/lantern/service/DefaultNetworkMonitor.kt
```diff
@@ -39,16 +41,18 @@ object DefaultNetworkMonitor {
     }
 
     suspend fun start() {
+        // During a server switch the active network may be our own VPN. Only
+        // seed a known non-VPN network, and do it before subscribing so a cached
+        // network or a fresh callback from the listener always takes precedence.
+        defaultNetwork = LanternApp.connectivity.activeNetwork?.takeIf { network ->
```
jigar-f: @atavism  Adding this break VPN, not sure why; I am reverting it for now.

PR #9051, opened 2026-09-14
Fix Unbounded globe drag gestures on iOS and Android

myleshorton:
Dragging the Unbounded globe on iOS and Android could scroll the panel or swipe tabs without rotating the globe. Its scale recognizer waits for twice the touch movement required by the surrounding drag recognizers, so the ancestors win the gesture first.

Give only the globe subtree half the inherited touch slop. Its scale recognizer then reaches the same movement threshold as the parent drag recognizers and, as the child, wins the gesture. Existing globe rotation and momentum callbacks handle the interaction. Automatic rotation policy is unchanged.

```mermaid
sequenceDiagram
    autonumber
    participant App as Globe host
    participant Globe as Globe recognizer
    participant Parent as Scroll and tabs
    Note over Globe,Parent: rotating_globe.dart:1733<br/>Scale recognizer competes with parent drags
    rect rgba(255, 200, 200, 0.3)
        Note over Parent: action_mode_widgets.dart:28 and home.dart:305<br/>🐛 Parent drags win before globe reaches its default threshold
    end
    App->>Globe: share_my_connection.dart:1425<br/>⚠️ Scope half the inherited touch slop to globe
    Note over Globe: share_my_connection.dart:1432<br/>Globe now claims drag and runs existing rotation callbacks
```

`rotating_globe.dart` refers to pinned flutter_earth_globe 2.2.1; other line references refer to this PR branch.

Validation: all 300 Flutter tests pass, targeted Dart analysis passes, and git diff --check passes. New iOS and Android widget tests exercise horizontal and vertical touch drags inside both a scroll view and tab pager, verify rotation changes while parent positions stay fixed, and all four fail against the original implementation. Physical iPhone validation remains outstanding.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved globe gesture handling so horizontal and vertical drags consistently rotate the globe instead of scrolling surrounding content or switching tabs.
  * Reduced the gesture threshold for the globe to make touch interactions more responsive.

* **Tests**
  * Added coverage verifying globe interactions take precedence within scrollable and tabbed layouts.
  * Expanded gesture testing across both iOS and Android platforms.


myleshorton (commented):
I traced the mechanism through Flutter 3.41.5 and ran the tests both with and without the production change. **The fix is correct and the tests are genuinely load-bearing** — reverting only the `gestureSettings` block fails both new tests (`+0 -2`), so this isn't a decorative regression test. Nice.

Confirming the chain for anyone reviewing after me, since it hops through three packages:

`FlutterEarthGlobe` → `RotatingGlobe` → `InteractiveViewer` → `GestureDetector(onScaleStart/Update/End)` → `ScaleGestureRecognizer`. `GestureDetector` (unlike `RawGestureDetector`) *does* forward `MediaQuery.maybeGestureSettingsOf(context)` onto its recognizers, so scoping the override here really does reach the globe's recognizer.

| Recognizer | Threshold | Before | After |
|---|---|---|---|
| Globe `ScaleGestureRecognizer` | `computePanSlop` = `touchSlop * 2` | 36 | **18** |
| `Scrollable` / `TabBarView` drag | `computeHitSlop` = `touchSlop` | 18 | 18 |

Worth noting the blast radius is well contained: `TapGestureRecognizer` uses the `kTouchSlop` *constant* for `preAcceptSlopTolerance`, not `gestureSettings`, so halving the slop does **not** make the globe's `onTapDown` (peer-point taps) twitchier. And `computeScaleSlop` ignores `gestureSettings` entirely, so pinch-zoom is untouched. Both good outcomes that the PR description doesn't claim but which I'd have worried about.

Three things below — one substantive, two housekeeping.

---

### 1. This makes ~60% of the panel a vertical-scroll dead zone (the substantive one)

The new test asserts `expect(scroll.position.pixels, scrollBefore)` for a *vertical* drag — i.e. it deliberately locks in "dragging down on the globe no longer scrolls the panel." That's a real product decision and I don't think the description surfaces it.

I measured the actual footprint by pumping `UnboundedTab` and printing the geometry:

| Viewport | Text scale | Globe rect | Globe % of viewport | `maxScrollExtent` |
|---|---|---|---|---|
| 393×852 | 1.0 | `LTRB(16, 80, 377, 596)` | **60.6%** | 146 px |
| 360×640 | 1.0 | `LTRB(16, 96, 344, 400)` | 47.5% | 162 px |
| 393×852 | **2.0** | `LTRB(16, 224, 377, 740)` | **60.6%** | **1106 px** |

The last row is the one that bothers me. At 2× accessibility text there is over a full viewport of content below the fold (1106 px), and once the globe is on screen it eats 224→740 of 852 — leaving a 224 px strip above and a 112 px strip below to scroll from. That's precisely the case the surrounding widget exists for:

> `action_mode_widgets.dart:7` — *"Scrolls on small screens and with large accessibility text instead of squeezing the globe or overflowing the controls."*

So the fix is strongest against the symptom that's unambiguously a bug (horizontal → tab swipe) and has a real cost on the one that's arguable (vertical → scroll).

Options, roughly in order of how much I'd lean toward them:

- **Accept it, but say so.** Embedded maps behave this way, and the globe does rotate on both axes so vertical tilt is a genuine feature. If that's the call, a line in the code comment saying vertical scroll is *intentionally* surrendered inside the globe's box would stop the next person from filing it as a bug.
- **Claim horizontal only**, letting vertical drags fall through to the scroll view. Preserves the scroll gutter and still kills the tab-swipe bug, at the cost of vertical tilt.
- **Cap the globe's height** so it never exceeds ~50% of the viewport, keeping a usable scroll gutter at large text sizes.

Happy either way — flagging it because the tradeoff is now pinned by an assertion, and un-pinning it later means arguing with a test.

### 2. Not actually iOS-specific

I copied the tests and swapped the variant to `TargetPlatform.android`: they **fail without the fix and pass with it**, same as iOS. That makes sense — nothing in the arena math is platform-conditional. `computePanSlop` is `touchSlop * 2` and `computeHitSlop` is `touchSlop` on *every* touch platform, so the 2:1 ratio (and therefore the bug) is identical on Android; only the base value differs (null → `kTouchSlop` 18 on iOS, the platform value from `DeviceGestureSettings.fromView` on Android).

Android users have this bug today and this PR fixes it for them too — but nothing guards the regression there. See the inline note on the `variant:` line.

### 3. Minor

See inline comments on the test file. Also agreeing with Copilot's import-ordering nit, though for a different reason than it gave — details inline.

---

**Reproduction notes**, in case you want to re-run any of this: `git worktree` on the PR head, `flutter pub get`, copy `app.env` in (it's gitignored and the asset bundle won't build without it), then `flutter test test/features/share_my_connection/action_mode_widgets_test.dart --plain-name "globe owns touch drag"`. Reverting just the `copyWith` block back to the one-liner is enough to see both tests go red.

lib/features/share_my_connection/share_my_connection.dart
```diff
@@ -1415,7 +1416,18 @@ class _GlobeViewState extends ConsumerState<_GlobeView> {
...
+                    size: widgetSize,
+                    // Scale gestures use twice the touch slop of page drags.
+                    // Match the parent's drag threshold so the child globe
+                    // wins before the scroll view or tab pager takes the touch.
+                    gestureSettings: DeviceGestureSettings(
+                      touchSlop:
+                          (MediaQuery.gestureSettingsOf(context).touchSlop ??
+                              kTouchSlop) /
+                          2,
+                    ),
```
myleshorton: The mechanism is right but the comment slightly misdescribes it, and the distinction matters for anyone tuning this later.

After the change the globe's `panSlop` is **exactly equal** to the parent's `hitSlop` (both 18 on iOS), so the globe doesn't win *before* the parent — for a pure-axis drag both recognizers cross their threshold on the *same* pointer event. The globe wins the tie because hit testing registers the deeper recognizer first, so its `resolve(accepted)` reaches the closed arena first and rejects the others.

That's actually the ideal value rather than a lucky one: `panSlop == hitSlop` is the *largest* slop at which the globe can never cross later than the parent, so it wins on every sample rate without being any twitchier than it has to be. Worth stating, because it reads like an arbitrary `/2`.

I probed the sensitivity with 1 px drag steps to see how much margin there is:

| divisor | resulting `panSlop` | result |
|---|---|---|
| 1.5 | 24.0 | ❌ parent wins |
| 1.9 | 18.95 | ✅ |
| **2** | **18.0** | ✅ |
| 2.5 | 14.4 | ✅ |

(1.9 passes only because 1 px sampling lets both land in the same step; at 2 the guarantee is exact rather than incidental.)

Suggested rewording:

```dart
// A ScaleGestureRecognizer accepts at DeviceGestureSettings.panSlop,
// which is touchSlop * 2 — twice what the surrounding scroll view and
// tab pager need, so they always claimed the touch first. Halving the
// slop here makes the globe's panSlop equal the parents' touchSlop, so
// it can never cross later than they do; on the resulting tie the globe
// wins because hit testing registers the deeper recognizer first.
```

One micro-nit while you're in here: `MediaQuery.of(context)` on the line above already establishes a dependency on the full `MediaQueryData`, so `MediaQuery.gestureSettingsOf(context)` can just be `MediaQuery.of(context).gestureSettings` — no behavioural difference, one less lookup to follow.

test/features/share_my_connection/action_mode_widgets_test.dart
```diff
@@ -86,6 +87,52 @@ void main() {
...
+        await tester.pump();
+        expect(
+          delta.dx != 0 ? state.rotationZ : state.rotationX,
+          isNot(delta.dx != 0 ? before.dx : before.dy),
+        );
+        expect(scroll.position.pixels, scrollBefore);
+        expect(tabs.animation!.value, 0);
+        await tester.pumpWidget(const SizedBox());
+      },
+      variant: TargetPlatformVariant.only(TargetPlatform.iOS),
```
myleshorton: I swapped this to `TargetPlatform.android` and ran it both ways: **fails without the fix, passes with it** — identical to iOS.

Nothing in the arena math is platform-conditional (`computePanSlop` is `touchSlop * 2` and `computeHitSlop` is `touchSlop` for every touch-like `PointerDeviceKind`), so the 2:1 ratio that causes the bug holds on Android too. Only the base value differs — null → `kTouchSlop` on iOS vs. the platform value from `DeviceGestureSettings.fromView` on Android — and the ratio is what matters here.

Suggest widening so the Android regression is guarded too:

```suggestion
      variant: const TargetPlatformVariant(<TargetPlatform>{
        TargetPlatform.iOS,
        TargetPlatform.android,
      }),
```

Desktop doesn't need covering: `ScrollBehavior.dragDevices` excludes `PointerDeviceKind.mouse`, so no parent drag recognizer competes for a mouse drag and the globe already wins uncontested.

test/features/share_my_connection/action_mode_widgets_test.dart
```diff
@@ -86,6 +87,52 @@ void main() {
...
+        }
+        await gesture.up();
+        await tester.pump();
+        expect(
+          delta.dx != 0 ? state.rotationZ : state.rotationX,
+          isNot(delta.dx != 0 ? before.dx : before.dy),
+        );
+        expect(scroll.position.pixels, scrollBefore);
+        expect(tabs.animation!.value, 0);
+        await tester.pumpWidget(const SizedBox());
```
myleshorton: I think this is dead. I removed it and ran both the two new tests and the full file (16 tests) — all green, no pending-timer or undisposed-ticker complaints.

If it *was* guarding something (draining the globe's deceleration controller before `tabs.dispose`, say), it needs a comment saying so, because it reads as accidental. Otherwise drop it.

test/features/share_my_connection/action_mode_widgets_test.dart
```diff
@@ -1,3 +1,4 @@
+import 'package:flutter_earth_globe/rotating_globe.dart';
```
myleshorton: Two things on this line.

**Ordering** — agreeing with Copilot that this should sit below `dart:ui`, but its stated reason is wrong: `directives_ordering` is *not* enabled here. `analysis_options.yaml` includes `package:flutter_lints/flutter.yaml` → `package:lints/recommended.yaml`, and that rule is in neither (which is why `flutter analyze` is green). So it's a convention nit, not a lint failure — still worth fixing since every other file in the repo puts `dart:` first, but nobody should go hunting for an analyzer error that isn't there.

**Reaching into package internals** — the more durable concern. `rotating_globe.dart` isn't exported by `package:flutter_earth_globe/flutter_earth_globe.dart`; that library only re-exports `misc.dart` and `satellite.dart`. So `RotatingGlobe` / `RotatingGlobeState` and the `rotationX` / `rotationZ` fields are effectively private API that Dart happens to let us import.

Combined with `flutter_earth_globe: ^2.2.0` in `pubspec.yaml`, a `flutter pub upgrade` to a 2.x minor could rename the file or those fields without any semver signal, and this is the only test that would break. Not blocking — there's no public rotation accessor on `FlutterEarthGlobeController` to use instead (it exposes `startRotation` / `stopRotation` / `isRotating`, but no rotation readout), so this may genuinely be the only hook available.

A one-line comment noting the import is deliberate and package-internal would save the next person the archaeology, e.g.:

```dart
// Not exported by flutter_earth_globe.dart — the package exposes no public
// rotation readout, so the test reads RotatingGlobeState directly. Revisit if
// the pinned 2.2.1 moves.
```

test/features/share_my_connection/action_mode_widgets_test.dart
```diff
@@ -1,3 +1,4 @@
...
 import 'dart:ui' show SemanticsAction;
 import 'package:flutter/material.dart';
```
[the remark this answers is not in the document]
myleshorton: Moved the package import below dart:ui and flutter/material.dart in 1216d5231. This follows repository convention; directives_ordering is not enabled, so the original ordering was not an analyzer failure.

lib/features/share_my_connection/share_my_connection.dart
myleshorton: Clarified the equal-threshold tie and deeper-recognizer ordering in 1216d5231, and used MediaQuery.of(context).gestureSettings consistently. The comment now explains why dividing by two is sufficient without reducing the threshold further.

test/features/share_my_connection/action_mode_widgets_test.dart
myleshorton: Added Android to the test variant in 1216d5231. All four horizontal/vertical iOS/Android cases fail against the original implementation and pass with the fix. All 300 tests pass; the PR title and description now reflect both platforms.

test/features/share_my_connection/action_mode_widgets_test.dart
myleshorton: Removed the explicit empty-widget pump in 1216d5231. The full 300-test suite passes without pending-timer or ticker failures.

test/features/share_my_connection/action_mode_widgets_test.dart
myleshorton: Reordered the import and documented the deliberate internal dependency in 1216d5231: the public controller has no rotation readout, so this test reads RotatingGlobeState and must be revisited on package upgrades. Targeted Dart analysis is clean.

PR #9055, opened 2026-09-15
Fix small-screen settings text and Pro renewal banners

atavism:
Fix two layout issues found while reviewing #9025.

- Let settings labels and values wrap on small screens so the selected routing mode stays readable.
- Use the compact banner for expiring and expired Pro accounts, preserving renewal messages, colors, and navigation to Plans. Full details remain available on long press or hover.
- Add regression coverage


lib/core/widgets/pro_banner.dart
```diff
@@ -141,34 +153,31 @@ class ProBanner extends HookConsumerWidget {
   }
 }
 
-/// One-line 40px pill upsell for small screens (engineering#3046, Figma
-/// node 2854-13197). Expired swaps to the status/error palette; the whole
-/// pill opens Plans.
 class _CompactProBanner extends StatelessWidget {
   const _CompactProBanner({
-    required this.isExpired,
+    required this.isError,
```
jigar-f, line 158: This is not an error state; it's just an error color. The actual state is expired, so I think "isExpired" makes sense here.

jigar-f (approved):
Thanks for adding tests. LGTM, One comment

PR #9056, opened 2026-09-15
Fix Action Mode stop errors and small-screen layout

atavism:
Show an error when Action Mode fails to stop, while keeping the sharing state accurate and allowing another attempt.

- Restore scrolling and a minimum globe height so large text doesn’t cause overflow or overlapping content.
- Let waiting and arrival messages wrap on narrow screens.


lib/features/action_mode/action_mode.dart
```diff
@@ -33,47 +35,42 @@ class ActionModeTab extends HookConsumerWidget {
       return null;
     }, [visible]);
 
-    // The globe fills whatever the cards leave over; the page only scrolls
-    // when the cards alone are taller than the screen.
     return SafeArea(
-      child: CustomScrollView(
```
jigar-f, line 39: @atavism One reason I choose Slivers is their memory efficiency: for unbounded lists, we must redraw every second, but Slivers offer more flexibility and use less memory.

jigar-f (approved):
@atavism  Just one comment, I think that's worth looking into.

lib/features/action_mode/action_mode.dart
atavism: @jigar-f Great point! I opened a PR here to restore the use of Silvers: https://github.com/getlantern/lantern/pull/9063

PR #9061, opened 2026-09-15
ci: fail closed when the Swift planner doesn't succeed

myleshorton:
Targets #9043's branch, since the `plan` job this concerns only exists there — `main` has no planner today.

## The hole

Both native gates hang off `needs: plan` with a plain `if`:

```yaml
  macos-compile-check:
    needs: plan
    if: needs.plan.outputs.macos_needed == 'true'
```

A plain `if` on a needs-output does not run when the dependency fails. So any failure of `plan` leaves both outputs empty and reports **macOS (Runner + PacketTunnel + RunnerTests)** and **iOS (Runner + Tunnel, simulator)** as *skipped* — which reads green — rather than failed.

A Swift break can then land with no native check ever having run. That is precisely what this workflow was added to prevent; its own header says so:

> Compiles macOS and iOS Swift on PRs. The release builds are workflow_call-only, so without this a Swift break is first caught by the nightly (engineering#3850).

Skipping because the planner says "unchanged" is the optimization. Skipping because the planner is *broken* is a hole.

## It isn't hypothetical

The `plan` job has three failure modes, none exotic:

1. **`Fetch PR base tree`** — `git fetch --no-tags --depth=1 origin "$BASE_SHA"` fails when a PR's base branch has been force-pushed and `base.sha` is no longer reachable.
2. **`Test native input selection`** — the planner's own unit tests run inline, so a red test skips the native checks it was meant to protect.
3. **`Compare native inputs and prior successes`** — an uncaught exception in `swift_ci.py`. Reachable today: `find_artifact`'s guard catches `CalledProcessError, TimeoutExpired, KeyError, ValueError` but not `TypeError` (`artifact['workflow_run']` and `run['pull_requests']` can both be `null` in the artifacts API) or `OSError` from `api()`.

That third one is worth noting on its own — it means the intended "fall back to building" path can instead abort the planner, which lands right back in this same skip.

## Fix

```yaml
if: ${{ !cancelled() && (needs.plan.result != 'success' || needs.plan.outputs.macos_needed == 'true') }}
```

| plan | gate |
|---|---|
| succeeded, needed | run |
| succeeded, not needed | **skip** — optimization preserved |
| failed / timed out | **run** — was: skip |
| run cancelled | skip |

`!cancelled()` rather than `always()` deliberately: `always()` would also run these jobs when someone cancels the workflow, which defeats `cancel-in-progress` and burns a macOS runner for 30 minutes on a run nobody wants.

## Verification

- `actionlint .github/workflows/swift-compile-check.yml` passes.
- Applied identically to both gates so macOS and iOS can't drift.

## Not addressed here

Deliberately scoped to the fail-open hole. The `TypeError`/`OSError` gap in `find_artifact`, the success marker not covering the Xcode/runner-image version, manual-dispatch still reusing a cached framework, the repo-wide `GOMOBILE_VERSION` pin, and the `cancel-in-progress` interaction with the new `push: main` trigger are all separate and belong with the split of #9043.


myleshorton:
Superseded by #9062, which merged as `14defc595`.

That PR carries this exact fix — `main` now has both gates fail closed:

```
$ git show origin/main:.github/workflows/swift-compile-check.yml | grep -c "needs.plan.result != 'success'"
2
```

This PR targeted `fisk/swift-ci-input-cache` (#9043's branch) because the `plan` job only existed there. Once #9043 was split, the change-detection half went to `main` with the fail-closed gates built in from the start, so there's nothing left here to land.

Closing as merged-in-substance rather than abandoned — the fix is live.

myleshorton:
Correction to my previous comment: this wasn't closed as superseded — it was actually **merged** (by @atavism on 2026-09-16) into `fisk/swift-ci-input-cache`. I went to close it and GitHub rightly refused.

The substance stands: the fail-closed gates are live on `main` via #9062, so the fix is in regardless of which route it took.

PR #9068, opened 2026-09-17
decouple desktop update checks from core startup

atavism:
Desktop update checks already ran at startup, but could be held up by core initialization or feature-flag requests. Start them independently so clients can check for updates even when the core is unavailable.

- Start background checks after five seconds and bound feature-flag reads, falling back to cached/default values.
- Retry failed setup and checks with backoff, including recovery after reconnect or app resume.

This addresses the startup and recovery portion of getlantern/engineering#3903

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved desktop update-check reliability, including retries after temporary failures.
  * Update checks can resume when the app returns to the foreground or the VPN reconnects.
  * Improved handling of feature-flag and update-service initialization during startup.
* **Testing**
  * Expanded coverage for update checks, retry behavior, manual checks, and service cleanup across supported platforms.


lib/core/updater/updater.dart
```diff
@@ -73,71 +114,148 @@ class Updater with UpdaterListener {
...
+      _scheduleRetry();
+      if (!inBackground) rethrow;
+    } finally {
+      // The method channel returns before the native check finishes.
+      // Result callbacks, rather than this future, reset the retry budget.
+      _dispatchingCheck = false;
+    }
+  }
+
+  void _scheduleCheck(Duration delay, String source) {
```
jigar-f, line 196: Why we are doing our schedule check, even though Sparkle is already doing `setScheduledCheckInterval`, 

Can't we get rid of the multiple schedules check and just keep the startup check and the Sparkle check interval?

PR #9069, opened 2026-09-17
Native logs fix

jigar-f:
This pull request implements robust, cross-platform log file rotation and backup for Android, iOS, and macOS clients. The new rotation logic ensures that log files do not grow unbounded, maintains a consistent backup naming scheme compatible with other components, and prevents issues with excessive uncompressed log files. The rotation system now maintains a bounded live log file and two gzipped backups, with logic to prune older backups and handle legacy files.

**Log rotation and backup improvements:**

* All platforms now rotate logs when the live file exceeds 5 MB, compressing the old log to a gzipped backup with a timestamped name (`<name>-<stamp>.log.gz`), and keep at most two backups, deleting older ones. This matches the Go/Dart/radiance conventions and prevents the issue-report archiver from reading excessive uncompressed logs. [[1]](diffhunk://#diff-a2ccac12c5c7aec356e6a8e596334cc1b8b0daf63e8e9f70095dee75604d3dafR36-R66) [[2]](diffhunk://#diff-2118d9b697aef3260763e7aef498084ec2dbe492cac93cf72edd71a13854ba60R9-R35) [[3]](diffhunk://#diff-49a709fc488a166d7b8d24c0f4582a9aa1ab719b33c4b99a15827fe2c1cafe84R10-R52)
* On Android, legacy backup files with the old naming scheme are deleted on initialization to prevent accumulation of uncompressed logs that could break report submission.
* On iOS and macOS, log rotation and backup logic is implemented in a dedicated dispatch queue to ensure thread safety and avoid races with app extensions. Only the main app process performs rotation. [[1]](diffhunk://#diff-2118d9b697aef3260763e7aef498084ec2dbe492cac93cf72edd71a13854ba60R51-R58) [[2]](diffhunk://#diff-49a709fc488a166d7b8d24c0f4582a9aa1ab719b33c4b99a15827fe2c1cafe84L46-R168)

**Concurrency and performance enhancements:**

* Android logger now uses a single-threaded coroutine scope to serialize writes and rotation, preventing race conditions during log file swapping.
* On iOS and macOS, log file writes and rotation are queued to a dedicated dispatch queue, and the file handle is opened and closed for each write to avoid holding a persistent handle. [[1]](diffhunk://#diff-2118d9b697aef3260763e7aef498084ec2dbe492cac93cf72edd71a13854ba60R51-R58) [[2]](diffhunk://#diff-49a709fc488a166d7b8d24c0f4582a9aa1ab719b33c4b99a15827fe2c1cafe84L46-R168)

**Backup and pruning logic:**

* All platforms use a fixed-width, timestamped backup naming scheme to ensure backups are easily identified and pruned in chronological order. The pruning logic keeps only the two most recent backups. [[1]](diffhunk://#diff-a2ccac12c5c7aec356e6a8e596334cc1b8b0daf63e8e9f70095dee75604d3dafL77-R158) [[2]](diffhunk://#diff-2118d9b697aef3260763e7aef498084ec2dbe492cac93cf72edd71a13854ba60R67-R165) [[3]](diffhunk://#diff-49a709fc488a166d7b8d24c0f4582a9aa1ab719b33c4b99a15827fe2c1cafe84L46-R168)

**Compression implementation:**

* On iOS and macOS, backup compression is implemented using zlib with the gzip container, streaming the log file in chunks to avoid loading it fully into memory. [[1]](diffhunk://#diff-2118d9b697aef3260763e7aef498084ec2dbe492cac93cf72edd71a13854ba60R67-R165) [[2]](diffhunk://#diff-49a709fc488a166d7b8d24c0f4582a9aa1ab719b33c4b99a15827fe2c1cafe84L46-R168)

**Other notable changes:**

* Android log rotation is now checked every 256 KB of appended data, rather than on every write, to reduce file stat overhead.
* Log file initialization and backup deletion are improved for reliability and compatibility across all platforms. [[1]](diffhunk://#diff-a2ccac12c5c7aec356e6a8e596334cc1b8b0daf63e8e9f70095dee75604d3dafR36-R66) [[2]](diffhunk://#diff-2118d9b697aef3260763e7aef498084ec2dbe492cac93cf72edd71a13854ba60R51-R58) [[3]](diffhunk://#diff-49a709fc488a166d7b8d24c0f4582a9aa1ab719b33c4b99a15827fe2c1cafe84R10-R52)

These changes ensure consistent, efficient, and safe log management across all platforms, preventing log file bloat and improving the reliability of issue reporting.

**References:**  
[[1]](diffhunk://#diff-a2ccac12c5c7aec356e6a8e596334cc1b8b0daf63e8e9f70095dee75604d3dafR17-R27) [[2]](diffhunk://#diff-a2ccac12c5c7aec356e6a8e596334cc1b8b0daf63e8e9f70095dee75604d3dafR36-R66) [[3]](diffhunk://#diff-a2ccac12c5c7aec356e6a8e596334cc1b8b0daf63e8e9f70095dee75604d3dafL77-R158) [[4]](diffhunk://#diff-a2ccac12c5c7aec356e6a8e596334cc1b8b0daf63e8e9f70095dee75604d3dafR173-R182) [[5]](diffhunk://#diff-2118d9b697aef3260763e7aef498084ec2dbe492cac93cf72edd71a13854ba60R9-R35) [[6]](diffhunk://#diff-2118d9b697aef3260763e7aef498084ec2dbe492cac93cf72edd71a13854ba60R51-R58) [[7]](diffhunk://#diff-2118d9b697aef3260763e7aef498084ec2dbe492cac93cf72edd71a13854ba60R67-R165) [[8]](diffhunk://#diff-49a709fc488a166d7b8d24c0f4582a9aa1ab719b33c4b99a15827fe2c1cafe84R10-R52) [[9]](diffhunk://#diff-49a709fc488a166d7b8d24c0f4582a9aa1ab719b33c4b99a15827fe2c1cafe84L46-R168)

## Summary by CodeRabbit

- **Improvements**
  - Improved logging reliability across Android, iOS, and macOS.
  - Log files now rotate into compressed backups when they reach their size limit.
  - Logging remains available during rotation, including for tunnel extensions.
  - Up to two compressed backups are retained automatically.
  - Older log backup files are cleaned up during initialization.
  - Improved recovery helps prevent incomplete backups and keeps active logs within size limits.
  - Improved append handling prevents new log entries from overwriting existing data.


ios/Shared/Logger.swift
```diff
@@ -29,17 +48,138 @@ class LanternLogger {
...
+          try outBuf.withUnsafeMutableBufferPointer { outPtr in
+            stream.next_out = outPtr.baseAddress
+            stream.avail_out = UInt32(chunk)
+            rc = deflate(&stream, flush)
+            guard rc == Z_OK || rc == Z_STREAM_END || rc == Z_BUF_ERROR else {
+              throw LogRotationError.deflate(rc)
+            }
+            let produced = chunk - Int(stream.avail_out)
+            if produced > 0 {
+              output.write(Data(bytes: outPtr.baseAddress!, count: produced))
```
myleshorton: [P1] Use throwing file I/O so compression failures reach recovery

`output.write(...)` raises an Objective-C `NSFileHandleOperationException` on write failure, which Swift `catch` cannot intercept. If the backup write fails (for example, because the disk fills), the process can abort instead of removing the partial backup and truncating the live log. I reproduced this with the PR's extracted compression function by imposing a file-size limit: it exited with SIGABRT and never reached the surrounding Swift catch. The same code exists in macos/Shared/Logger.swift. Please use `try output.write(contentsOf:)` and `try input.read(upToCount:)` so I/O failures follow the intended recovery path.

ios/Shared/Logger.swift
```diff
@@ -29,17 +48,138 @@ class LanternLogger {
...
+      let formatted = "time=\"\(timestamp)\" \(message)\n"
+      guard let data = formatted.data(using: .utf8) else { return }
+      if let fileHandle = try? FileHandle(forWritingTo: logFileURL) {
+        // Exclusive lock so an append never lands mid-rotation in the other process.
+        flock(fileHandle.fileDescriptor, LOCK_EX)
+        fileHandle.seekToEndOfFile()
+        fileHandle.write(data)
+        try? fileHandle.close()  // releases the lock
+      }
+      guard canRotate else { return }
```
myleshorton: [P2] Enforce the size bound when only the tunnel extension is running

This guard disables every rotation check in the tunnel extension. During normal background VPN operation, the main app may be suspended or terminated while the extension continues appending, so the shared log can grow beyond 5 MB without a bound. Even with the app running, it checks only after writing 256 KB of its own logs, regardless of how much the extension has written. Please allow extension-triggered rotation, coordinating the size check and rotation under the cross-process file lock so concurrent callers recheck the size after acquiring it.

myleshorton (commented):
Two issues to address in the native log rotation. Gzip round-trip checks passed for empty, random, compressible, and chunk-boundary inputs; the failure-path check below reproduced a process abort.

ios/Shared/Logger.swift
myleshorton: Fixed in 148cdf756 — with one deviation worth flagging.

`write(contentsOf:)` / `read(upToCount:)` don't compile here. They're macOS 10.15.4+, and `macos/Runner.xcodeproj` deploys to 10.15 (uniformly, and `macos/Podfile` says `platform :osx, '10.15'`). Confirmed directly:

```
$ swiftc -typecheck -target arm64-apple-macos10.15 probe.swift
error: 'write(contentsOf:)' is only available in macOS 10.15.4 or newer
error: 'read(upToCount:)' is only available in macOS 10.15.4 or newer
```

That's also why `SystemExtensionManager.swift:1044` wraps its `read(upToCount:)` in `if #available(macOS 10.15.4, *)`, and probably why this file reached for the legacy APIs in the first place. An availability ladder would leave the raising API as the fallback branch — fixing the common case and keeping the bug on the OSes least able to cope.

So: `write(2)`/`read(2)`/`ftruncate(2)` instead. No availability constraint, failure is a return value on every OS we ship, and the file already worked at the descriptor level for `flock`.

**Three call sites, not two.** Beyond the compression loop, `writeToFile` had `fileHandle.write(data)` — added by this PR inside the new lock block — which runs on *every log line* and so is far likelier to meet a full disk than the rotation path. And on macOS the recovery itself used `handle.truncateFile(atOffset: 0)`, also raising, so the truncate meant to run "even if compression failed (e.g. disk full)" could abort instead.

Verified: helpers typecheck at `arm64-apple-macos10.15` and `arm64-apple-ios15.6`, round-trip 5000 bytes across chunk boundaries, empty write is a no-op, and writing to a read-only descriptor throws `io(9)` rather than aborting.

ios/Shared/Logger.swift
myleshorton: Fixed in 148cdf756.

`canRotate` is gone — both processes rotate now. The size check and the truncate are both under the cross-process `flock`, with the recheck you asked for:

```swift
private func rotateIfNeeded() {
  guard currentSize() >= Self.maxFileBytes else { return }   // cheap pre-check
  guard let handle = try? FileHandle(forWritingTo: logFileURL) else { return }
  flock(handle.fileDescriptor, LOCK_EX)
  // Recheck now that the lock is held: the other process may have rotated
  // while we waited, and rotating again would archive a nearly empty file and
  // discard the backup it just wrote.
  guard currentSize() >= Self.maxFileBytes else {
    try? handle.close()
    return
  }
  ...
}
```

The pre-check stays so the common case doesn't stat-then-lock on every interval; the post-lock check is what makes concurrent callers safe. The early return closes the handle, which releases the lock.

Your point about `sinceCheck` was the sharper half, I think — it only ever counted the current process's own `data.count`, so even with the app running the 256 KB interval ignored everything the extension wrote. With both processes rotating, each now trips its own interval against the shared file, and the lock plus recheck makes the overlap harmless.

One corroborating data point: in ticket 183129's bundle `lantern_ios.log` was 6,571,695 bytes — 6.27 MB against the 5 MB `maxFileBytes`, ~25% over. That capture predates this PR so it isn't evidence about this code specifically, but it does show the extension-only-writer case overshooting in the field.

myleshorton:
Pushed 148cdf756 addressing both review threads, since the first one had a wrinkle that needed resolving in code rather than in a comment. Both threads resolved. Flagging the deviations here so they're visible without digging into threads:

**The requested `write(contentsOf:)` / `read(upToCount:)` don't compile on macOS.** They're 10.15.4+, and `macos/Runner.xcodeproj` deploys to 10.15 throughout (`macos/Podfile`: `platform :osx, '10.15'`). Verified with `swiftc -typecheck -target arm64-apple-macos10.15`. It's the same reason `SystemExtensionManager.swift:1044` guards its `read(upToCount:)` with `if #available(macOS 10.15.4, *)`, and likely why this file used the legacy APIs originally. An availability ladder would leave the raising API as the fallback, so I used `write(2)`/`read(2)`/`ftruncate(2)` — no availability constraint, failure is a return value, and the file already worked at the descriptor level for `flock`.

**Three raising call sites, not two.** The review caught the compression loop. Also fixed:
- `writeToFile`'s `fileHandle.write(data)` — added by this PR inside the new lock block, runs on every log line, so it's the likeliest to meet a full disk
- macOS `handle.truncateFile(atOffset: 0)` — the recovery path itself, meant to run "even if compression failed (e.g. disk full)" and able to abort instead

**Rotation now runs in the extension too**, with the size rechecked after taking the cross-process lock so concurrent rotations don't archive a nearly-empty file.

Verification: helpers typecheck at both `arm64-apple-macos10.15` and `arm64-apple-ios15.6`; round-trip 5000 bytes across chunk boundaries; empty write is a no-op; write to a read-only descriptor throws `io(9)` instead of aborting.

Two things for you, @jigar-f:
1. I couldn't build the full Xcode targets locally, so CI's iOS/macOS compile checks are the real verification — worth watching before merge.
2. The 10.15 deployment target is arguably the underlying issue here. macOS 10.15.4 shipped March 2020, so bumping it would let this file use the plain throwing APIs and let `SystemExtensionManager.swift` drop its guard. That's a product call and doesn't belong in this PR, but it's worth someone filing.

myleshorton:
0b09c547d is correct and it's fixing a regression I introduced in 148cdf756 — thanks for catching it.

When I replaced `seekToEndOfFile()` + `write(data)` with `appendAll(fd, data)`, I dropped the seek. `FileHandle(forWritingTo:)` opens at offset 0, and `write(2)` writes at the current offset, so every log line overwrote from the start of the file instead of appending. Demonstrated against the two variants:

```
WITHOUT lseek (mine):  "line-4-aaaaaaaaaa\n"
WITH lseek (the fix):  "line-1-aaaaaaaaaa\nline-2-aaaaaaaaaa\nline-3-aaaaaaaaaa\nline-4-aaaaaaaaaa\n"
```

Four writes, four separate open/write/close cycles. Mine left a one-line file — strictly worse than the raising-API bug I was fixing.

Placing the `lseek` inside the `flock` is right: seek-then-write has to be atomic against the other process, or an append can land at a stale offset after a rotation truncates.

**Why my verification missed it**, since it's worth recording: I tested `appendAll` with a round-trip harness that opened one descriptor, wrote, closed, and read back. On a fresh empty file offset 0 *is* the end, so a single write passes whether or not the seek is there. The bug only appears across repeated open/write/close cycles — which is the actual usage pattern, and the one thing my test didn't model. Compile checks can't catch it either, since it's runtime behavior with no assertion over file contents.

I audited the other descriptor sites for the same class of mistake and they're fine:

| site | offset | correct? |
|---|---|---|
| `ftruncate(handle.fd, 0)` in `rotateIfNeeded` | offset-independent | yes |
| `readChunk(input.fd)` in `compress` | opens at 0, reads sequentially — we want the whole file | yes |
| `appendAll(output.fd)` in `compress` | `createFile(contents: nil)` makes it empty, so 0 == end | yes |
| `appendAll(fileHandle.fd)` in `writeToFile` | **was the bug** | fixed by 0b09c547d |

So the append path was the only offset-dependent one, and both `ios/` and `macos/` got the fix.

One suggestion, entirely optional: a test that writes several lines through `writeToFile` and asserts all of them are present would pin this permanently. It's the assertion that would have failed on my commit and passed on yours, and nothing in the current suite covers log file contents.

myleshorton:
Added the regression test in 72d64bd47.

`testLogLinesAccumulateAcrossWrites` writes five lines through the logger and asserts all five are still present. Mutation-checked it against the actual bug:

| `lseek` in `writeToFile` | result |
|---|---|
| present (0b09c547d) | **passes**, 0.022s |
| removed (my 148cdf756) | **fails** after the full poll deadline |

So it fails on the regression and passes on the fix, which is the only property that makes it worth having.

The writes are dispatched async onto the logger's serial queue, so it polls with a 10s deadline and breaks as soon as all five appear — fast when passing, generous enough not to flake on a slow runner when something is genuinely wrong.

Testing the real write path needs a writable directory, so the log location is now an init parameter:

```swift
convenience init() {
  self.init(logsDirectory: FilePath.logsDirectory)
}

init(logsDirectory: URL) { ... }
```

Production behavior is unchanged — `init()` still resolves `FilePath.logsDirectory`. Applied to `ios/` as well to keep the two files in step, even though only macOS has a test target; these files are maintained as near-copies and letting them drift is how the macOS `truncateFile` instance got missed earlier.

Verified locally on both platforms before pushing:
- `make macos-unit-tests` → `** TEST SUCCEEDED **`
- `make ios-compile-check` → `✓ Built build/ios/iphonesimulator/Runner.app`

PR #9080, opened 2026-09-18
macOS: avoid using stale tunnel interfaces on reconnect

atavism:
A failed tunnel start can leave an old utun descriptor in the extension process. A later connection could select that descriptor and appear connected without passing traffic.

Match fallback descriptors to the IPv4/IPv6 addresses configured for the tunnel, instead of relying on descriptor order. If no interface matches, or multiple interfaces match, fail startup rather than risk using the wrong tunnel. Descriptor ownership remains with NetworkExtension.

Adds regression coverage for stale and reused descriptors, ambiguous matches, duplicate handles, and equivalent IPv6 addresses.

Refs getlantern/engineering#3781

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS tunnel startup to reliably select the active network interface using configured IPv4 and IPv6 addresses.
  * Prevents stale or incorrect tunnel interfaces from being selected during startup.
  * Added handling for missing, invalid, unmatched, or ambiguous interface address configurations.
  * Ensures tunnel address settings are applied consistently, including when automatic routing is disabled.

* **Tests**
  * Added coverage for interface matching, descriptor reuse, IPv4/IPv6 comparisons, and error conditions.


jigar-f (commented):
@atavism, before merging this
- Confirm the Singbox code for macOS  (We did the same for Android recently)
- Can we cut a build and test these changes first?

PR #9084, opened 2026-09-20
logging: use the throwing FileHandle API now that macOS deploys to 11.0

myleshorton:
Follow-up to #9069 and #9083, both merged. The POSIX helpers in `Logger.swift` existed only to dodge an availability constraint that no longer exists.

## Why they were there

`FileHandle`'s throwing `read(upToCount:)` / `write(contentsOf:)` / `truncate(atOffset:)` are **macOS 10.15.4+**, and the macOS project declared **10.15**. The legacy alternatives raise `NSFileHandleOperationException`, which Swift cannot catch — a full disk aborted the process instead of reaching the recovery path that truncates the live log. An `#available` ladder would have left the raising API as its own fallback, so #9069 used `write(2)` / `read(2)` / `ftruncate(2)` instead.

`bf43e5cf3` raised the target to 11.0, so the plain API is now available and carries its own errors.

## What goes

`appendAll`, `readChunk`, the manual `EINTR` and short-write loops, `LogRotationError.io`, and the `ftruncate` / `lseek` calls. **106 lines out, 21 in.**

## What stays, deliberately

The explicit seek. `write(contentsOf:)` writes at the **current offset**, exactly as `write(2)` did, and `FileHandle(forWritingTo:)` opens at zero — so dropping it would overwrite from the start of the file:

```swift
// Each new descriptor starts at zero and write(contentsOf:) writes at
// the current offset, so seek -- while holding the append lock.
try fileHandle.seekToEnd()
try fileHandle.write(contentsOf: data)
```

That's the regression introduced in `148cdf756` and fixed in `0b09c547d`: without the seek the log ends up holding only the most recent line. It stays inside the `flock` so seek-and-write remains atomic against the other process.

## Verification

`testLogLinesAccumulateAcrossWrites` exists precisely to guard this through refactors like this one, so I checked that it still does rather than assuming:

| `seekToEnd()` | result |
|---|---|
| present | **passes**, 0.022s |
| removed | **fails** after the full poll deadline |

So the guard survives the refactor rather than merely compiling alongside it.

Both platforms built locally:
- `make macos-unit-tests` → `** TEST SUCCEEDED **`
- `make ios-compile-check` → `✓ Built build/ios/iphonesimulator/Runner.app`

Grepped the result for `appendAll`, `readChunk`, `ftruncate`, `lseek`, `Darwin.`, `LogRotationError.io` — none remain.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of log file writing, reading, compression, and rotation on iOS and macOS.
  * Log file operation failures are now handled and reported more consistently.


jigar-f (approved):
LGTM, Thanks for making changes

PR #9086, opened 2026-09-21
Fix session tracking and retries for in-app reviews

atavism:
- Count sessions only after confirmed user disconnects.
- Prevent duplicate stops and preserve sessions when stopping fails.
- Skip review requests while backgrounded and retain progress for retry
- Add regression coverage


lib/features/vpn/provider/vpn_notifier.dart
```diff
@@ -20,6 +20,8 @@ part 'vpn_notifier.g.dart';
 @Riverpod(keepAlive: true)
 class VpnNotifier extends _$VpnNotifier {
   bool _hasStatusStreamEmission = false;
+  Future<Either<Failure, String>>? _stopRequest;
+  bool _userDisconnectRequested = false;
```
jigar-f, line 24: @atavism I didn’t add it because we’re counting multiple sessions, not just one. There’s little chance the user will experience continuous crashes, and we show the prompt, so I avoided managing a separate state.

Most of the time, the user connects and then disconnects, so I think we should treat a disconnect as a user having left. Let me know what you think.

atavism: The main issue is that we previously counted the session before stopVPN() succeeded, so a failed disconnect could still advance the counter and trigger a rating request.
The extra state lets us wait for the native disconnected status, avoid overlapping stop calls, and distinguish an explicit user disconnect from other stops. We’re still counting five sessions; this only coordinates the asynchronous disconnect.

